PR #63, opened 2015-04-01
Fix #52 by setting the base layer for dns

### LINKED ISSUE #52: LayerContents nil for DNS layer
```
On DNS packet (testPacketDNSRegression in decode_test.go), I cannot obtain the 
LayerContents.  Root issue is base layer not assigned.

While fixing this bug, I have also added in unit test (in base_test.go) to 
ensure the lengths of all layers will add up to the original packet's data 
length.

Patch can be found in http://pastebin.com/FqvhsUgv

```

Original issue reported on code.google.com by `adrian.c...@gmail.com` on 4 Dec 2014 at 3:58

adriantam:
Fix #52 by setting the baselayer.

### DIFF
--- a/layers/dns.go
+++ b/layers/dns.go
@@ -207,6 +207,10 @@ func (d *DNS) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 		df.SetTruncated()
 		return fmt.Errorf("DNS packet too short")
 	}
+
+	// since there are no further layers, the baselayer's content is
+	// pointing to this layer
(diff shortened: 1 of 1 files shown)


fiorix:
I also need this one.

gconnell:
Thanks for the commit!  I'm happy to merge this in.  Before I do, though, could you sign https://cla.developers.google.com/about/google-individual?  It's required legaleze I need once from every committer, which basically says "you still own the code, but you give us permission to use it".

adriantam:
Done.

PR #65, opened 2015-04-06
fix radiotap field order and size

hkwi:
/FHSS comes after /Channel.
/channel is u16,u16 as http://www.radiotap.org/defined-fields/Channel


gconnell:
Do you happen to have a test packet you could use to create a test for this?  That would be great to make sure we don't regress on this in the future.

hkwi:
Added test code.

PR #68, opened 2015-04-08
implemented support for prism monitor mode header

nl5887:


layers/prism.go
```diff
@@ -0,0 +1,65 @@
...
+
+package layers
+
+import (
+	"encoding/binary"
+
+	"github.com/google/gopacket"
+)
+
+const ()
```
gconnell: not necessary, maybe remove?

layers/prism.go
```diff
@@ -0,0 +1,65 @@
...
+
+func decodePrismValue(data []byte, pv *PrismValue) {
+	pv.Did = binary.LittleEndian.Uint32(data[0:4])
+	pv.Status = binary.LittleEndian.Uint16(data[4:6])
+	pv.Length = binary.LittleEndian.Uint16(data[6:8])
+	pv.Data = binary.LittleEndian.Uint32(data[8:12])
+}
+
+type PrismValue struct {
+	Did    uint32
```
gconnell: Maybe DID?

layers/prism.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (m *PrismHeader) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	m.Code = binary.LittleEndian.Uint16(data[0:4])
+	m.Length = binary.LittleEndian.Uint16(data[4:8])
+	m.DeviceName = string(data[8:24])
+	m.BaseLayer = BaseLayer{Contents: data[:m.Length], Payload: data[m.Length:len(data)]}
+
+	offset := uint16(24)
+
+	m.Values = make([]PrismValue, (m.Length-offset)/12)
```
gconnell: check that remaining data is length 12, return error if not?

layers/prism_test.go
```diff
@@ -0,0 +1,125 @@
+// Copyright 2014, Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	_ "fmt"
```
gconnell: I think this can be removed?

layers/prism.go
gconnell: Also, DID should probably be an enum.  Maybe something like:

type PrismDID uint32
const (
  PrismHostTime PrismDID = X
  PrismMACTime PrismDID = Y
  ...
)

gconnell:
Thanks for the commit!  Just a few minor comments, and I think we'll be good to go.

nl5887:
Our Wireshark friends helped me out again, apparently there are two types (type1 and type2). I was looking at an old (or different spec) Implemented both the type1 and type2 Prism DID types.

PR #71, opened 2015-04-20
defrag: Implement a IPv4 defragmenter (EXPERIMENTAL)

lhausermann:
This is an *experiment*, I've made during a cloudy weekend ;)
It is based on the DefragIPv4 interface suggested by Graeme some months ago.

I am submitting it for  **reviewing purpose and get extensive comments** from the GoPacket Authors

This implements the DefragIPv4 interface, which embed IPv4
defragmentation, and modify packets in place. It returns false
if the packet is fragmented, and caller will need to give subsequent
packets. It returns true, if the packet is in its final state (defragmented
or just pristine).


defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"code.google.com/p/gopacket"
+	"code.google.com/p/gopacket/layers"
+)
+
+// Quick and Easy yo use debug code to trace
```
gconnell: nit: yo -> to

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+var debug debugging = false // or flip to false
+type debugging bool
+
+func (d debugging) Printf(format string, args ...interface{}) {
+	if d {
+		log.Printf(format, args...)
+	}
+}
+
+type Defragmenter interface {
```
gconnell: I'd either change the package name to be ip4defrag, or rename this "IPv4".  Basically, something to say: this code is IPv4 specific.  If we want to implement other (IPv6, etc) defrag logic, I don't want naming conflicts.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	// immediately return true without modifying the layer.
+	//
+	// If the IPv4 layer is a fragment and we don't have all
+	// fragments, it will return false and store whatever internal
+	// information it needs to eventually defrag the packet.
+	//
+	// If the IPv4 layer is the last fragment needed to reconstruct
+	// the packet, *ip will be set to the entire defragmented packet,
+	// and the function will return true.
+	DefragIPv4(IPv4 *layers.IPv4) (bool, error)
```
gconnell: How about:

DefragIPv4(in *layers.IPv4) (*layers.IPv4, error)

Where if 'in' isn't fragmented, we return it.  If it is and it's the last fragment, we return the unfragmented value.  If it's NOT the last fragment, we return (nil, nil).

In particular, I'd like to maintain the guarantee that 'in' is not modified.  This will mean allocating packets for output, but that can be done in batch, which should lessen the problem, and only needs to be done during actual fragmentation (since returning 'in' doesn't require any modification to 'in' OR any allocation).

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+
+	// DiscardOlderThan discards all packets we haven't seen a
+	// fragment for since 't'.
+	// It returns the number of packets discarded in this way.
```
gconnell: When you say "number of packets discarded this way", do you mean packets, or fragments?

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	ipf := new(IPv4Flow)
+	ipf.net = ip.NetworkFlow()
+	ipf.id = ip.Id
+
+	return ipf
+}
+
+// IPv4Defragmenter is a struct which embedded a map of
+// all fragment/packet.
+type IPv4Defragmenter struct {
```
gconnell: Just call this "IPv4", to avoid stutter?

defrag/defrag.go
gconnell: Is this interface actually necessary?  Why not just return *IPv4Defragmenter?  If folks need an interface for testing, they can always make their own, possibly with just the subset of methods they care about.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+
+	return true, nil
+}
+
+// FragmentList holds a container/list used to contains IP
+// packets/fragments.  It stores internal counters to track the
+// maximum total of byte, and the current length it has received.
+// It also stores a flag to know if he has seen the last packet.
+type FragmentList struct {
```
gconnell: Let's make sure to only export the stuff we expect other folks to use.  I specifically think we don't want to export this, because there are many possible ways of defrag'ing packets, and we might want to switch which we use in the future away from a linked-list approach.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	// TODO recompute IP Checksum
+	ip.Payload = final
+	ip.Length = f.Total
+	ip.FragOffset = 0
+	ip.Flags = 0
+
+	return true, nil
+}
+
+// IPv4Flow is a struct to be used as a key.
```
gconnell: don't export

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+// all fragment/packet.
+type IPv4Defragmenter struct {
+	sync.RWMutex
+	IPFlows map[IPv4Flow]*FragmentList
+}
+
+// NewIPv4Defragmenter returns a new IPv4Defragmenter
+// with an initialized map.
+func NewIPv4Defragmenter() *IPv4Defragmenter {
+	d := new(IPv4Defragmenter)
```
gconnell: combine?

d := &IPv4Defragmenter{
  IPFlows: map[IPv4Flow]*FragmentList{},
}

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// IPv4Flow is a struct to be used as a key.
+type IPv4Flow struct {
+	net gopacket.Flow
+	id  uint16
+}
+
+// NewIPv4Flow returns a new initialized IPv4Flow
+func NewIPv4Flow(ip *layers.IPv4) *IPv4Flow {
```
gconnell: Above, you immediately deref this pointer.  So instead, just return IPv4Flow.  It'll save you a whole memory allocation.

func NewIPv4Flow(ip *layers.IPv4) IPv4Flow {
  return IPv4Flow{
    net: ip.NetworkFlow(),
    id: ip.Id,
  }
}

gconnell:
Initial set of comments is mostly stylistic and API stuff (what to export, how API should look, etc).  I haven't done a deep review of the actual logic yet, but hope to in the next day or so.

lhausermann:
OK thanks for the initial comments. 
Waiting yoru other review round about logic and so one,  to make all fixes at once, and submit it again

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+			delete(d.IPFlows, k)
+		}
+	}
+	d.Unlock()
+	return nb
+}
+
+// DontDefrag returns true if the IPv4 packet do not need
+// any defragmentation
+func (d *IPv4Defragmenter) DontDefrag(ip *layers.IPv4) bool {
```
gconnell: Even on the Defragmenter, I'd only export the stuff you're currently exporting with the Defragmenter interface.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	// don't defrag packet with DF flag
+	if ip.Flags&layers.IPv4DontFragment != 0 {
```
gconnell: If the DontFragment bit is set on a fragment, we should probably return an error from DefragIPv4

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	// don't defrag not fragmented ones
+	if ip.Flags == 0 && ip.Id == 0 {
+		return true
+	}
+	return false
+}
+
+// SecurityChecks performs the needed security checks
+func (d *IPv4Defragmenter) SecurityChecks(ip *layers.IPv4) (bool, error) {
```
gconnell: just return error, and check if it's nil.  No need for the extra bool.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	return true, nil
+}
+
+// FragmentList holds a container/list used to contains IP
+// packets/fragments.  It stores internal counters to track the
+// maximum total of byte, and the current length it has received.
+// It also stores a flag to know if he has seen the last packet.
+type FragmentList struct {
+	List          list.List
+	Total         uint16
```
gconnell: "Total" seems like it would be a sum, but it's actually the highest index of a byte we've seen.  Maybe call it "Highest", or something like that, to signify that it's a max, not a sum?

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+			frag, _ := e.Value.(*layers.IPv4)
+			if ip.FragOffset <= frag.FragOffset {
+				debug.Printf("defrag: inserting frag %d before existing frag %d \n",
+					ip.FragOffset*8, frag.FragOffset*8)
+				f.List.InsertBefore(ip, e)
+				break
+			}
+		}
+	}
+	// packet.Metadata().Timestamp should have been better, but
```
gconnell: plumb this in by passing the packet's timestamp in with the DefragIPv4 call?  That's what we ended up doing with tcpassembly.

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+	FinalReceived bool
+	LastSeen      time.Time
+}
+
+// Insert insert an IPv4 fragment/packet into the Fragment List
+// It use the following strategy : we are inserting fragment based
+// on their offset, latest first. This is sometimes called BSD-Right.
+// See: http://www.sans.org/reading-room/whitepapers/detection/ip-fragment-reassembly-scapy-33969
+func (f *FragmentList) Insert(ip *layers.IPv4) (bool, error) {
+	if ip.FragOffset*8 >= f.Total {
```
gconnell: You use ip.FragOffset*8 a bunch, maybe start with:

fragOffset := ip.FragOffset * 8

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+	// packet.Metadata().Timestamp should have been better, but
+	// we don't have this info there...
+	f.LastSeen = time.Now()
+
+	fragLength := ip.Length - 20
+	// After inserting the Fragment, we update the counters
+	if f.Total < ip.FragOffset*8+fragLength {
+		f.Total = ip.FragOffset*8 + fragLength
```
gconnell: Maybe set Total based on whether the MoreFragments flag is zero?

if ip.MF == 0 {
  total = ip.FragOffset*8+fraglen
}

defrag/defrag.go
```diff
@@ -0,0 +1,276 @@
...
+			debug.Printf("defrag: building - overlapping, starting at %d\n",
+				startAt)
+			if startAt > frag.Length-20 {
+				return false, fmt.Errorf("defrag: building - invalid fragment")
+			}
+			final = append(final, frag.Payload[startAt:]...)
+			currentOffset = currentOffset + frag.FragOffset*8
+		} else {
+			// Houston - we have an hole !
+			return false, nil
```
gconnell: add debug message about where the hole is?

gconnell: Is this an implementation error?  Should we ever call Build unless we already have all the packets?

gconnell:
First round of implementation comments all done.  PTAL.

defrag/defrag.go
lhausermann: OK

defrag/defrag.go
lhausermann: Well.

1 - I was thinking we would add next a IPv6 defrag, edven if IPv6  is not supposed to be fragment at all. I will rename the package ip4defrag

2 - Regarding the interface, it's the way you specified in the Wishlist ;) Isn't it a good practice to have some high level interface for each major package ? Give me your views on that.

defrag/defrag.go
lhausermann: I am not sure to understand your point correctly. Would you please give the caller skeleton ?
I mean, if we return "false", the caller already have the packet because he just passed it... there is no new allocation.

defrag/defrag.go
lhausermann: In fact packets because we are keeping a list of packets. I don't see a incentive to copy the IP Payload and keep the header elsewhere... so I've kept a list of full IPv4 layers..

Do you think the term "packet" is confusing here ?

defrag/defrag.go
lhausermann: OK

defrag/defrag.go
lhausermann: Yes, done

defrag/defrag.go
lhausermann: OK

defrag/defrag.go
lhausermann: In fact, the NewIPv4Defrag is intended to be used outside the package, without the need to disclose the internal fields...

defrag/defrag.go
lhausermann: OK

defrag/defrag.go
lhausermann: OK, done

defrag/defrag.go
lhausermann: At that point, I don't know if it's a fragment. It's in fact the first test I am doing when I received a IP layer... DF flag means "do nothing" for a defragmenter, it's not an error..

defrag/defrag.go
lhausermann: Well, it's not an error. It's the bool of the interface. 
True means "OK I do not need to defrag that packet", false means "yes launch the defrag process"

defrag/defrag.go
lhausermann: OK done

defrag/defrag.go
lhausermann: As at last this is only used for Discarding packets, I am not sure, it worth the complexity

defrag/defrag.go
lhausermann: OK done

defrag/defrag.go
lhausermann: well, as you noted below, Total is a wrong variable name : it's in fact the highest value we have seen.
So no, let's keep it that way

defrag/defrag.go
lhausermann: We are calling build when we got the last packet with MF=true and we now that we got the same volume of byte (total sum of fraglenth is equal to highest seen)...

But that do not mean we have a full suite of fragment, so the test is valid. Should we consider this as an error... I am not sure.

defrag/defrag.go
gconnell: Re: wishlist:  sorry about that.  My attempt with that was to show what functions/args the type should provide, not to actually use an interface.

I think it would be best to work like https://godoc.org/github.com/google/gopacket/tcpassembly#Assembler does.  Don't worry about a high-level interface here... just create a type with a small number of exported functions/vars.  If folks want, they can create high-level interfaces for that type on their own.

I think that with Go, a high-level interface per package actually isn't necessary or particularly useful.  In any language with a hierarchical inheritance model, it's hugely necessary, since otherwise users can't implement their own implementations.  But in Go, because of duck-typing, if I want to make my code modular I can write the interface to match the implementation.  So if I want to, for example, inject test code, I can write my own interface with just the functions I care about, pass in the production class in normal code, then pass in my own test code in a test.

A good example of this approach is gopacket.PacketSource, pcap.Handle, afpacket.TPacket, etc.  The packages that provide implementations (pcap.Handle, afpacket.TPacket) don't expose interfaces... they just give you types you can use.  But the system that needs an interface (gopacket) creates one (PacketSource) where/when it needs to.

defrag/defrag.go
gconnell: Here's the code I'm envisioning for the defragger:

    func (d *IPv4) DefragIPv4(in *layers.IPv4) (*layers.IPv4, error) {
      if err := checkValidity(in); err != nil {
        return nil, err
      } else if notFragmented(in) {
       return in, nil
      }
      packetDefragger := d.findDefragger(in.src, in.dst, in.id)
      if err := packetDefragger.addFragment(in); err != nil {
        return nil, err
      }
      if packetDefragger.complete() {
        return packetDefragger.defragged(), nil
      }
      return nil, nil
    }

And here's the code I'm envisioning for the client:

    func HandlePacket(in *layers.IPv4) err {
      in, err := defragger.DefragIPv4(in)
      if err != nil {
        return err
      } else if in == nil {
        return nil  // packet fragment, we don't have whole packet yet.
      }
      // At this point, we know that 'in' is defragmented.  It may be the same 'in' passed to
      // HandlePacket, or it may not, but we don't really care :)
      ... do stuff to 'in' ...
    }

defrag/defrag.go
gconnell: Never mind, I think this is fine as is.

defrag/defrag.go
gconnell: Sorry, I meant _inside_ the NewIPv4Defragmenter method, IE:

    func NewIPv4() *IPv4 {
      return &IPv4{
        ipFlows: map[IPv4Flow]*FragmentList{},
        ... other stuff ...
      }
    }

defrag/defrag.go
gconnell: My concern with not passing in timestamps is that:
1) this makes it impossible to handle .pcap files like we handle on-the-wire traffic, since they have old timestamps
2) testing will be more difficult, since we'll have a nondeterministic time

defrag/defrag.go
lhausermann: OK

defrag/defrag.go
lhausermann: OK, allright.
Your are right : I have some old java-ish habits ;)
Done, I removed the interface.

defrag/defrag.go
lhausermann: OK. I'll push a new version with that prototype. Let me know.

defrag/defrag.go
lhausermann: OK. Done.
By the way, is only stylish or anything related to performance ?

defrag/defrag.go
lhausermann: OK. I suggest we keep that point for the global TCP assembly discussion

defrag/defrag.go
```diff
@@ -0,0 +1,293 @@
...
+
+// dontDefrag returns true if the IPv4 packet do not need
+// any defragmentation
+func (d *IPv4Defragmenter) dontDefrag(ip *layers.IPv4) bool {
+	// don't defrag packet with DF flag
+	if ip.Flags&layers.IPv4DontFragment != 0 {
+		return true
+	}
+	// don't defrag not fragmented ones
+	if ip.Flags == 0 && ip.Id == 0 {
```
gconnell: As far as I know, it's actually not forbidden to set IPv4.Id on a packet even if it's not fragmented.  I think you want:

return ip.Flags & MoreFragments == 0 && ip.FragOffset == 0

defrag/defrag.go
```diff
@@ -0,0 +1,293 @@
+package ip4defrag
```
gconnell: Would you mind adding a package comment here?

Also, move file into dir ip4defrag/ (instead of defrag/) if you're going to use that package name.

gconnell:
I think we're mostly down to very small comments/nits, so I'd expect we'd be good to merge after the next commit.

Could you add ip4defrag to .travis.cfg in the toplevel dir so it's tested automatically?

defrag/defrag.go
gconnell: Not done?

lhausermann: Oops... done now

defrag/defrag.go
lhausermann: Yep

defrag/defrag.go
lhausermann: Hum... After some research, I found RFC 6864 and yes, Id field is/would be used for other stuff.

lhausermann:
Mmm, I did not find any .travis file ?

lhausermann:
Well, how should we use it inside the examples/* ?
Add it as a non-default option of dumpcommand at the begining of the main loop ?

gconnell:
Travis config is here:  https://github.com/google/gopacket/blob/master/.travis.yml

I just added it a few days ago, so you may need to pull from master to get it.

I like the idea of adding some example code to dumpcommand.

gconnell:
Also, I think this is in a fine state for an initial push to master.  I've added you as a collaborator, so you should be able to merge as you see fit after the .travis.yml addition.  Please merge whenever you feel comfortable with it, or let me know if you have additioanl questions.

gconnell:
Quick ping on this.  Feel free to merge it whenever you think it's ready.

lhausermann:
Hi Graham, 
I've commited the dumpcommand part. 

There is still an issue : when I defrag packet, it looks like the packet I am writing as "out", is still considered as a Fragment... I suppose it's because we dont re-decode it..

How would like to handle this issue ? The IPv4 defrag should call the decode routine once again ?

lhausermann:
I've also added a secondary email address to my CLA...

lhausermann:
Well...
I've try to understand how to handle that decoding stuff.. I've done what you asked (ie rebuilding the layer), and more (calling once again decode), but I still have incoherent decoding...

On one side: 
$ ping -c1 -s2000 google.com 
On the other 
$ sudo ./pcapdump -i=wlan0 -defrag=true icmp
Using BPF filter "icmp"
Starting to read packets
Decoding re-assembled packet: ICMPv4
PACKET: 562 bytes, wire length 562 cap length 562 @ 2015-06-13 00:20:14.902541 +0200 CEST
- Layer 1 (14 bytes) = Ethernet {Contents=[..14..] Payload=[..548..] SrcMAC=7c:7a:91:7d:7c:74 DstMAC=f4:ca:e5:4e:e1:46 EthernetType=IPv4 Length=0}
- Layer 2 (20 bytes) = IPv4     {Contents=[..20..] Payload=[..528..] Version=4 IHL=5 TOS=0 Length=548 Id=18251 Flags= FragOffset=185 TTL=64 Protocol=ICMPv4 Checksum=39123 SrcIP=192.168.1.23 DstIP=173.194.40.128 Options=[] Padding=[]}
- Layer 3 (528 bytes) = Fragment        528 byte(s)
- Layer 4 (08 bytes) = ICMPv4   {Contents=[..8..] Payload=[..2000..] TypeCode=EchoRequest(0) Checksum=58624 Id=8864 Seq=1}
- Layer 5 (2000 bytes) = Payload        2000 byte(s)

Decoding re-assembled packet: ICMPv4
PACKET: 562 bytes, wire length 562 cap length 562 @ 2015-06-13 00:20:14.950885 +0200 CEST
- Layer 1 (14 bytes) = Ethernet {Contents=[..14..] Payload=[..548..] SrcMAC=f4:ca:e5:4e:e1:46 DstMAC=7c:7a:91:7d:7c:74 EthernetType=IPv4 Length=0}
- Layer 2 (20 bytes) = IPv4     {Contents=[..20..] Payload=[..528..] Version=4 IHL=5 TOS=0 Length=548 Id=20433 Flags= FragOffset=185 TTL=55 Protocol=ICMPv4 Checksum=39245 SrcIP=173.194.40.128 DstIP=192.168.1.23 Options=[] Padding=[]}
- Layer 3 (528 bytes) = Fragment        528 byte(s)
- Layer 4 (08 bytes) = ICMPv4   {Contents=[..8..] Payload=[..2000..] TypeCode=EchoReply(0) Checksum=60672 Id=8864 Seq=1}
- Layer 5 (2000 bytes) = Payload        2000 byte(s)

lhausermann:
Hey cla/google, I am one person with two emails ! Both are in the CLA site, and my Github account !

PR #74, opened 2015-05-04
Fix GRE header decoding

kaorimatz:
To follow [RFC 1701](https://tools.ietf.org/html/rfc1701) and [RFC 2784](https://tools.ietf.org/html/rfc2784)

- check if the optional fields exist before consuming data
- `Routing` field is a list of SREs

I have a [pcap](https://github.com/kaorimatz/pcaps/raw/master/gre/ip/gre-ip__ping__veth2.pcap) containing GRE encapsulated packets, and tested this change with the following steps:

    $ cat <<EOF > pcapdump.go
    package main
    
    import (
            "fmt"
            "github.com/google/gopacket"
            "github.com/google/gopacket/pcapgo"
            "os"
    )
    
    func main() {
            if r, err := pcapgo.NewReader(os.Stdin); err != nil {
                    panic(err)
            } else {
                    src := gopacket.NewPacketSource(r, r.LinkType())
                    for p := range src.Packets() {
                            fmt.Printf("%+v\n", p)
                    }
            }
    }
    EOF
    $ go build pcapdump.go
    $ curl -fsSL https://github.com/kaorimatz/pcaps/raw/master/gre/ip/gre-ip__ping__veth2.pcap | ./pcapdump

Unfortunately, none of the GRE headers in the pcap contain routing information, so I have not tested the latter change yet.


layers/gre.go
```diff
@@ -19,12 +19,15 @@ type GRE struct {
...
 	Checksum, Offset                                                           uint16
 	Key, Seq                                                                   uint32
-	*GRERouting
+	GRERouting                                                                 GRERouting
 }
 
 // GRERouting is GRE routing information, present if the RoutingPresent flag is
 // set.
-type GRERouting struct {
+type GRERouting []*GRESourceRouteEntry
```
gconnell: This type switching here is backwards-incompatible to anyone who's already using this code.  How about instead of a slice, you make GRERouting a linked list?  Basically, just add a

    Next *GRERouting

field to the existing struct, which is nil if it's the last one and non-nil if there's more.

layers/gre_test.go
```diff
@@ -0,0 +1,41 @@
...
+	0x1e, 0x1f, 0x20, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x28, 0x29, 0x2a, 0x2b, 0x2c, 0x2d,
+	0x2e, 0x2f, 0x30, 0x31, 0x32, 0x33, 0x34, 0x35, 0x36, 0x37,
+}
+
+func TestPacketGRE(t *testing.T) {
+	p := gopacket.NewPacket(testPacketGRE, LinkTypeEthernet, gopacket.Default)
+	if p.ErrorLayer() != nil {
+		t.Error("Failed to decode packet:", p.ErrorLayer().Error())
+	}
+	checkLayers(p, []gopacket.LayerType{LayerTypeEthernet, LayerTypeIPv4, LayerTypeGRE, LayerTypeIPv4, LayerTypeICMPv4, gopacket.LayerTypePayload}, t)
```
gconnell: check the number of GRE source-routing entries in LayerTypeGRE?

kaorimatz:
I signed it!

gconnell:
Thanks for your contribution!  If you'd like, feel free to add yourself as a contributor in the top-level AUTHORS file and drop me a pull request for it.

kaorimatz:
Thank you! I will add myself to AUTHORS file in my next contribution.

PR #75, opened 2015-05-05
Fix wrong interpretation of the length of AH

kaorimatz:
[RFC 4302](http://tools.ietf.org/html/rfc) states the payload length field in AH specifies the length of AH in 32-bit words, not the 64-bit words used by other IPv6 extension headers. This PR removes embedded `ipv6ExtensionBase` from `IPSecAH` and make `IPSecAH` calculate the length of its layer contents.

This PR also fixes an decoding error of tunnel mode AH packets by specifying the appropriate decoder for IPv4 encapsulation (protocol 4).

The pcaps used in tests are [`testPacketIPSecAHTransport`](https://github.com/kaorimatz/pcaps/raw/master/ah/transport/ah-transport__ping__veth2.pcap) and [`testPacketIPSecAHTunnel`](https://github.com/kaorimatz/pcaps/raw/master/ah/tunnel/route/ah-tunnel-route__ping__veth2.pcap).


layers/enums.go
```diff
@@ -66,6 +66,7 @@ const (
 	IPProtocolIPv6HopByHop    IPProtocol = 0
 	IPProtocolICMPv4          IPProtocol = 1
 	IPProtocolIGMP            IPProtocol = 2
+	IPProtocolIPv4            IPProtocol = 4
```
gconnell: Maybe call this IPinIP?

layers/ipsec.go
```diff
@@ -14,10 +14,10 @@ import (
 // IPSecAH is the authentication header for IPv4/6 defined in
 // http://tools.ietf.org/html/rfc2402
 type IPSecAH struct {
-	// While the auth header can be used for both IPv4 and v6, its format is that of
-	// an IPv6 extension (NextHeader, PayloadLength, etc...), so we use ipv6ExtensionBase
-	// to build it.
-	ipv6ExtensionBase
```
gconnell: how about keeping ipv6ExtensionBase up here, but not using decodeIPvExtensionBase below?

I think that's slightly more backwards-compatible.

kaorimatz:
Thank you for merging.

PR #76, opened 2015-05-05
Support decoding Transparent Ethernet Bridging packets

kaorimatz:
The protocol type field of an Ethernet over GRE packet is filled with 0x6558 (Transparent Ethernet Bridging). This PR adds support for decoding such packets.


gconnell:
Nice work, thanks for the tests :)

PR #82, opened 2015-05-22
[layers/Radiotap] add HT, VHT field support

hkwi:
add support for 802.11n, 802.11ac fields

### DIFF
--- a/layers/radiotap.go
+++ b/layers/radiotap.go
@@ -331,9 +331,18 @@ type RadioTapMcs struct {
 func (self RadioTapMcs) String() string {
 	var tokens []string
 	if self.Known.Bandwidth() {
-		tokens = append(tokens, []string{
-			"20", "40", "40(20L)", "40(20U)",
-		}[self.Flags.Bandwidth()])
+		token := "?"
+		switch self.Flags.Bandwidth() {
+		case 0:
+			token = "20"
+		case 1:
+			token = "40"
+		case 2:
+			token = "40(20L)"
+		case 3:
+			token = "40(20U)"
+		}
+		tokens = append(tokens, token)
 	}
 	if self.Known.McsIndex() {
 		tokens = append(tokens, fmt.Sprintf("mcsIndex#%d", self.Mcs))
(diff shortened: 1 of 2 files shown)


layers/radiotap.go
```diff
@@ -42,6 +43,10 @@ const (
 	RadioTapPresentTxFlags
 	RadioTapPresentRtsRetries
 	RadioTapPresentDataRetries
+	_
+	RadioTapPresentMcs
```
gconnell: please capitalize acronyms:  MCS and VHT

layers/radiotap.go
```diff
@@ -99,6 +104,15 @@ func (r RadioTapPresent) RtsRetries() bool {
 func (r RadioTapPresent) DataRetries() bool {
 	return r&RadioTapPresentDataRetries != 0
 }
+func (r RadioTapPresent) Mcs() bool {
```
gconnell: Ditto down here... MCS and VHT

layers/radiotap.go
```diff
@@ -260,6 +274,335 @@ func (a RadioTapChannelFrequency) String() string {
...
+	if self.Known.Beamformed() {
+		if self.Flags.Beamformed() {
+			tokens = append(tokens, "beamformed")
+		} else {
+			tokens = append(tokens, "no beamformed")
+		}
+	}
+	if self.Known.Bandwidth() {
+		tokens = append(tokens, []string{
+			"20",
```
gconnell: Move this slice up into a global var:

var bandwidthStrings = []string{...}

layers/radiotap.go
gconnell: In general, prefer capitalizing acronyms everywhere... Not sure if MCS and VHT are acronyms, but if so, capitalizing them and other new vars/functions/etc. with acronyms in them would be great.  We don't do it as consistently as we like, but try to do it for new code.

gconnell:
Quick ping on this?

hkwi:
I'd send patch within few days.

hkwi:
I updated the patch.

PR #88, opened 2015-06-16
Add rough draft BSD BPF sniffer API

david415:
OK... here's the code. I haven't done a smoke test lately but last time I tested it worked on OpenBSD, FreeBSD and NetBSD.

Maybe we should be calling SetBpfBuflen instead of using the default. OpenBSD and NetBSD both default to a BPF bufLen of 32767 but FreeBSD uses 4096... obviously having a larger bufLen should improve performance.

### DIFF
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -1,21 +1,10 @@
 // +build darwin dragonfly freebsd netbsd openbsd
 
-/*
- * This rough sketch of a golang API for a Berkeley packet filter sniffer for BSD variants.
- * It provides only the capability to sniff ethernet frames... but could easily be extended for
- * sending frames as well.
- *
- * This file was inspired by BSD licensed code from https://github.com/songgao/ether
- *
- * ( http://opensource.org/licenses/BSD-3-Clause )
- *
- * The bpf_wordalign function was borrowed... and readFrames was very much inspired by
- * design I saw in songgao's ether git repository; the rest of the code I wrote myself ;-)
- *
- * Author: David Anthony Stainton
- * License: BSD
- *
- */
+// Copyright 2012 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
 
 package bpf_sniffer
 
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -37,9 +37,10 @@ type BPFSniffer struct {
 	readChan chan TimedFrame
 }
 
-func NewBPFSniffer(name string) *BPFSniffer {
+func NewBPFSniffer(sniffDeviceName, bpfDeviceName string) *BPFSniffer {
 	return &BPFSniffer{
-		sniffDeviceName: name,
+		sniffDeviceName: sniffDeviceName,
+		bpfDeviceName: bpfDeviceName,
 		stopChan: make(chan bool, 0),
 		readChan: make(chan TimedFrame, 0),
 	}
@@ -61,7 +62,9 @@ func (b *BPFSniffer) Init() error {
 	var err error
 	enable := 1
 
-	b.pickBpfDevice()
+	if b.bpfDeviceName == "" {
+		b.pickBpfDevice()
+	}
 
 	err = syscall.SetBpfInterface(b.fd, b.sniffDeviceName)
 	if err != nil {
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -24,22 +24,44 @@ func bpfWordAlign(x int) int {
 	return (((x) + (wordSize - 1)) &^ (wordSize - 1))
 }
 
+// Options is used to configure various properties of the BPF sniffer.
+// Default values are used when a nil Options pointer is passed to NewBPFSniffer.
 type Options struct {
-	bpfDeviceName    string
-	readBufLen       int
-	timeout          *syscall.Timeval
-	promisc          bool
-	immediate        bool
-	preserveLinkAddr bool
+	// BPFDeviceName is name of the bpf device to use for sniffing
+	// the network device. The default value of BPFDeviceName is empty string
+	// which causes the first available BPF device file /dev/bpfX to be used.
+	BPFDeviceName string
+	// ReadBufLen specifies the size of the buffer used to read packets
+	// off the wire such that multiple packets are buffered with each read syscall.
+	// The packet will be clipped if it's size exceeds the buffer size.
+	// A larger buffer should increase performance because fewer read syscalls would be made.
+	// If zero is used the system's default buffer length will be used.
+	// ReadBufLen defaults to 32767.
+	ReadBufLen int
+	// Timeout is the length of time to wait before timing out on a read request.
+	// Timeout defaults to nil which means no timeout is used.
+	Timeout *syscall.Timeval
+	// Promisc is set to true for promiscuous mode ethernet sniffing.
+	// Promisc defaults to true.
+	Promisc bool
+	// Immediate is set to true to make our read requests return as soon as a packet becomes available.
+	// Otherwise, a read will block until either the kernel buffer becomes full or a timeout occurs.
+	// The default is false.
+	Immediate bool
+	// PreserveLinkAddr is set to false if the link level source address should be filled in automatically
+	// by the interface output routine. Set to true if the link level source address will be written,
+	// as provided, to the wire.
+	// The default is true.
+	PreserveLinkAddr bool
 }
 
 var defaultOptions = Options{
-	bpfDeviceName:    "",
-	readBufLen:       0,
-	timeout:          nil,
-	promisc:          true,
-	immediate:        false,
-	preserveLinkAddr: true,
+	BpfDeviceName:    "",
+	ReadBufLen:       32767,
+	Timeout:          nil,
+	Promisc:          true,
+	Immediate:        false,
+	PreserveLinkAddr: true,
 }
 
 type BPFSniffer struct {
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -20,7 +20,7 @@ import (
 
 const wordSize = int(unsafe.Sizeof(uintptr(0)))
 
-func bpf_wordalign(x int) int {
+func bpfWordAlign(x int) int {
 	return (((x) + (wordSize - 1)) &^ (wordSize - 1))
 }
 
@@ -29,7 +29,7 @@ type TimedFrame struct {
 	Timestamp time.Time
 }
 
-type BpfSniffer struct {
+type BPFSniffer struct {
 	sniffDeveiceName string
 	bpfDeviceName string
 	fd       int
@@ -37,15 +37,15 @@ type BpfSniffer struct {
 	readChan chan TimedFrame
 }
 
-func NewBpfSniffer(name string) *BpfSniffer {
-	return &BpfSniffer{
+func NewBPFSniffer(name string) *BPFSniffer {
+	return &BPFSniffer{
 		sniffDeviceName: name,
 		stopChan: make(chan bool, 0),
 		readChan: make(chan TimedFrame, 0),
 	}
 }
 
-func (b *BpfSniffer) pickBpfDevice() {
+func (b *BPFSniffer) pickBpfDevice() {
 	for i := 0; i < 99; i++ {
 		b.bpfDeviceName = fmt.Sprintf("/dev/bpf%d", i)
 		b.fd, err = syscall.Open(b.bpfDeviceName, syscall.O_RDWR, 0)
@@ -57,7 +57,7 @@ func (b *BpfSniffer) pickBpfDevice() {
 
 // Init is used to initialize a BPF device for promiscuous sniffing.
 // It also starts a goroutine to continuously read frames.
-func (b *BpfSniffer) Init() error {
+func (b *BPFSniffer) Init() error {
 	var err error
 	enable := 1
 
@@ -92,11 +92,11 @@ func (b *BpfSniffer) Init() error {
 	return nil
 }
 
-func (b *BpfSniffer) Stop() {
+func (b *BPFSniffer) Stop() {
 	b.stopChan <- true
 }
 
-func (b *BpfSniffer) readFrames() {
+func (b *BPFSniffer) readFrames() {
 	bufLen, err := syscall.BpfBuflen(b.fd)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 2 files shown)


bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
+// +build darwin dragonfly freebsd netbsd openbsd
+
+/*
+ * This rough sketch of a golang API for a Berkeley packet filter sniffer for BSD variants.
```
gconnell: Please use the same copyright as the rest of gopacket, specifically:

// Copyright 2012 Google, Inc. All rights reserved.
//
// tree.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+ * sending frames as well.
+ *
+ * This file was inspired by BSD licensed code from https://github.com/songgao/ether
+ *
+ * ( http://opensource.org/licenses/BSD-3-Clause )
+ *
+ * The bpf_wordalign function was borrowed... and readFrames was very much inspired by
+ * design I saw in songgao's ether git repository; the rest of the code I wrote myself ;-)
+ *
+ * Author: David Anthony Stainton
```
gconnell: Instead of your name here, please add yourself to the AUTHORS file in the top-level directory.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+ * License: BSD
+ *
+ */
+
+package bpf_sniffer
+
+import (
+	"fmt"
+	"github.com/google/gopacket"
```
gconnell: Split up imports into standard library vs. other:

import (
  "fmt"
  "syscall"
  "time"
  "unsafe"

  "github.com/google/gopacket"
  "golang.org/x/sys/unix"
)

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+	"golang.org/x/sys/unix"
+	"syscall"
+	"time"
+	"unsafe"
+)
+
+const wordSize = int(unsafe.Sizeof(uintptr(0)))
+
+func bpf_wordalign(x int) int {
```
gconnell: please use camelCase for all functions and vars.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+	return (((x) + (wordSize - 1)) &^ (wordSize - 1))
+}
+
+type TimedFrame struct {
+	RawFrame  []byte
+	Timestamp time.Time
+}
+
+type BpfSniffer struct {
```
gconnell: capitalize acronyms:  BPFSniffer

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+		b.fd, err = syscall.Open(b.bpfDeviceName, syscall.O_RDWR, 0)
+		if err == nil {
+			break
+		}
+	}
+}
+
+// Init is used to initialize a BPF device for promiscuous sniffing.
+// It also starts a goroutine to continuously read frames.
+func (b *BpfSniffer) Init() error {
```
gconnell: make this part of NewBPFSniffer()?  Or is there other configuration that can be done between NewBPFSniffer() and Init()?

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+	return &BpfSniffer{
+		sniffDeviceName: name,
+		stopChan: make(chan bool, 0),
+		readChan: make(chan TimedFrame, 0),
+	}
+}
+
+func (b *BpfSniffer) pickBpfDevice() {
+	for i := 0; i < 99; i++ {
+		b.bpfDeviceName = fmt.Sprintf("/dev/bpf%d", i)
```
gconnell: allow user to pick their device?

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+
+	go b.readFrames()
+	return nil
+}
+
+func (b *BpfSniffer) Stop() {
+	b.stopChan <- true
+}
+
+func (b *BpfSniffer) readFrames() {
```
gconnell: using a readFrames goroutine and readChan channel seems overly complex... instead, how about just calling lines 121-136 in ReadPacketData().

That way, if folks want to do things in a separate goroutine, they can, but if not, they don't have to.  It also makes error propagation from syscall.Read easier (it's just returned as an error from ReadPacketData), and there's no need  for stopChan.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+	err = syscall.SetBpfPromisc(b.fd, enable)
+	if err != nil {
+		return err
+	}
+
+	go b.readFrames()
+	return nil
+}
+
+func (b *BpfSniffer) Stop() {
```
gconnell: Other ones seem to call this Close() instead of Stop()

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,153 @@
...
+						Timestamp: time.Unix(int64(hdr.Tstamp.Sec), int64(hdr.Tstamp.Usec)*1000),
+					}
+					p += bpf_wordalign(int(hdr.Hdrlen) + int(hdr.Caplen))
+				}
+			}
+		}
+	}
+}
+
+func (b *BpfSniffer) ReadTimedFrame() TimedFrame {
```
gconnell: I wouldn't export this, since it's just meant to be used internally (I think?)

gconnell:
Hey, thanks for this code!

The initial set of comments contains a bunch of stuff, some of it simple style nits (camelCase), some of it annoying legaleze (copyright notice), and some more structural (use of goroutine for readFrames).  Please feel free to push back on anything you don't agree with... this is a conversation, not a mandate :)  And thanks again for taking the time to contribute!

david415:
Thanks for the review! I'll have those changes ready soon.

bsdbpf/bsd_bpf_sniffer.go
david415: There are many configuration options the user could possibly care about... for instance zero-copy, filters etc. See here:
- https://golang.org/src/syscall/bpf_bsd.go
- https://www.freebsd.org/cgi/man.cgi?bpf%284%29

david415:
I do believe I've made all the changes you requested. I added a Close method as well.

david415:
I've got a little test ethernet program that I use to test this API. After a minor code cleanup I got it to build and run again on FreeBSD. However I wasn't able to adjust the bpf buflen on FreeBSD and I wonder if the other BSDs behave differently.

david415:
adjustable bpf buflen now works.

also... an optional timeout value can be set.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,155 @@
+// +build darwin dragonfly freebsd netbsd openbsd
```
gconnell: Little nit:  could you move the copyright to the top and put this down below it?

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,155 @@
...
+type TimedFrame struct {
+	RawFrame  []byte
+	Timestamp time.Time
+}
+
+type BPFSniffer struct {
+	sniffDeviceName string
+	bpfDeviceName string
+	fd       int
+	stopChan chan bool
```
gconnell: stopChan and readChan can now be deleted.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,155 @@
...
+	} else {
+		b.readBufLen, err = syscall.SetBpfBuflen(b.fd, b.readBufLen)
+		if err != nil {
+			panic(err)
+		}
+	}
+	b.readBuffer = make([]byte, b.readBufLen)
+
+
+	err = syscall.SetBpfInterface(b.fd, b.sniffDeviceName)
```
gconnell: As you've mentioned, there's a bunch of possible options that can be set on this fd... how about this approach:

    opts := &bsdbpf.Options{
      ... set what you'd like ...
    }
    sniffer := bsdbpf.NewBPFSniffer(interfaceName, opts)

Things like "immediate", "timeout", "promisc", etc, could all be in opts.  And we could provide a default, which we use if folks pass in nil:

    default := Options{ReadBufferLen: defaultReadBufferLen, ...}

    func NewBPFSniffer(iface string, opts *Options) {
      if opts == nil { opts = &default }
      ...
    }

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,155 @@
...
+	// forces the interface into promiscuous mode
+	err = syscall.SetBpfPromisc(b.fd, enable)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (b *BPFSniffer) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
```
gconnell: Currently, the use of readBuffer means that ReadPacketData actually doesn't work the way it does for other sniffers.

ReadPacketData should ensure that the returned buffer is owned by the caller and won't be overridden by subsequent calls to ReadPacketData.  For example, the buffer could be passed to a set of goroutines, which may take variable amounts of time to handle the data.  You can implement this by creating a new buffer when you set readBytesComsumed to zero:

    b.readBytesConsumed = 0
    b.readBuffer = make([]byte, ...)

The issue here is that rawFrame, which is returned to the user, references the shared buffer space, so a subsequent call will destroy the user's buffer by writing over its data.

If you'd like to reuse a buffer, you should implement ZeroCopyReadPacketData, see for example http://godoc.org/github.com/google/gopacket/afpacket#TPacket.ZeroCopyReadPacketData.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,155 @@
...
+	var err error
+	if b.readBytesConsumed >= b.lastReadLen {
+		b.lastReadLen, err = syscall.Read(b.fd, b.readBuffer)
+		if err != nil {
+			return nil, gopacket.CaptureInfo{}, err
```
gconnell: I'm not sure if syscall.Read guarantees that its first return value will be zero if it returns an error... to be safe here, you may want to reset b.lastReadLen to zero on an error.

bsdbpf/bsd_bpf_sniffer.go
david415: ok

bsdbpf/bsd_bpf_sniffer.go
david415: right right

bsdbpf/bsd_bpf_sniffer.go
david415: yes i like that approach.

bsdbpf/bsd_bpf_sniffer.go
david415: Yes you are right about that... good catch. I'll fix that...

BSD BPF does have zero copy capability however I am not interested in implementing that feature at this time.

david415:
OK... ReadPacketData should be fixed now.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+const wordSize = int(unsafe.Sizeof(uintptr(0)))
+
+func bpfWordAlign(x int) int {
+	return (((x) + (wordSize - 1)) &^ (wordSize - 1))
+}
+
+type Options struct {
+	bpfDeviceName    string
```
gconnell: You'll need to export (capitalize) these names, so users of your API can set them.  Also, please add a comment to each detailing what they do

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,169 @@
...
+var defaultOptions = Options{
+	bpfDeviceName:    "",
+	readBufLen:       0,
+	timeout:          nil,
+	promisc:          true,
+	immediate:        false,
+	preserveLinkAddr: true,
+}
+
+type BPFSniffer struct {
```
gconnell: Please add comments to all exported functions and structs, so appropriate godoc can be generated.  Comments should follow http://blog.golang.org/godoc-documenting-go-code format; they should start with the name of the exported object, like so, then describe what it does, what its arguments do, etc:

    // ExportedName foos the bar.
    func ExportedName(...) { }

gconnell:
Great (and really fast) work on all this!  I think once we have comments in place, we should be ready to merge this in.

david415:
What do you think of these comments?

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -0,0 +1,199 @@
...
+		b.fd, err = syscall.Open(b.options.BPFDeviceName, syscall.O_RDWR, 0)
+		if err == nil {
+			break
+		}
+	}
+}
+
+// Init is used to initialize a BPF device for promiscuous sniffing.
+// It also starts a goroutine to continuously read frames.
+func (b *BPFSniffer) Init() error {
```
gconnell: Forgot if I mentioned this earlier, but would it make sense to put this Init logic inside of NewBPFSniffer?

    func NewBPFSniffer(...) (*BPFSniffer, error) {
    }

That way we could avoid people forgetting to call init after New.

bsdbpf/bsd_bpf_sniffer.go
gconnell: Add a brief comment on ReadPacketData... this works fine:

// ReadPacketData implements gopacket.PacketDataSource.

gconnell:
Nice!  The comments about option fields are particularly good.  I'm happy to merge whenever you think we're ready to go, just let me know.

david415:
Would it be OK if we wrote a few notes about tuning the Options for different behavior and performance? I think it would be nice to link back to a BSD manpage about BPF. The reason I suggest this is because it may not be obvious that the "Immediate" flag if set to false will cause less syscall context switching to occur for N number of packets... with the trade off being slightly "reactive"; the read syscall blocks until the buffer is filled or timeout is reached. Certain applications will achieve "smoother" performance by tuning the buffer length, adjusting the timeout and setting Immediate flag to false.

david415:
I have not yet tested on dragonfly and darwin... however it is very likely that this API will works on all BSD systems that the official sys package has a BPF header struct for. It annoys me greatly that these are variable width depending on which BSD system and architecture it is. I think also if darwin and freebsd default to a bufLen of 4096 that this might possibly clip packets if larger ones are transmitted. Therefore although we should set Immediate to true as a default option because this is probably the behavior that most developers expect we should perhaps set the default bufLen to some non-zero value so that the default Options results in the same size bufLen on all BSD systems.

david415:
Good thing we use a largish default buffer size...

Haha.. the default size of bufLen in freebsd and darwin being 4096 bytes is in fact totally broken for the context of say a laptop with a wifi card and several other link layers:
https://en.wikipedia.org/wiki/Maximum_transmission_unit#Table_of_MTUs_of_common_media

What I'd like to do is provide an additional method that can be used to compute the approximate largest valid BPF buffer length. That is, a simple function with a for loop and an exponential step...

david415:
Yeah our current default bufLen is sensible... whereas max possible bufLen just doesn't make sense. I just now updated the comments and added a GetReadBufLen method. I think this is a good start and if you are OK with additional changes to the comments then we should merge it.

bsdbpf/bsd_bpf_sniffer.go
david415: oh i overlooked this comment earlier...

david415:
oh yes i think we should call Init from the constructor...

david415:
OK i moved the initialization to NEWBPFSniffer. During that change I got rid of the calls to panic and incorrect/old comments. Unless you can spot more things to change then I agree we should merge it. I think future additions to this API can easily be backwards compatible...

PR #89, opened 2015-06-25
[dot11] use SetTruncated() for too short packet

hkwi:
Captured dot11 packets are sometimes broken and too short by nature. That caused slice-bounds-out-of-range runtime error very often, and triggered builtin recovery, which cost counted.


gconnell:
Thanks!

PR #91, opened 2015-06-29
Fix segfault (ReadPacketData after Close)

DrJosh9000:
If you call Close on an open Handle and then try to read a packet, it can segfault in pcap_next_ex.
This is easy to do, for example, trying to use Handle.Close to cause the goroutine spawned by PacketSource.Packets to stop.


gconnell:
Thanks!

PR #92, opened 2015-07-03
add dhcpv4 parsing

vtolstov:
Signed-off-by: Vasiliy Tolstov <v.tolstov@selfip.ru>

### DIFF
--- a/layers/dhcpv4.go
+++ b/layers/dhcpv4.go
@@ -1,0 +1,407 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+// Copyright 2009-2011 Andreas Krennmair. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/google/gopacket"
+)
+
+var opTypes = [3]string{
+	0: "UNSPEC",
+	1: "REQUEST",
+	2: "RESPONSE",
+}
+
+type DHCPOperation uint8
+
+const (
+	DHCP_MSG_UNSPEC DHCPOperation = iota
+	DHCP_MSG_DISCOVER
+	DHCP_MSG_OFFER
+	DHCP_MSG_REQUEST
+	DHCP_MSG_DECLINE
+	DHCP_MSG_ACK
+	DHCP_MSG_NAK
+	DHCP_MSG_RELEASE
+	DHCP_MSG_INFORM
+)
+
+var messageTypes = map[DHCPOperation]string{
+	0: "UNSPEC",
+	1: "DISCOVER",
+	2: "OFFER",
+	3: "REQUEST",
+	4: "DECLINE",
+	5: "ACK",
+	6: "NAK",
+	7: "RELEASE",
+	8: "INFORM",
+}
+
+func (o DHCPOperation) String() string {
+	return opTypes[o]
+}
+
+const (
+	_ = iota
+	DHCP_MSG_REQ
+	DHCP_MSG_RES
+)
+
+//RFC 2131 "magic cooke"
+var dhcpMagic uint32 = 0x63825363
+
+type DHCPv4 struct {
+	BaseLayer
+	Operation    DHCPOperation
+	HardwareType byte
+	HardwareLen  uint8
+	HardwareOpts uint8
+	Xid          uint32
+	Secs         uint16
+	Flags        uint16
+	ClientIP     net.IP
+	YourIP       net.IP
+	ServerIP     net.IP
+	GatewayIP    net.IP
+	ClientHWAddr net.HardwareAddr
+	ServerName   []byte
+	File         []byte
+	Options      []DHCPOption
+}
+
+const (
+	DHCP_OPT_REQUEST_IP     byte = iota + 50 // 0x32, 4, net.IP
+	DHCP_OPT_LEASE_TIME                      // 0x33, 4, uint32
+	DHCP_OPT_EXT_OPTS                        // 0x34, 1, 1/2/3
+	DHCP_OPT_MESSAGE_TYPE                    // 0x35, 1, 1-7
+	DHCP_OPT_SERVER_ID                       // 0x36, 4, net.IP
+	DHCP_OPT_PARAMS_REQUEST                  // 0x37, n, []byte
+	DHCP_OPT_MESSAGE                         // 0x38, n, string
+	DHCP_OPT_MAX_DHCP_SIZE                   // 0x39, 2, uint16
+	DHCP_OPT_T1                              // 0x3a, 4, uint32
+	DHCP_OPT_T2                              // 0x3b, 4, uint32
+	DHCP_OPT_CLASS_ID                        // 0x3c, n, []byte
+	DHCP_OPT_CLIENT_ID                       // 0x3d, n >=  2, []byte
+
+)
+
+const (
+	DHCP_HW_TYPE_ETHERNET byte   = 0x01
+	DHCP_FLAG_BROADCAST   uint16 = 0x80
+)
+
+// LayerType returns gopacket.LayerTypeDHCPv4
+func (d *DHCPv4) LayerType() gopacket.LayerType { return LayerTypeDHCPv4 }
+
+func (dhcp *DHCPv4) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	dhcp.Operation = DHCPOperation(data[0])
+	dhcp.HardwareType = data[1]
+	dhcp.HardwareLen = data[2]
+	dhcp.HardwareOpts = data[3]
+	dhcp.Xid = binary.BigEndian.Uint32(data[4:8])
+	dhcp.Secs = binary.BigEndian.Uint16(data[8:10])
+	dhcp.Flags = binary.BigEndian.Uint16(data[10:12])
+	dhcp.ClientIP = net.IP(data[12:16])
+	dhcp.YourIP = net.IP(data[16:20])
+	dhcp.ServerIP = net.IP(data[20:24])
+	dhcp.GatewayIP = net.IP(data[24:28])
+	dhcp.ClientHWAddr = net.HardwareAddr(data[28 : 28+dhcp.HardwareLen])
+	dhcp.ServerName = data[44:108]
+	dhcp.File = data[108:236]
+	if binary.BigEndian.Uint32(data[236:240]) != dhcpMagic {
+		return errors.New("Bad DHCP header")
+	}
+
+	if len(data) <= 240 {
+		// DHCP Packet could have no option (??)
+		return nil
+	}
+
+	options := make([]byte, len(data)-240)
+	if binary.Read(bytes.NewBuffer(data[240:]), binary.BigEndian, &options) != nil {
+		return errors.New("failed to unmarshal options")
+	}
+
+	stop := len(options)
+	start := 0
+	for start < stop {
+		o := DHCPOption{}
+		if err := o.Unmarshal(options[start:]); err != nil {
+			return err
+		}
+		if o.Type == DHCP_OPT_END {
+			break
+		}
+		dhcp.Options = append(dhcp.Options, o)
+		start += int(o.Length) + 2
+	}
+	return nil
+}
+
+func (dhcp *DHCPv4) Len() uint16 {
+	n := uint16(240)
+	for _, o := range dhcp.Options {
+		n += uint16(o.Length) + 2
+	}
+	n += 1 // for opt end
+	return n
+}
+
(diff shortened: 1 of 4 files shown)


vtolstov:
I signed it!

vtolstov:
@gconnell can you help me to determine why test failed?

gconnell:
Re: test failure:

Test failure is here:  TestPacketDot11DataIP @ https://github.com/google/gopacket/blob/master/layers/dot11_test.go#L416

Test expects packet to contain "Payload" but packet instead contains "DHCPv4".

Also, sorry for the delays, just got back from honeymoon.

holleman:
Was coming to use this library and saw DHCPv4 was missing, was going to add it myself, but then found this.  This looks like it stalled, all that is left is this change and then the test passes:

diff --git a/layers/dot11_test.go b/layers/dot11_test.go
index e86a158..0b6c103 100644
--- a/layers/dot11_test.go
+++ b/layers/dot11_test.go
@@ -413,7 +413,7 @@ func TestPacketDot11DataIP(t *testing.T) {
        if p.ErrorLayer() != nil {
                t.Error("Failed to decode packet:", p.ErrorLayer().Error())
        }
-       checkLayers(p, []gopacket.LayerType{LayerTypeRadioTap, LayerTypeDot11, LayerTypeDot11Data, LayerTypeLLC, LayerTypeSNAP, LayerTypeIPv4, LayerTypeUDP, gopacket.LayerTypePayload},
+       checkLayers(p, []gopacket.LayerType{LayerTypeRadioTap, LayerTypeDot11, LayerTypeDot11Data, LayerTypeLLC, LayerTypeSNAP, LayerTypeIPv4, LayerTypeUDP, LayerTypeDHCPv4}, t)
 }
 func BenchmarkDecodePacketDot11DataIP(b *testing.B) {
        for i := 0; i < b.N; i++ {

@vtolstov  - would like to finish this off and get it included since you did all the work to get it to this point?

vtolstov:
Thanks, @holleman i'm try to update pr, also i want to simplify dhcp layer to remove function for creating new packets with specific type (i think this is suitable for some high level package).

vtolstov:
@gconnell can you check this pr? i'm fix tests, thanks @holleman and cleanup some things.

lhausermann:
I will review your patch ASAP and merge it if it can be merged.

layers/dhcpv4.go
```diff
@@ -0,0 +1,426 @@
...
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/google/gopacket"
+)
+
+var LayerTypeDHCPv4 = gopacket.RegisterLayerType(6768, gopacket.LayerTypeMetadata{"DHCPv4", gopacket.DecodeFunc(decodeDHCPv4)})
```
lhausermann: I would prefer we keep this inside layertypes.go as other definition.

gconnell: we should probably also renumber it to match internal gopacket numbering, now that it'll be merged.

layers/dhcpv4.go
```diff
@@ -0,0 +1,426 @@
...
+	return opTypes[o]
+}
+
+const (
+	_ = iota
+	DHCP_MSG_REQ
+	DHCP_MSG_RES
+)
+
+var dhcpMagic uint32 = 0x63825363
```
lhausermann: For code readiness would you add here as a comment some short extract of the RFC + RFC Number, you are using to extract these magic values. That always help on the long run to know where does magic numbers come from.

layers/dhcpv4.go
```diff
@@ -0,0 +1,426 @@
...
+	dhcp.ServerIP = net.IP(data[20:24])
+	dhcp.GatewayIP = net.IP(data[24:28])
+	dhcp.ClientHWAddr = net.HardwareAddr(data[28 : 28+dhcp.HardwareLen])
+	dhcp.ServerName = data[44:108]
+	dhcp.File = data[108:236]
+	if binary.BigEndian.Uint32(data[236:240]) != dhcpMagic {
+		return errors.New("Bad DHCP header")
+	}
+
+	if len(data) > 240 {
```
lhausermann: I would prefer we keep the code structure short and do the test to return ASAP. 

Let's do here something like:

if len(data) <= 240 {
 // DHCP Packet could have no option (??)
 return nil
}

layers/dhcpv4.go
```diff
@@ -0,0 +1,426 @@
...
+	dhcp := &DHCPv4{}
+	err := dhcp.DecodeFromBytes(data, p)
+	if err != nil {
+		return err
+	}
+	p.AddLayer(dhcp)
+	return p.NextDecoder(gopacket.LayerTypePayload)
+}
+
+func (p *DHCPv4) String() string {
```
lhausermann: Is it really needed ?
Why not rely on the standard generic function ?
Returning some other strings will break formating in tool like pcapdump for example

layers/dhcpv4.go
```diff
@@ -0,0 +1,426 @@
...
+		pFile = "<nil>"
+	}
+	pServerName := string(p.ServerName)
+	if pServerName == "" {
+		pServerName = "<nil>"
+	}
+	return fmt.Sprintf("op: %s, htype: %v, hlen: %v, hopts: %v, xid: %#x, secs: %v, flags: %v, ciaddr: %v, yiaddr: %v, siaddr: %v, giaddr: %v, chaddr: %s, sname: %v, file: %v, options: %v", p.Operation, p.HardwareType, p.HardwareLen, p.HardwareOpts, p.Xid, p.Secs, p.Flags, p.ClientIP, p.YourIP, p.ServerIP, p.GatewayIP, p.ClientHWAddr, pServerName, pFile, p.Options)
+}
+
+func (p *DHCPv4) SetBroadcast(broadcast bool) {
```
lhausermann: broadcast parameter is never used ?

vtolstov:
Thanks! I'm fix all commented issues.

sentryo:
That's all for me, but I'd like to have a double check on the DHCP options
parsing...
The buffer allocation and copy is not usual in GoPacket.

Le mar. 5 janv. 2016 02:16, Vasiliy Tolstov <notifications@github.com> a
écrit :

> Thanks! I'm fix all commented issues.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/gopacket/pull/92#issuecomment-168861940>.
>
-- 

Laurent Hausermann

Mob. +33 (0) 617 108 433

Sentryo - Co Founder

vtolstov:
what next i need to do to get this pr merged?

lhausermann:
Hi Vasily

I'd like you to explain why you are using a temporary buffer to read the option and next create a DHCPOption project.
I would like also @gconnell to check this before merging your great PR.

Laurent

vtolstov:
i'm fix copy allocations.

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+// Copyright 2009-2011 Andreas Krennmair. All rights reserved.
```
gconnell: New files shouldn't have the Andreas copyright... that's just for files that originally came from https://github.com/akrennmair/gopcap :)

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/google/gopacket"
+)
+
+var opTypes = [3]string{
```
gconnell: use [...]string instead of [3]string?

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	0: "UNSPEC",
+	1: "REQUEST",
+	2: "RESPONSE",
+}
+
+type DHCPOperation uint8
+
+const (
+	DHCP_MSG_UNSPEC DHCPOperation = iota
+	DHCP_MSG_DISCOVER
```
gconnell: Use CamelCase over UPPER_WITH_UNDER for constants.

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	DHCP_MSG_OFFER
+	DHCP_MSG_REQUEST
+	DHCP_MSG_DECLINE
+	DHCP_MSG_ACK
+	DHCP_MSG_NAK
+	DHCP_MSG_RELEASE
+	DHCP_MSG_INFORM
+)
+
+var messageTypes = map[DHCPOperation]string{
```
gconnell: [...]string, like line 20?

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	3: "REQUEST",
+	4: "DECLINE",
+	5: "ACK",
+	6: "NAK",
+	7: "RELEASE",
+	8: "INFORM",
+}
+
+func (o DHCPOperation) String() string {
+	return opTypes[o]
```
gconnell: check bounds here:

if int(o) > len(opTypes) { return "INVALID" }

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+const (
+	_ = iota
+	DHCP_MSG_REQ
```
gconnell: CamelCase

Instead of the tricky `_ = iota`, just say Req=1, Res=2

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	DHCP_MSG_RES
+)
+
+//RFC 2131 "magic cooke"
+var dhcpMagic uint32 = 0x63825363
```
gconnell: Export this?

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	GatewayIP    net.IP
+	ClientHWAddr net.HardwareAddr
+	ServerName   []byte
+	File         []byte
+	Options      []DHCPOption
+}
+
+const (
+	DHCP_OPT_REQUEST_IP     byte = iota + 50 // 0x32, 4, net.IP
+	DHCP_OPT_LEASE_TIME                      // 0x33, 4, uint32
```
gconnell: CamelCase, here and all other constants :)

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	ServerIP     net.IP
+	GatewayIP    net.IP
+	ClientHWAddr net.HardwareAddr
+	ServerName   []byte
+	File         []byte
+	Options      []DHCPOption
+}
+
+const (
+	DHCP_OPT_REQUEST_IP     byte = iota + 50 // 0x32, 4, net.IP
```
gconnell: instead of commenting on the value, just set the value?

```go
DHCPOptRequestIP = 0x32
DHCPOptLeaseTime = 0x33
...
```

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+	DHCP_OPT_MESSAGE:                  "Message",
+	DHCP_OPT_MAX_DHCP_SIZE:            "MaxDHCPSize",
+	DHCP_OPT_T1:                       "Timer1",
+	DHCP_OPT_T2:                       "Timer2",
+	DHCP_OPT_CLASS_ID:                 "ClassID",
+	DHCP_OPT_CLIENT_ID:                "ClientID",
+}
+
+type DHCPOption struct {
+	Type   uint8
```
gconnell: Make this an enum, so you can pretty print?  Then you can just say this:

```go
func (o DHCPOption) String() string { return fmt.Sprintf("Option(%v:%v)", o.Type, o.Data) }
```

layers/dhcpv4.go
```diff
@@ -0,0 +1,407 @@
...
+func NewDHCPOption(t uint8, data []byte) DHCPOption {
+	o := DHCPOption{Type: t}
+	if data != nil {
+		o.Data = data
+		o.Length = uint8(len(data))
+	}
+	return o
+}
+
+func (o *DHCPOption) Marshal() ([]byte, error) {
```
gconnell: It's a bit of a pain, but please add Godoc comments to all exported functions.

```go
// Marshal marshals the given option into a byte array.
func (o *DHCPOption) Marshal() {...}
```

gconnell: When marshaling, please take in a https://godoc.org/github.com/google/gopacket#SerializeBuffer and serialize to it.  This avoids having to alloc a new byte array for every DHCP option.

gconnell:
Hey, excellent patch!

A few conventions to follow:  CamelCase for constants, and docstrings for exported types/functions.  Other than that, mostly small nits/suggestions, feel free to take them or push back as you see fit.

tzneal:
@vtolstov Do you have time to work on this and if not, would you mind if I grabbed your patch and finished it up?

vtolstov:
Sorry, i don't have enough time. Please take this and finish it...

tzneal:
Thanks! Done, please see PR #200

lhausermann:
I'll close this one and we will review #200

PR #93, opened 2015-07-03
add basic openflow constants

vtolstov:
Signed-off-by: Vasiliy Tolstov v.tolstov@selfip.ru


vtolstov:
Firstly as openflow have different versions a think to add openflow main layers that parse only minimal header with version info, after that next layer type determine by openflow header revision and inner message type. Does this right?

vtolstov:
Also for openflow we need to goes in tcp to next layer (that openflow) not go to LayerTypePayload

layers/openflow.go
```diff
@@ -0,0 +1,718 @@
...
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/gopacket"
+)
+
+var LayerTypeOpenflow = gopacket.RegisterLayerType(12340, gopacket.LayerTypeMetadata{"Openflow", gopacket.DecodeFunc(decodeOFP)})
+
+const (
+	OFPMC_ADD    = 0 /* New meter. */
```
gconnell: To match with other enums in gopacket, please use CamelCase instead of CAPS_WITH_UNDERS, here and below.

Probably prefix all with "Openflow" or "OpenFlow".

gconnell:
Apologies for delay, just returned from honeymoon.

layers/openflow.go
```diff
@@ -0,0 +1,718 @@
...
+package layers
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/gopacket"
+)
+
+var LayerTypeOpenflow = gopacket.RegisterLayerType(12340, gopacket.LayerTypeMetadata{"Openflow", gopacket.DecodeFunc(decodeOFP)})
```
gconnell: move this with the rest of the RegisterLayerType calls, and use internal standard numbering, since it'll be part of gopacket.

layers/openflow.go
```diff
@@ -0,0 +1,718 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+// Copyright 2014-2015 Vasiliy Tolstov. All rights reserved.
```
gconnell: please remove second copyright line

layers/openflow.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+}
+
+func decodeOFP(data []byte, p gopacket.PacketBuilder) error {
+	//	o := &OFP{}
+	//return decodingLayerDecoder(o, data, p)
+	return nil
+}
+
+/*
```
gconnell: please remove commented code

lhausermann:
Dear Vasiliy
Would you consider fixing the issues raised by Graeme ?

Laurent

vtolstov:
@lhausermann yes, i'm try to fix it today.

vtolstov:
@gconnell can you helps me? Does it possible in gopacket create layer, that check 1 byte in packet and set next layer based on this value, but layer that does this check does not displayed anywhere (via Layers() function )
For example i'm try to decode some packet that have Ethernet, IPv4, TCP, Openflow14.

I want to write code that after TCP, check payload byte and determine openflow version and via nextLayer says that this is Openflow14 packet.

vtolstov:
@gconnell what do you think about this decoding?
TCP->OPENFLOW->OPENFLOWXX

Also what do you think about decoding message payload inside openflow packet ?

gconnell:
Take a look at how this is done with MPLS at
https://github.com/google/gopacket/blob/master/layers/mpls.go, lines
27-53.  In short, you map the OpenFlow protocol to a decoder function that
checks bytes and defers to version-specific decode code.

vtolstov:
@gconnell you mean something like in my last commit?

also does this output looks right?
```
       base_test.go:25: Checking packet layers, want [Openflow Openflow14]
        base_test.go:28:   Got layer Openflow, 8 bytes, payload of 16 bytes
        base_test.go:28:   Got layer Openflow14, 16 bytes, payload of 0 bytes
        base_test.go:30: PACKET: 16 bytes
                - Layer 1 (08 bytes) = Openflow {Contents=[..8..] Payload=[..16..] Version=5 Type=0 Length=16 Xid=950}
                - Layer 2 (16 bytes) = Openflow14       {Contents=[..16..] Payload=[] Version=5 Type=0 Length=16 Xid=950 Message=[]}

        openflow_test.go:40: &layers.Openflow14{BaseLayer:layers.BaseLayer{Contents:[]uint8{0x5, 0x0, 0x0, 0x10, 0x0, 0x0, 0x3, 0xb6, 0x0, 0x1, 0x0, 0x8, 0x0, 0x0, 0x0, 0x20}, Payload:[]uint8(nil)}, Version:0x5, Type:0x0, Length:0x10, Xid:0x3b6, Message:layers.OFP14MessageHello{Elements:[]uint8(nil)}}
```

lhausermann:
Can we fix the merge issues ?

lhausermann:
pong @vtolstov : can you rebase it to master?

vtolstov:
Sorry for delay, i'm try to add some more message types and rebase.

lhausermann:
ping @vtolstov anything do to do the rebase ?

notti:
Original repository disappeared, so I think this won't happen.

PR #94, opened 2015-07-03
Bugfix

randstr:
Please review.

### DIFF
--- a/layers/ip4.go
+++ b/layers/ip4.go
@@ -115,11 +115,8 @@ func (ip *IPv4) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeO
 	binary.BigEndian.PutUint16(bytes[6:], ip.flagsfrags())
 	bytes[8] = ip.TTL
 	bytes[9] = byte(ip.Protocol)
-	if err := checkIPv4Address(ip.SrcIP); err != nil {
-		return fmt.Errorf("invalid IPv4 src address (%s)", err)
-	}
-	if err := checkIPv4Address(ip.DstIP); err != nil {
-		return fmt.Errorf("invalid IPv4 dst address (%s)", err)
+	if err := ip.AddressTo4(); err != nil {
+		return err
 	}
 	copy(bytes[12:16], ip.SrcIP)
 	copy(bytes[16:20], ip.DstIP)
@@ -260,12 +257,30 @@ func decodeIPv4(data []byte, p gopacket.PacketBuilder) error {
 	return p.NextDecoder(ip.NextLayerType())
 }
 
-func checkIPv4Address(addr net.IP) error {
-	if len(addr) == net.IPv4len {
-		return nil
+func checkIPv4Address(addr net.IP) (net.IP, error) {
+	if c := addr.To4(); c != nil {
+		return c, nil
 	}
 	if len(addr) == net.IPv6len {
-		return fmt.Errorf("address is IPv6")
+		return nil, fmt.Errorf("address is IPv6")
 	}
-	return fmt.Errorf("wrong length of %d bytes instead of %d", len(addr), net.IPv4len)
+	return nil, fmt.Errorf("wrong length of %d bytes instead of %d", len(addr), net.IPv4len)
+}
+
+func (ip *IPv4) AddressTo4() error {
+	var src, dst net.IP
+
+	if addr, err := checkIPv4Address(ip.SrcIP); err != nil {
+		return fmt.Errorf("Invalid source IPv4 address (%s)", err)
+	} else {
+		src = addr
+	}
+	if addr, err := checkIPv4Address(ip.DstIP); err != nil {
+		return fmt.Errorf("Invalid destination IPv4 address (%s)", err)
+	} else {
+		dst = addr
+	}
+	ip.SrcIP = src
+	ip.DstIP = dst
+	return nil
 }
--- a/layers/ip4.go
+++ b/layers/ip4.go
@@ -115,11 +115,11 @@ func (ip *IPv4) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeO
 	binary.BigEndian.PutUint16(bytes[6:], ip.flagsfrags())
 	bytes[8] = ip.TTL
 	bytes[9] = byte(ip.Protocol)
-	if len(ip.SrcIP) != 4 {
-		return fmt.Errorf("invalid src IP %v", ip.SrcIP)
+	if err := checkIPv4Address(ip.SrcIP); err != nil {
+		return fmt.Errorf("invalid IPv4 src address (%s)", err)
 	}
-	if len(ip.DstIP) != 4 {
-		return fmt.Errorf("invalid dst IP %v", ip.DstIP)
+	if err := checkIPv4Address(ip.DstIP); err != nil {
+		return fmt.Errorf("invalid IPv4 dst address (%s)", err)
 	}
 	copy(bytes[12:16], ip.SrcIP)
 	copy(bytes[16:20], ip.DstIP)
@@ -258,4 +258,14 @@ func decodeIPv4(data []byte, p gopacket.PacketBuilder) error {
 		return err
 	}
 	return p.NextDecoder(ip.NextLayerType())
+}
+
+func checkIPv4Address(addr net.IP) error {
+	if len(addr) == net.IPv4len {
+		return nil
+	}
+	if len(addr) == net.IPv6len {
+		return fmt.Errorf("address is IPv6")
+	}
+	return fmt.Errorf("wrong length of %d bytes instead of %d", len(addr), net.IPv4len)
 }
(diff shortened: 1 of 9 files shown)


randstr:
I signed it!

randstr:
I signed it!

randstr:
There was a mismatch with the email address for the CLA because of a domain change, but it should be fixed now.

gconnell:
Apologies, but due to another merge, we now have a merge conflict here.  Would you resolve this and resend?

And sorry for the delay... I've just returned from my honeymoon.

randstr:
No problem, will do and resend. :)

randstr:
I rebased it onto master and did a small cleanup.

Let me know if I screwed this up somehow, I'm still learning the ropes around this workflow. Thanks.

layers/ip4.go
```diff
@@ -115,14 +115,16 @@ func (ip *IPv4) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeO
 	binary.BigEndian.PutUint16(bytes[6:], ip.flagsfrags())
 	bytes[8] = ip.TTL
 	bytes[9] = byte(ip.Protocol)
-	if len(ip.SrcIP) != 4 {
-		return fmt.Errorf("invalid src IP %v", ip.SrcIP)
+	src := ip.SrcIP.To4()
```
gconnell: Why is checking for len == 4 not enough here?  If someone sets ipv4.SrcIP to an IPv6 which maps to a v4 address, that's still wrong.  Also, To4() may do extra memory allocs (not sure, haven't looked too closely).

layers/tcpip.go
```diff
@@ -18,27 +18,41 @@ type tcpipchecksum struct {
...
 
-func (ip *IPv4) pseudoheaderChecksum() (csum uint32) {
-	csum += (uint32(ip.SrcIP[0]) + uint32(ip.SrcIP[2])) << 8
-	csum += uint32(ip.SrcIP[1]) + uint32(ip.SrcIP[3])
-	csum += (uint32(ip.DstIP[0]) + uint32(ip.DstIP[2])) << 8
-	csum += uint32(ip.DstIP[1]) + uint32(ip.DstIP[3])
-	csum += uint32(ip.Protocol)
-	return
+func (ip *IPv4) pseudoheaderChecksum() (csum uint32, err error) {
+	src := ip.SrcIP.To4()
```
gconnell: as above, just a len(SrcIP) should be sufficient here.

layers/tcpip.go
```diff
@@ -18,27 +18,41 @@ type tcpipchecksum struct {
...
-	pseudoheaderChecksum() uint32
+	pseudoheaderChecksum() (uint32, error)
 }
 
-func (ip *IPv4) pseudoheaderChecksum() (csum uint32) {
-	csum += (uint32(ip.SrcIP[0]) + uint32(ip.SrcIP[2])) << 8
-	csum += uint32(ip.SrcIP[1]) + uint32(ip.SrcIP[3])
-	csum += (uint32(ip.DstIP[0]) + uint32(ip.DstIP[2])) << 8
-	csum += uint32(ip.DstIP[1]) + uint32(ip.DstIP[3])
-	csum += uint32(ip.Protocol)
```
gconnell: Why remove the addition of the protocol here, then add it back in later?

gconnell:
Thanks a ton for this code!

I was initially confused as to exactly what was going on here, but I think I've got it now :)  So please ignore some of my earlier, now removed comments.

Could you please update the pull request description to describe what you're doing in more detail?

I'm currently still unclear on why IP length checks were replaced with To4/To16 calls, and I'm concerned about the possible memory allocations it might cause.

randstr:
Thanks for your comments. I should not have removed the explanation from the commit message for the To4()/To16() calls. I'll update the commit message.

Actually I find it strange other people are not hitting on this issue but the gist is that when building an IPv4 header using net.ParseIP() the IPv4 net.IP object it returns will have length == 16 and therefore fail the len(ip) == 4 test in SerializeTo(), unless a To4() call is made. That's the design behind the golang net package as I understand it. Many functions return IPv4-mapped IPv6 addresses, such as net.IPv4().

If the IPv4 header is always constructed using 4 byte slices the To4() is unnecessary but that's not guaranteed without some constructor method. Took me a while to understood why my IPv4 net.IPs were failing the len(ip4) == 4 test.

"If someone sets ipv4.SrcIP to an IPv6 which maps to a v4 address, that's still wrong."

Why is it wrong? What tripped me up was doing ipv4.SrcIP = net.ParseIP('1.2.3.4'). It will fail the len == 4 test. Is this usage wrong?

There is memory allocation when calling To16() on a 4 byte net.IP. That's the only case where memory is allocated.

To be honest I don't really like using net.IP because of the many ways variable length byte slices can blow up or worse fail silently in this context but given that it's there I think this is the most robust approach.

gconnell:
Ah, ParseIP... yeah, it always returns IPv6 IPs, which is pretty darn
annoying.

We currently require users to say:

ip.SrcIP = net.ParseIP("1.2.3.4").To4()

In general, I really don't like the idea of an IPv6 address on an IPv4
packet being okay.  But we could be much better with error messages.  How
would you feel about checking for len == 16, then returning an error
message somethign like: "src IP is an IPv6.  If you used ParseIP to create
it, you must call To4()" or something like that?

randstr:
I was operating under the assumption that net.IP does not distinguish between v4v6, it's just "semantics". So an IPv4-mapped IPv6 address is semantically the same as IPv4.

But I won't press the point any further. I totaly understand your concerns and it's unfortunate to litter the code with To4/To16 calls. I'll happily consider it user error and withdraw that patch. :)

Regarding the error message just a reminder to point to the length of the IP address in the message would suffice, IMO.

Do we need to check the length in the pseudoheader functions? Consider the case where a user mistakenly inserts a 4 byte address in an IPv6 header. I think that will lead to out of bound errors when serializing the transport header (happens before serializing the network header). Well, I should check anyway.

I'll resend, thanks!

randstr:
Giving it another go. :)

Please review.

gconnell:
Code looks great.  One final thing, though:  would it be possible to add a test case for a packet that would have failed with the old code?  IE: a IPv6 with an extension, along with the new (correct) checksum?

randstr:
Will do.

I had a bug, I'm rebasing the fix meanwhile, sorry.

```Diff
diff --git a/layers/ip4.go b/layers/ip4.go
index c3ef52e..e73f7bc 100644
--- a/layers/ip4.go
+++ b/layers/ip4.go
@@ -118,7 +118,7 @@ func (ip *IPv4) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeO
        if err := checkIPv4Address(ip.SrcIP); err != nil {
                return fmt.Errorf("invalid IPv4 src address (%s)", err)
        }
-       if err := checkIPv6Address(ip.DstIP); err != nil {
+       if err := checkIPv4Address(ip.DstIP); err != nil {
                return fmt.Errorf("invalid IPv4 dst address (%s)", err)
        }
        copy(bytes[12:16], ip.SrcIP)
diff --git a/layers/tcpip.go b/layers/tcpip.go
index 810404f..ad43ae7 100644
--- a/layers/tcpip.go
+++ b/layers/tcpip.go
@@ -25,7 +25,7 @@ func (ip *IPv4) pseudoheaderChecksum() (csum uint32, err error) {
        if err := checkIPv4Address(ip.SrcIP); err != nil {
                return 0, fmt.Errorf("invalid IPv4 src address (%s)", err)
        }
-       if err := checkIPv6Address(ip.DstIP); err != nil {
+       if err := checkIPv4Address(ip.DstIP); err != nil {
                return 0, fmt.Errorf("invalid IPv4 dst address (%s)", err)
        }
        csum += (uint32(ip.SrcIP[0]) + uint32(ip.SrcIP[2])) << 8
```

randstr:
What test did you have in mind exactly? There are two ways to test that I can see.

One is to serialize a packet like:

IPv6 | IPv6DestOpts | UDP

and test agains a valid pcap capture. It will fail but why it fails is far from obvious, I think, so not very useful.

The other way is to do:

u := &UDP{}
// fill UDP header
ip6 := &IPv6{}
// fill IPv6 header
ip6.NextHeader = IPProtocolIPv6Destination
u.SetNetworkLayer(ip6)
u.SerializeTo(buf, opts)
got := binary.BigEndian.Uint16(buf[6:8])
want := uint16(TrueChecksum)
if got != want {
    ...
}


More useful but still not very. What do you think?

gconnell:
How about:

Serialize IP6/DestOpts/UDP
Deserialize
Check checksum against known good value

randstr:
Sure. That works best, I agree.

randstr:
Added test/more fixes/minor cleanup.

randstr:
Have you had a chance to look at this? Thanks.

gconnell:
Merged, thanks for the work!

PR #98, opened 2015-07-23
Adding ability to encode ipv4 options

adriantam:
Currently, the serialize function for ipv4 will return error if there are any ip4 options.  This change is to allow serialize function to work with ipv4 options.

### DIFF
--- a/layers/ip4.go
+++ b/layers/ip4.go
@@ -72,18 +72,40 @@ func (i IPv4Option) String() string {
 	return fmt.Sprintf("IPv4Option(%v:%v)", i.OptionType, i.OptionData)
 }
 
+// for the current ipv4 options, return the number of bytes (including
+// padding that the options used)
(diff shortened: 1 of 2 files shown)


layers/ip4.go
```diff
@@ -72,18 +72,40 @@ func (i IPv4Option) String() string {
 	return fmt.Sprintf("IPv4Option(%v:%v)", i.OptionType, i.OptionData)
 }
 
+// for the current ipv4 options, return the number of bytes (including
+// padding that the options used)
+func (ip *IPv4) getIPv4OptionSize() uint8 {
+	optionSize := uint8(0)
+	for _, opt := range ip.Options {
+		switch opt.OptionType {
+		case 0:
```
gconnell: could also say:

case 0, 1:

fiorix:
Great, thanks!

PR #103, opened 2015-08-10
Add serialization of IPv6 Hop-by-hop extension header

randstr:
This patch is a feature improvement for IPv6 adding IPv6 hop-by-hop header serialization with TLV padding, extensive jumbogram support and some other minor improvements/fixes.

Please review.

### DIFF
--- a/layers/ip6.go
+++ b/layers/ip6.go
@@ -269,9 +269,6 @@ func decodeIPv6(data []byte, p gopacket.PacketBuilder) error {
 	p.AddLayer(ip6)
 	p.SetNetworkLayer(ip6)
 	if ip6.HopByHop != nil {
-		// TODO(gconnell):  Since HopByHop is now an integral part of the IPv6
-		// layer, should it actually be added as its own layer?  I'm leaning towards
-		// no.
 		p.AddLayer(ip6.HopByHop)
 	}
 	if err != nil {
--- a/layers/ip6.go
+++ b/layers/ip6.go
@@ -41,13 +41,140 @@ const (
 	IPv6HopByHopOptionJumbogram = 0xC2 // RFC 2675
 )
 
+const (
+	ipv6MaxPayloadLength = 65535
+)
+
+// Search for Jumbo Payload TLV in IPv6HopByHop and return (length, true) if found
+func getIPv6HopByHopJumboLength(hopopts *IPv6HopByHop) (uint32, bool, error) {
+	var tlv *IPv6HopByHopOption
+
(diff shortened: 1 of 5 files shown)


layers/ip6.go
```diff
@@ -138,21 +268,57 @@ func (i *IPv6) NextLayerType() gopacket.LayerType {
...
+	bytes[0] = uint8(i.NextHeader)
+	if opts.FixLengths {
+		i.HeaderLength = uint8((length / 8) - 1)
+	}
+	bytes[1] = uint8(i.HeaderLength)
+	return b.Bytes(), nil
+}
+
+func (i *IPv6HopByHop) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
+	buf, err := i.serialize(opts)
```
gconnell: Instead of calling serialize to get back a bytes buffer, pass the SerializeBuffer into the serialize call to minimize memory allocations and copying.

gconnell:
Great change, thanks for the code!  Just switch up your use of SerializeBuffer and I think we should be good to merge.

layers/ip6.go
randstr: The idea was to avoid reversing the order of the TLVs by prepending but now that you mention it it can be worked around by starting the tlv serialization back to front. Missed that.

Will resubmit.

randstr: One would have to precompute the serialization front to back first because of the padding... then prepend the final amount of bytes. Then copy front to back. Is it worth it? I have no problem working on it.

I think we should be strict about adhering to the supplied TLV ordering in IPv6HopByHop.Options. Do you agree?

randstr: Just in case I wasn't clear, normal layer serialization is done "back to front", i.e., payload->network (prepending), but TLV option serialization must be done "front to back, i.e., network->payload (starting on the hop-by-hop/destination options header obviously).

randstr:
Fixed. Please review.

randstr:
Hold up please, I have some tests to run.

randstr:
Done. I changed the options TLVs to use pointers, please check if you agree to that.

randstr:
By the way it's troublesome to run tests with jumbograms, the screen just fills up with the BaseLayer contents. Any ideas to fix that?

gconnell:
I'd be happy to accept changes to bytediff... it was a quick/dirty method for diffing, and it's actually not very good (it uses recursive longest-common-substring instead of longest-common-subsequence, for example).

PR #106, opened 2015-08-21
ip4defrag: fix maximum fragment offset security check

sentryo:
The maximum acceptable value for the fragment offset is expressed in
units of 8 bytes (8189), so the comparison must be done directly on the
header value, before doing the multiplication.


sentryo:
Sorry, I'm going to resubmit with a different author address to match the corporate CLA agreement we have on file.

PR #108, opened 2015-08-23
add Fd() to get BPF file descriptor

hb9cwp:
Fd returns the BPF file descriptor, as required by SetBPF() to set filter program, etc. from golang/go/src/syscall/bpf_bsd.go

### DIFF
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -200,3 +200,10 @@ func (b *BPFSniffer) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
 func (b *BPFSniffer) GetReadBufLen() int {
 	return b.options.ReadBufLen
 }
+
+// Fd returns the BPF file descriptor, as required by SetBPF() to set filter program, etc.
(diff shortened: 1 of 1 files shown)


hb9cwp:
I've just signed the CLA.

gconnell:
Would it make more sense to create a SetBPF function on the sniffer, instead of exposing the FD and making the user call out to C-code?

hb9cwp:
As pointed out in the description of the pull request, SetBPF() to set filter program, as well as similar functions are already available in Go implementation from  golang/go/src/syscall/bpf_bsd.go . They require FD as input parameter.

IMHO, the next step should be to make a BPF compiler available in bsdbpf which allows to compile filter expressions into BPF programs as input to SetBFP().

hb9cwp:
SetBPF() etc are already availble from golang/go/src/syscall/bpf_bsd.go .
However, in order to be able to use them, read-only access to FD is required.

IMHO, the next step should be to make a filter expression to BPF program compiler available whose output can be fed into SetBPF().

hb9cwp:
Reopen, because closed accidentally.

hb9cwp:
Will submit improved pull request with methods to set BPF filter program, etc.

PR #111, opened 2015-08-25
ip6: convert more option TLVs to use pointers

randstr:


gconnell:
What's the reasoning behind this chane?

randstr:
Just code consistency and I assumed passing by pointer was cheaper than passing by value (?).

PR #113, opened 2015-08-27
Use interface to represent IPv6 Hop-By-Hop/Destination options

randstr:
And other misc improvements.


layers/ip6.go
```diff
@@ -280,165 +229,191 @@ func decodeIPv6(data []byte, p gopacket.PacketBuilder) error {
 	return p.NextDecoder(ip6.NextHeader)
 }
 
-func (i *IPv6HopByHop) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
-	var bytes []byte
-	var err error
+type IPv6HeaderTLVOption interface {
```
gconnell: Originally, ipv6headertlvoption was not exported... is there a reason to export it here?

If it IS being exported, we'll need to add a comment for the interface and also for its functions.

I'm worried about the use of an interface here, since interfaces generally tend to require additional mallocs.  If an interface survives outside a function call, like in decodeIPv6HeaderTLVOption, a malloc will happen.  It looks like the old decodeIPv6HeaderTLVOption also malloc'd, but I'm not sure it would have had to.

randstr: The IPv6HeaderTLVOption interface doesn't have to be exported but it does allow users of the API to implement their own TLV types. Right? Though it is possible to use just the IPv6HeaderTLVOptionUnknown type.

It doesn't have a DecodeFromBytes() method, though I suppose it could, with a suitable type code table. Unless I'm missing something.

Dunno about the use of malloc. I feel like the code is more idiomatic Go and it shows. It doesn't treat IPv6HopByHopOption and IPv6DestinationOption as different objects. There are no BigEndian.Uint32() calls to set/get the values (for Pad and Jumbo at least). And the code to align and serialize the TLVs is also simpler and perhaps more elegant. It computes the length of the aligned/padded HopByHop/Destination header without a serialize dry run. Is that issue a showstopper?

gconnell: For gopacket, I've tried to err on the side of:

idiomatic Go - for exported functions, enums, etc.
very fast (possibly not idiomatic) Go - for actual implementation

Especially when it comes to packet decoding/encoding, where high throughput
is key, I think the ends justify the means when it comes to things like
avoiding mallocs inside the code.

I'm wondering if there might be a way to have our cake and eat it too...
maybe implement an interface, but only use it internally:

type fooIface interface {
  bar()
  blah()
}

func DoStuff(t *thing) {
  fiface := fooIface(t)
  // so long as fiface doesn't escape this function, no malloc.  And in
here, we can use the nice features of the interface.
}

Thoughts?

randstr: It appears to me that the interface must escape such a function because of IPv6AlignHeaderTLVOptions(options *[]IPv6HeaderTLVOption)? Or ipv6AlignHeaderTLVOptions(), same thing...

My comment on being idiomatic reflected more on the serialization code I wrote than anything else. I was working within the constraints of the existing API (which is excellent by the way) but this small change really ties it in for me.

My personal opinion (and it's just that) is that it's not worth trading code complexity (and possibly some functionality) for an unquantified efficiency concern. An extra malloc call seems minor, if I understand correctly. But I do understand the concern, of course.

I'm also not too cognizant on how the go compiler or the garbage collector works under the hood, altough I can make an educated guess sometimes.

Let me know, I'll work with it...

Anything you'd like to see documented? Is it just code comments or something else?

randstr: > It appears to me that the interface must escape such a function because of IPv6AlignHeaderTLVOptions(options *[]IPv6HeaderTLVOption)? Or ipv6AlignHeaderTLVOptions(), same thing...

Bad example, I think, but yeah, decodeIPv6HeaderTLVOption()...

randstr:
I'm pushing the rest of my tree and we can go over it all if you agree.

PR #115, opened 2015-08-28
httpassembly example with bsdbpf 

hb9cwp:


examples/httpassembly/main_bpf.go
```diff
@@ -0,0 +1,214 @@
+// +build darwin dragonfly freebsd linux netbsd openbsd solaris
```
gconnell: It looks like 'linux' is specified both here and in main_pcap?

examples/httpassembly/main_pcap.go
```diff
@@ -1,3 +1,5 @@
+// +build linux
```
gconnell: We probably want to add windows here, too, since they use winpcap.

examples/httpassembly/main_bpf.go
```diff
@@ -0,0 +1,214 @@
...
+	assembler := tcpassembly.NewAssembler(streamPool)
+
+	log.Println("reading in packets")
+	// Read in packets, pass to assembler.
+	//packetSource := gopacket.NewPacketSource(handle, handle.LinkType())
+	packetSource := gopacket.NewPacketSource(handle, layers.LayerTypeEthernet)
+	packets := packetSource.Packets()
+	ticker := time.Tick(time.Minute)
+
+	for {
```
gconnell: Can we factor a bunch of this logic out into a third file, then call process(packetSoruce) once we have the packet source?

lhausermann:
I think this should use the new reassembly layer. Closing it.

Issue #112, opened 2015-08-27
Handling chunked HTTP responses


vijayendrabvs:
Hi,

It looks like the http assembly in gopkt is reading in an EOF at the end of the first chunk of a chunked HTTP response, and the TCP assembly underneath never passes on the rest of the chunks to the http assembler. Is there a way to be able to read the rest of the chunked HTTP response and assemble the complete HTTP response?

Thanks,
Regards,
Vijay

sentryo:
Do you have an example PCAP capture file that shows the problem? Thanks.

vijayendrabvs:
Hi @sentryo - I will set one up using a reproducible env in a day ping back. Thanks for the prompt response!

vijayendrabvs:
@sentryo : I ran a modified version of the http assembly program that comes with gopacket, with a pcap file as you suggested, and interestingly, it works with it. However, the same program does not work in the same way when listening on an interface. This should not be the case, should it? I am attaching both the sample pcap file (which I generated using tcpdump -ni <interface> tcp -w chunked.pcap` ) and the gopkt program to capture and assemble the http streams, for your reference.

Here's how I run the program on the chunked.pcap file -

`go run ./main.go -r ./chunked.pcap -f tcp`

To replicate this, you could run a web server that returns chunked HTTP responses, issue a curl to it, while running main.go on it.

PS - Github doesn't let me upload the files with their default names, so I've named them both .png files. Please rename main.png to main.go and chunked.png to chunked.pcap.
![chunked](https://cloud.githubusercontent.com/assets/1093825/9620061/5ea8ed12-50cb-11e5-8130-07d2bfb88824.png)
![main](https://cloud.githubusercontent.com/assets/1093825/9620062/5ea98614-50cb-11e5-9d7b-6a616cef780a.png)

vijayendrabvs:
@sentryo : I ran a modified version of the http assembly program that comes with gopacket, with a pcap file as you suggested, and interestingly, it works with it. However, the same program does not work in the same way when listening on an interface. This should not be the case, should it? I am attaching both the sample pcap file (which I generated using `tcpdump -ni tcp -w chunked.pcap` ) and the gopkt program to capture and assemble the http streams, for your reference.

Here's how I run the program on the chunked.pcap file -

`go run ./main.go -r ./chunked.pcap -f tcp`

To replicate this, you could run a web server that returns chunked HTTP responses, issue a curl to it, while running main.go on it.

[...]

PR #116, opened 2015-09-01
[RFC] Add new reassembly package and accompanying reassemblydump example

sentryo:
This is the current state of our new reassembly package which does full
(IPv4 and TCP) bidirectional reassembly, implements a TCP state machine
and various security checks on TCP traffic. It is based on the existing
tcpassembly package. A new reassemblydump program is provided which uses
the reassembly package, with support for DNS and HTTP reassembly.

The code is probably not fit for merging as-is, however all comments and
feedback are welcome at this stage.

The old tcpassembly package and its example programs are removed as the
reassemblydump tool supersedes them.


gconnell:
My most important comment right now is that we can't (and shouldn't) remove tcpassembly and its examples.  We can mark it deprecated, but removing it is a backwards-incompatible breakage which will break all current users, and we really don't want to do that.

Having both around and marking tcpassembly as deprecated seems like it shouldn't be too bad.  Thoughts?

reassembly/memory.go
```diff
@@ -0,0 +1,249 @@
...
+		min = initialAllocSize
+	}
+	if len(c.free) <= min {
+		return
+	}
+	for i, _ := range c.free[min:] {
+		c.free[min+i] = nil
+	}
+	c.size -= len(c.free) - min
+	c.free = c.free[:min]
```
gconnell: note:  this free doesn't actually release anything to GC, as far as I know... Go still considers the underlying array and its elements reachable, so it won't GC them.  They're reachable because of http://play.golang.org/p/OpM_EFZria

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+// The reassembly package implements bi-directional TCP reassembly, for use in
+// packet-sniffing applications.  The caller reads packets off the wire, then
+// presents them to an Assembler in the form of gopacket layers.TCP packets
+// (github.com/google/gopacket, github.com/google/gopacket/layers).
+//
+// The Assembler uses a user-supplied
+// StreamFactory to create a user-defined Stream interface, then passes packet
+// data in stream order to that object.  A concurrency-safe StreamPool keeps
+// track of all current Streams being reassembled, so multiple Assemblers may
+// run at once to assemble packets while taking advantage of multiple cores.
```
gconnell: maybe expand this comment to include "world's simplest reassembly example"?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+func (s Sequence) Add(t int) Sequence {
+	return (s + Sequence(t)) & uint32Max
+}
+
+type TCPAssemblyStats struct {
+	// For this ScatterGather
+	NbChunks  uint
+	NbPackets uint
+	// For the half connection, since last call to ReassembledSG()
+	NbQueuedBytes    uint
```
gconnell: what's "Nb" mean?  Maybe drop it?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+	// Returns the length of available bytes and saved bytes
+	Lengths() (int, int)
+	// Returns the bytes up to length
+	Fetch(length int) []byte
+	// Tell to keep from offset
+	KeepFrom(offset int)
+	// Return CaptureInfo of packet corresponding to given offset
+	CaptureInfo(offset int) gopacket.CaptureInfo
+	// Return some info about the reassembled chunks
+	Info() (direction TcpFlowDirection, start bool, end bool, skip int)
```
gconnell: Looks like TcpFlowDirection uses different initial caps than TCPAssemblyStats and other TCP-related structs.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+	AssemblerContext() AssemblerContext
+	Release(*pageCache) int
+	Start() bool
+	End() bool
+	Seq() Sequence
+	IsPacket() bool
+}
+
+// Implements a ScatterGather
+type ReassemblyObject struct {
```
gconnell: should this be exported?  If it's being passed as a ScatterGather, there doesn't seem to be a need.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+}
+
+// Add adds an integer to a sequence and returns the resulting sequence.
+func (s Sequence) Add(t int) Sequence {
+	return (s + Sequence(t)) & uint32Max
+}
+
+type TCPAssemblyStats struct {
+	// For this ScatterGather
+	NbChunks  uint
```
gconnell: maybe return these as 'int'?  While uint is more correct, it makes comparisons against len(foo) etc. more difficult.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+}
+
+const pageBytes = 1900
+
+// TCP flow direction
+type TcpFlowDirection bool
+
+const (
+	TcpDirServerToClient TcpFlowDirection = false
+	TcpDirClientToServer TcpFlowDirection = true
```
gconnell: what happens if direction is unknown?  how does reassembly handle streams it starts seeing halfway through?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+}
+func (p *page) IsPacket() bool {
+	return p.isPacket
+}
+func (p *page) String() string {
+	return fmt.Sprintf("page%p{seq: %v, Bytes:%d, -> nextSeq:%v} (prev:%p, next:%p)", p, p.seq, len(p.bytes), p.seq+Sequence(len(p.bytes)), p.prev, p.next)
+}
+
+/* LivePacket: implements a ByteContainer */
+type LivePacket struct {
```
gconnell: needs to be exported, or is it just passed as a ByteContainer?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+// TCP data.  Callers create a StreamFactory, then StreamPool uses
+// it to create a new Stream for every TCP stream.
+//
+// assembly will, in order:
+//    1) Create the stream via StreamFactory.New
+//    2) Call ReassembledSG 0 or more times, passing in reassembled TCP data in order
+//    3) Call ReassemblyComplete one time, after which the stream is dereferenced by assembly.
+type Stream interface {
+	// Tell whether the tcp packet should be accepted, start could be modified to force a start even if no SYN have been seen
+	Accept(tcp *layers.TCP, ci gopacket.CaptureInfo, dir TcpFlowDirection, ackSeq Sequence, start *bool, ac AssemblerContext) bool
```
gconnell: if no SYN is seen, what's the value of 'dir'?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+// can be handled without any locking.
+//
+// NOTE:  If you can guarantee that packets going to a set of Assemblers will
+// contain information on different connections per Assembler (for example,
+// they're already hashed by PF_RING hashing or some other hashing mechanism),
+// then we recommend you use a seperate StreamPool per Assembler, thus
+// avoiding all lock contention.  Only when different Assemblers could receive
+// packets for the same Stream should a StreamPool be shared between them.
+//
+// Avoids Memory Copying
```
gconnell: Test this with http://golang.org/pkg/testing/#AllocsPerRun ?

gconnell: I'm worried that this isn't actually the case, due to the use of interfaces.  I could very well be wrong, thought.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+	pool.mu.Unlock()
+	return &Assembler{
+		ret:              make([]ByteContainer, assemblerReturnValueInitialSize),
+		pc:               newPageCache(),
+		connPool:         pool,
+		AssemblerOptions: DefaultAssemblerOptions,
+	}
+}
+
+func (a *Assembler) Dump() string {
```
gconnell: Add comments to exported funcs.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+	// ReassembledSG is called zero or more times.
+	// Reassembly objects are reused after each Reassembled call,
+	// so it's important to copy anything you need out of them
+	// or to not call Release() to keep those bytes
+	ReassembledSG(sg ScatterGather, ac AssemblerContext)
+	// ReassemblyComplete is called when assembly decides there is
+	// no more data for this Stream, either because a FIN or RST packet
+	// was seen, or because the stream has timed out without any new
+	// packet data (due to a call to FlushCloseOlderThan).
+	// It returns true if connection should be removed from the pool
```
gconnell: what if it returns false?  does that mean it will still wait for packets?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1227 @@
...
+		}
+	}
+	if !half.closed && half.first == nil && half.lastSeen.Before(tc) {
+		a.closeHalfConnection(conn, half)
+		closed = true
+	}
+	return flushed, closed
+}
+
+// FlushCloseAll flushes all remaining data into all remaining connections and closes
```
gconnell: name doesn't match comment.

gconnell:
And, lest I forget to say this earlier, thanks so much to all involved for this work.  This is a REALLY excellent effort, and I'm amazingly pleased that my horribly hacky TCP assembly code is being replaced by something really solid.  Kudos to all who worked on this... it's really exciting.

lhausermann:
Thanks Graeme for your congratulations. At Sentryo we really love GoPacket and we think we should give some value back to the GoPacket community. We have more to come and as already said, we are willing to participate in any OSS project around Gopacket. I will forward to all the team your last comment !

reassembly/tcpassembly.go
orebokech: I guess it's short for "number", and I agree that it's superfluous. I'll drop it for the next round.

reassembly/tcpassembly.go
orebokech: Fixed.

reassembly/tcpassembly.go
orebokech: Yes, in that case we keep the stream around to allow tracking the end of the connection (the TCP LAST-ACK state).

reassembly/tcpassembly.go
orebokech: If we start seeing the connection mid-stream then the direction doesn't really matter because we have no way of telling which is the client and which is the server anyway. In our application which uses the package we have some heuristics to try to guess the correct direction based on ports and upper-layer contents...

reassembly/tcpassembly.go
orebokech: Then the direction has a 50% chance of being correct depending on which packet creates the stream.

reassembly/tcpassembly.go
orebokech: Ok, I changed the types for the next round, the difference in range is not particularly relevant and it gets rid of a few casts.

orebokech:
Thanks for the review, Graeme.

The rationale for the removal of the old tcpassembly package was that we were under the impression that it had no external users given the low level of activity, and since the new package is based on it the removal would avoid some code duplication.

But if you think it's important to keep providing the old package, we can keep it. What of its example programs, should we keep them as well?

gconnell:
I actually spoke to a few folks at GopherCon this year who were using tcpassembly, so there are some uses out there.  As far as the examples go, maybe keep them around for a bit?  Let's comment that tcpassembly is deprecated inside them.

Or, we could port them to use reassembly.  While reassemblydump looks great and does everything, I think it suffers from doing everything.  Maybe having httpassembly and statsassembly use it, but do the barest minimum to get things off the ground, would allow users to have a better starting point.  Thoughts?

gconnell:
Any progress on this?

sentryo:
Since the initial submission we have collected a set of fixes based on our production usage and we plan to submit a new version which addresses your previous comments and includes these fixes soon, but we've been very busy lately.... sorry for the delay.

lhausermann:
And by the way, we would be happy to hire a freelancer to help submit a mergeable version. Anyone interested should contact me directly.

postwait:
If you are interested in hiring someone to help merge and putback, try http://omniti.com/ they might be able to help with that.

Can you point me to the branch that has the latest version with all the fixes you mentioned?

tcz001:
@lhausermann So what's the state of tcpassembly? I can help if time is an issue here.

lhausermann:
We are planning to release an updated and production proof code in October.

xixiliguo:
@lhausermann updated tcpassembly is still not ready ?

carwyn:
@sentryo @lhausermann is the latest version with fixes available somewhere? Keen to test this as I think I'm seeing the same issues as #112

sentryo:
The branch has (finally) been updated with the current version of our code, rebased on master. I believe we've addressed all Graeme's comments, and left the old tcpassembly package for now.

Reviews are welcome, and we can hopefully merge this soon! Thanks.

decode.go
```diff
@@ -16,11 +16,15 @@ type DecodeFeedback interface {
 	// is shorter than internal layer variables (HeaderLength, or the like) say it
 	// should be.  It sets packet.Metadata().Truncated.
 	SetTruncated()
+	// SetConsumed should be called when all payload has not been consumed
```
gconnell: I'm not clear on what "consumed" means here, and it appears that SetConsumed is never actually called?

layers/tcp.go
```diff
@@ -299,9 +307,21 @@ func decodeTCP(data []byte, p gopacket.PacketBuilder) error {
 	if err != nil {
 		return err
 	}
-	return p.NextDecoder(gopacket.LayerTypePayload)
+	if p.DecodeApplicationLayers() {
+		return p.NextDecoder(tcp.NextLayerType())
```
gconnell: Does it make sense to attempt decoding of a TCP payload?  I'd assume that with its streaming nature, you have no guarantees of packet boundaries in TCP, meaning this could often fail.

I'm okay with doing this in UDP, since it's datagram-based, but for TCP I think it's still best to just use Payload, then rely on reassembly elsewhere for application-level stuff.  Otherwise, callers may expect magic from this library that it's not prepared to give.

Also, DecodeApplicationLayers() is somewhat strange of a name, since UDP application layers (at least DNS and one or two others) are decoded regardless.

layers/tcp.go
```diff
@@ -299,9 +307,21 @@ func decodeTCP(data []byte, p gopacket.PacketBuilder) error {
...
+		return p.NextDecoder(gopacket.LayerTypePayload)
+	}
 }
 
 func (t *TCP) TransportFlow() gopacket.Flow {
 	return gopacket.NewFlow(EndpointTCPPort, t.sPort, t.dPort)
 }
+
+// For testing only
+func (t *TCP) SetInternalPorts() {
```
gconnell: rename: SetInternalPortsForTesting

reassembly/cap2test.go
```diff
@@ -0,0 +1,107 @@
...
+	"strings"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcap"
+)
+
+var input = flag.String("i", "", "Input filename")
+
+//var output = flag.String("output", "", "Output filename")
```
gconnell: remove comment?

reassembly/cap2test.go
```diff
@@ -0,0 +1,107 @@
...
+			b.WriteString("}},\n")
+			b.WriteString("},\n")
+			// CaptureInfo
+			b.WriteString("ci: gopacket.CaptureInfo{\n")
+			ts := packet.Metadata().CaptureInfo.Timestamp
+			b.WriteString(fmt.Sprintf("  Timestamp: time.Unix(%d,%d),\n", ts.Unix(), ts.Nanosecond()))
+			b.WriteString("},\n")
+			// Struct
+			b.WriteString("},\n")
+			fmt.Print(b.String())
```
gconnell: any reason not to just do this?:
```go
fmt.Printf("%v %+v\n",
  gopacket.LayerDump(tcp),
  packet.Metadata().CaptureInfo)
```

reassembly/memory.go
```diff
@@ -0,0 +1,251 @@
...
+		if c.pageRequests&0xFFFF == 0 {
+			log.Println("PageCache:", c.pageRequests, "requested,", c.used, "used,", len(c.free), "free")
+		}
+	}
+	if len(c.free) == 0 {
+		c.grow()
+	}
+	i := len(c.free) - 1
+	p, c.free = c.free[i], c.free[:i]
+	p.prev = nil
```
gconnell: I believe it's actually better to clear this on replace() than on next(), since it could allow for next/prev values to be garbage collected (IE: if next or prev references a page that's deref'd by tryShrink())

reassembly/memory.go
```diff
@@ -0,0 +1,251 @@
...
+	p.next = nil
+	p.seen = ts
+	p.bytes = p.buf[:0]
+	c.used++
+	if *memLog {
+		fmt.Printf("allocator returns %s\n", p)
```
gconnell: for logging debug stuff, consider using the standard 'log' package instead of fmt, since it gives some additional information on the log lines, goes to stderr, and is otherwise nice :)

reassembly/memory.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if *memLog {
+		log.Println("StreamPool: created", p.nextAlloc, "new connections")
+	}
+	p.nextAlloc *= 2
+}
+
+// Dump logs all connections
+func (p *StreamPool) Dump() {
+	log.Printf("Remaining %d connections: ", len(p.conns))
```
gconnell: p.mu.Lock() while dumping

reassembly/memory.go
```diff
@@ -0,0 +1,251 @@
...
+	if len(p.free) == 0 {
+		p.grow()
+	}
+	index := len(p.free) - 1
+	c, p.free = p.free[index], p.free[:index]
+	c.reset(k, s, ts)
+	return c, &c.c2s, &c.s2c
+}
+
+// getConnection returns a connection.  If end is true and a connection
```
gconnell: Move this comment down to its associated function.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+//
+// assembly will, in order:
+//    1) Create the stream via StreamFactory.New
+//    2) Call ReassembledSG 0 or more times, passing in reassembled TCP data in order
+//    3) Call ReassemblyComplete one time, after which the stream is dereferenced by assembly.
+type Stream interface {
+	// Tell whether the TCP packet should be accepted, start could be modified to force a start even if no SYN have been seen
+	Accept(tcp *layers.TCP, ci gopacket.CaptureInfo, dir TCPFlowDirection, ackSeq Sequence, start *bool, ac AssemblerContext) bool
+	// ReassembledSG is called zero or more times.
+	// Reassembly objects are reused after each Reassembled call,
```
gconnell: Update comment to talk about ScatterGather/AssemblerContext instead of Reassembly obects.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+// Dump returns a short string describing the page usage of the Assembler
+func (a *Assembler) Dump() string {
+	s := ""
+	s += fmt.Sprintf("pageCache: used: %d, size: %d, free: %d", a.pc.used, a.pc.size, len(a.pc.free))
+	return s
+}
+
+// AssemblerContext provides method to get metadata
+type AssemblerContext interface {
+	GetCaptureInfo() gopacket.CaptureInfo
```
gconnell: I like this use of AssemblerContext... really elegant way to pass cature info around.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+}
+
+// Implements AssemblerContext for Assemble()
+type assemblerSimpleContext struct {
```
gconnell: btw, I think you could also say:

```go
type assmeblerSimpleContext gopacket.CaptureInfo
func (a ...) GetCaptureInfo() { return gopacket.CaptureInfo(a) }
```

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+	QueuedPackets  int
+	OverlapBytes   int
+	OverlapPackets int
+}
+
+// ScatterGather is used to pass reassembled data and metadata of reassembled
+// packets to a Stream via ReassembledSG
+type ScatterGather interface {
+	// Returns the length of available bytes and saved bytes
+	Lengths() (int, int)
```
gconnell: might be simpler to have AvailableBytes() and SavedBytes() functions?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+	// Returns the bytes up to length
+	Fetch(length int) []byte
```
gconnell: up to available?  or saved?

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+		l += r.Length()
+	}
+	return l, rl.saved
+}
+
+func (rl *reassemblyObject) Fetch(l int) []byte {
+	if len(rl.all) == 1 && l == rl.all[0].Length() {
+		return rl.all[0].Bytes()
+	}
+	bytes := []byte{}
```
gconnell: Maybe:  `bytes := make([]byte, 0, l)`

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+func (rl *reassemblyObject) Lengths() (int, int) {
+	l := 0
+	for _, r := range rl.all {
+		l += r.Length()
+	}
+	return l, rl.saved
+}
+
+func (rl *reassemblyObject) Fetch(l int) []byte {
+	if len(rl.all) == 1 && l == rl.all[0].Length() {
```
gconnell: I think you can do this if `l <= r1.all[0].Length()`, by adding `[:l]` to the end of line 145.

decode.go
sentryo: Ok, the comment was updated.
This is used in our private layers which decode TCP protocols, see next comment.

layers/tcp.go
sentryo: Yes, we can rename the function to something more meaningful. Suggestions?

As we see it, for some protocols it does absolutely make sense to use gopacket's existing framework to decode TCP-based protocols in gopacket layers. We primarily deal with industrial protocols which were originally transported over serial lines and are correctly "packetized". This is the purpose of SetConsumed(): to tell the application of much of the payload has already been parsed by the layer, and not resend the same bytes when more data arrives. See also DNS over TCP, which is not stream-based.

For some protocols it is obviously better to use a stream interface, e.g. for HTTP or FTP, and that is what we use. But we see value in both modes, and most of our TCP dissectors use the "packet" mode.

layers/tcp.go
sentryo: Done.

reassembly/cap2test.go
sentryo: Done.

reassembly/memory.go
sentryo: Done.

reassembly/memory.go
sentryo: Done.

reassembly/tcpassembly.go
sentryo: Simpler yes, but since some effort has been put for performance, we wanted to avoid two function calls.

reassembly/tcpassembly.go
sentryo: Done.

reassembly/tcpassembly.go
sentryo: Yep, done.

reassembly/tcpassembly.go
sentryo: :)

reassembly/tcpassembly.go
sentryo: Done.

reassembly/tcpcheck.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	return nil
+}
+
+// TCPSimpleFSM implements a very simple TCP state machine
+//
+// Limitations:
+// - packet should be received in-order.
+// - no check on sequence number is performed
+// - no RST
```
gconnell: Add comment here on when/why a user would want to use this?  I'm unclear as to why I would, and whether it's necessary to export it vs. just keeping it internal to the subpackage.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+// otherwise it will wait until the next FlushCloseOlderThan to see if bytes
+// [25-30) come in.
+//
+// If it pushes all bytes (or there were no sets of bytes to begin with)
+// AND the connection has not received any bytes since the passed-in time,
+// the connection will be closed.
+//
+// Returns the number of connections flushed, and of those, the number closed
+// because of the flush.
+func (a *Assembler) FlushCloseOlderThan(t time.Time, tc time.Time) (flushed, closed int) {
```
gconnell: Note some recent changes to tcpassembly (https://godoc.org/github.com/google/gopacket/tcpassembly#Assembler.FlushWithOptions) have shown it's possibly much nicer to pass in an Options struct that allows users to, for example, flush all existing streams without closing them, etc.  I'd recommend that approach here as well.

decode.go
```diff
@@ -63,6 +67,10 @@ type PacketBuilder interface {
 	// data will be dumped to stderr so you can create a test.  This should never
 	// be called from a production decoder.
 	DumpPacketData()
+	// DecodeApplicationLayers returns true if the packet was
+	// created with a decode option to enable application-level
+	// decoding (currently only checked by the TCP decoder).
+	DecodeApplicationLayers() bool
```
gconnell: Let's instead do:

```go
DecodeOptions() *DecodeOptions
```

That way, as other options are added to DecodeOptions in the future, additional functions aren't necessary here.

decode.go
```diff
@@ -16,11 +16,15 @@ type DecodeFeedback interface {
 	// is shorter than internal layer variables (HeaderLength, or the like) say it
 	// should be.  It sets packet.Metadata().Truncated.
 	SetTruncated()
+	// SetConsumed should be called for streams to mention that only a part of
+	// the payload has been consumed (currently useful for TCP only).
+	SetConsumed(uint)
```
gconnell: I'm not seeing anywhere this is currently in use... could we hold off on the whole `Consumed` thing until we actually code in a use for it?

sentryo:
Thanks for the review! We'll get back to you ASAP.
We're all for merging the code into master if you deem it good enough after this last round. It would, in fact, be a great achievement :)

Regarding this account: the wizard behind the curtain is @rfrancoise but the code was authored by several people, the copyright belongs to Sentryo (the company) and we signed the corresponding CLA.

decode.go
gconnell: Or, optionally, something like:

if tcp := packet.Layer(LayerTypeTCP); tcp != nil { tcp.Consumed = x }

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+	Info() (direction TCPFlowDirection, start bool, end bool, skip int)
+	// Return some stats regarding the state of the stream
+	Stats() TCPAssemblyStats
+}
+
+// byteContainer is either a page or a livePacket
+type byteContainer interface {
+	Bytes() []byte
+	Length() int
+	ConvertToPages(*pageCache, int, AssemblerContext) (*page, *page, int)
```
sentryo: Ok, fixed.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+
+const pageBytes = 1900
+
+// TCPFlowDirection distinguish the two half-connections directions.
+//
+// TCPDirClientToServer is assigned to half-connection for the first received
+// packet, hence might be wrong if packets are not received in order.
+// It's up to the caller (e.g. in Accept()) to decide if the direction should
+// be interpretted differently.
+type TCPFlowDirection bool
```
sentryo: It's just an enum for the direction of the half-stream: we need to have exactly two different directions. We could change here the name to Dir1 and Dir2 for instance if the Client/Server names are too misleading. We can also add info (here ternary) in ScaterGather.Info() telling whether Dir1 is Client→Server or Server→Client or unknown.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+type page struct {
+	bytes []byte
+	seq   Sequence
+	//index      int
+	prev, next *page
+	buf        [pageBytes]byte
+	ac         AssemblerContext
+	seen       time.Time
+	start, end bool
+	isPacket   bool // only the first page of a packet has this being set
```
sentryo: It is just used to count the number of TCP packets for a given chunk given to the caller, changing to a pointer seems a bit overkill.
Using ac as boolean is doable: Done.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+
+type connection struct {
+	key      key // client->server
+	c2s, s2c halfconnection
+	mu       sync.Mutex
+}
+
+func (c *connection) reset(k key, s Stream, ts time.Time) {
+	c.key = k
+	c.c2s.pages = 0
```
sentryo: Done: clearer + fixed lastSeen missing for c2s.

reassembly/tcpassembly.go
```diff
@@ -0,0 +1,1319 @@
...
+	c.s2c.pages = 0
+	c.c2s.first, c.c2s.last = nil, nil
+	c.s2c.first, c.s2c.last = nil, nil
+	c.c2s.nextSeq = invalidSequence
+	c.s2c.nextSeq = invalidSequence
+	c.c2s.ackSeq = invalidSequence
+	c.s2c.ackSeq = invalidSequence
+	c.c2s.created = ts
+	c.s2c.created = ts
+	c.s2c.lastSeen = ts
```
sentryo: Fixed. Well spotted!

reassembly/tcpassembly.go
sentryo: Done: the bool has been replaced by a value, to allow different values.

reassembly/tcpcheck.go
sentryo: Done.

reassembly/tcpassembly.go
sentryo: It was hidden in the small TODO on the top:"- implement chunked (cheap) reads and Reader() interface".

layers/tcp.go
sentryo: Ok, renamed in the last push.

decode.go
sentryo: I think the best way forward would be for us to provide a sample TCP layer which decodes in "packet mode" and uses this to notify if there's not enough data. Let us prepare something and get back to you.

emfree:
This looks like a nice addition to a great library! I'd love to have it available in mainline gopacket -- are there any obstacles to merging? Huge thanks to all the authors as well!

gconnell:
Got back from xmas leave a week ago, and mostly caught up on real work (read: email) since then.  Is this ready for another look?

decode.go
gconnell: looks like this is still outstanding... is this blocking another review?

carwyn:
FYI this latest sentryo:reassembly branch is working really well for me on long running http captures, albeit low bandwidth, with no issues. (Fedora 25 and OS X)

lhausermann:
@gconnell the team would like to add a real case for 'set consumed' API we think it's needed for some case. We'd like to do it for DNS and next envision some 'binary payload' which are not really stream based like SNMP for example.

sentryo:
We'd like to get this code merged, so for now we have dropped the new SetConsumed() API, and we'll resubmit it later when we have a real-world example in one of the layers.

Thanks!

sentryo:
Also, if this is fit for merging, let us know if you want us to clean up the branch history beforehand.

gconnell:
Awesome work again, guys!  Merging in.

Keeping history around is always nice :)

gconnell:
Bookkeeping-wise:

I've made Laurent a project admin, and added sentryo as a committer.  If there's any individual accounts for the sentryo folks you'd like added in as committers, please reach out and I'll be happy to do so.

sentryo:
Thanks Graeme! It's been a long road, but we finally got there with your invaluable help. :)
Also thanks for granting us push access, we appreciate your trust! We'll use it wisely.

lhausermann:
Thanks Graeme, our team here is very happy. That patch will be the foundation of upcoming project. Gopacket followers hope you appreciate !

PR #117, opened 2015-09-01
Fix decoding of TXT records

orivej:
[RFC1035](https://www.ietf.org/rfc/rfc1035.txt) defines `<character-string>` as a length prefixed octet array:

> \<character-string\> is a single length octet followed by that number of characters.  \<character-string\> is treated as binary information, and can be up to 256 characters in length (including the length octet).

and uses it in definition of TXT, which is "One or more \<character-string\>s", and HINFO, which is a CPU \<character-string\> and an OS \<character-string\>.

`DNSResourceRecord.TXT` stores the array of character-strings unparsed. My patch leaves it as is for compatibility, and introduces `DNSResourceRecord.TXTs` with parsed records.


layers/dns.go
```diff
@@ -464,16 +465,32 @@ func (rr *DNSResourceRecord) String() string {
 	return "..."
 }
 
+func decodeCharacterStrings(data []byte) ([][]byte, error) {
+	var strings [][]byte
+	end := len(data)
+	for index, index2 := 0, 0; index != end; index = index2 {
+		index2 = index + 1 + int(data[index])
```
gconnell: maybe add quick comment here: Since index2 is an int{32,64} and data[index] is a uint8, there's no way we can overflow and have index2 < start.

layers/dns.go
```diff
@@ -464,16 +465,32 @@ func (rr *DNSResourceRecord) String() string {
...
+		if index2 > end {
+			return nil, errors.New("Insufficient data for a <character-string>")
+		}
+		strings = append(strings, data[index+1:index2])
```
gconnell: Are TXT records guaranteed to be at least one char?  If so, maybe check for that here?

layers/dns.go
```diff
@@ -464,16 +465,32 @@ func (rr *DNSResourceRecord) String() string {
 	return "..."
 }
 
+func decodeCharacterStrings(data []byte) ([][]byte, error) {
+	var strings [][]byte
```
gconnell: not sure I like the 3-4 mallocs this could cause as the slice grows... maybe start with:

strings := make([][]byte, 0, 4)

so that for most cases we only get one?  None would be better, but what can you do?

gconnell:
Thanks a ton for the code!  Just a few minor comments, hopefully we can get this turned around quickly and into the codebase.

layers/dns.go
```diff
@@ -431,6 +431,7 @@ type DNSResourceRecord struct {
 	// RDATA Decoded Values
 	IP                  net.IP
 	NS, CNAME, PTR, TXT []byte
+	TXTs                [][]byte
```
gconnell: Add comment here differentiating TXT vs. TXTs.

layers/dns.go
orivej: `decodeCharacterStrings` will be called for TXT and HINFO records.

HINFO records contain exactly 2 strings, but HINFO does not appear in practice; FWIW, [wikipedia](https://en.wikipedia.org/w/index.php?title=List_of_DNS_record_types&diff=prev&oldid=213247738) claims they are not in current use since 2008.

TXT records may contain more than one string, but almost never do. So I'd say that `make([][]byte, 0, 1)` would be more practical; but this, I believe, performs the same number of allocations as `make([][]byte)` followed by one `append(strings, slice)`, so maybe this line of code should remain as is?

layers/dns.go
orivej: How about this?
```go
type DNSResourceRecord struct {
	...
	
	// RDATA Decoded Values
	IP             net.IP
	NS, CNAME, PTR []byte
	TXTs           [][]byte
	SOA            DNSSOA
	SRV            DNSSRV
	MX             DNSMX

	// Undecoded TXT for backward compatibility
	TXT []byte
}
```

layers/dns.go
orivej: Is this OK?
```go
index2 = index + 1 + int(data[index]) // index2 increases by 1..256 and does not overflow
```

gconnell: looks great :)

layers/dns.go
orivej: Technically they may be 0 bytes long.

layers/dns.go
gconnell: If there's generally only one, that should be fine.

layers/dns.go
gconnell: LGTM

layers/dns.go
orivej: RFC says that TXT has at least one string, but currently gopacket would not raise an error for a malformed packet with an empty TXT: in this case rr.TXT and rr.TXTs would be empty rather than nil slices, but most applications will not see a difference and I do not think that gopacket should behave differently.

orivej:
Please see the updated patch.

PR #119, opened 2015-09-04
Add LayerGoString() to packet.go

randstr:
And minor fixups.

### DIFF
--- a/layers/ip6.go
+++ b/layers/ip6.go
@@ -14,6 +14,14 @@ import (
 	"net"
 )
 
+const (
+	IPv6HopByHopOptionJumbogram = 0xC2 // RFC 2675
+)
+
+const (
+	ipv6MaxPayloadLength = 65535
+)
+
 // IPv6 is the layer for the IPv6 header.
 type IPv6 struct {
 	// http://www.networksorcery.com/enp/protocol/ipv6.htm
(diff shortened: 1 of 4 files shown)


layers/ip6.go
```diff
@@ -399,8 +334,14 @@ func serializeTLVOptionPadding(data []byte, padLength int) {
 	return
 }
 
-func serializeIPv6HeaderTLVOptions(buf []byte, options []*ipv6HeaderTLVOption, fixLengths bool, dryrun bool) int {
+// If buf is 'nil' do a dryrun 
+func serializeIPv6HeaderTLVOptions(buf []byte, options []*ipv6HeaderTLVOption, fixLengths bool) int {
 	var l int
+	var dryrun bool
```
gconnell: dryrun := buf == nil

gconnell:
Thanks for the code!  Just a few minor issues, and we should be good to go.

randstr:
Nice catch. Please confirm fix. And thanks for the review.

packet.go
```diff
@@ -300,6 +301,91 @@ func layerString(i interface{}, anonymous bool, writeSpace bool) string {
...
+
+func layerGoString(i interface{}, b *bytes.Buffer) {
+	if s, ok := i.(fmt.GoStringer); ok {
+		b.WriteString(s.GoString())
+		return
+	}
+
+	var v reflect.Value
+	var ok bool
+	if v, ok = i.(reflect.Value); !ok {
```
gconnell: general style nit, optional:

    v, ok := i.(reflect.Value)
    if !ok {
      ...
    }

PR #120, opened 2015-09-04
Improve IPv6 Routing header support

randstr:


layers/ip6.go
```diff
@@ -34,6 +75,7 @@ type IPv6 struct {
 	HopLimit     uint8
 	SrcIP        net.IP
 	DstIP        net.IP
+	RoutingDstIP net.IP
```
gconnell: Instead of a new field here, I think it would be best to make this a function:

func (i *IPv6) RoutingDstIP() net.IP {
  ... walk down IPv6 options, look for routing options, etc ...
}

Since this is a logical part of the packet and not something actually encoded therein.

Also, I think that by not having this be a field, we can remove the requirement for the new LastNetworkLayer() function in the PacketBuilder.

randstr: If I understand correctly RoutingDst() will look ahead interpreting the IPv6 layer payload bytes until and if we find an IPv6 routing extension header?

gconnell: Yeah, it'd return the "next hop".

gconnell: Which could be DstIP if there is no Routing header.

randstr: It'd have to return the final destination of the routing header, which is not a concern at all once the header is found in the payload.

Having one layer parse the others doesn't feel quite right yet but maybe it'll work. I'll give it some more thought.

gconnell: Cool, thanks!  Happy to accept any suggestions you might have.

I think it should be fine for IPv6 to look at its own options... and it
seems easier than having the options have to go back and find the IPv6
layer, which we currently can't do without changing the packet builder.

randstr: It is possible to simply not use the RoutingDst field in IPv6 struct for decoding, therefore avoiding the whole issue of changing PacketBuilder with LastNetworkLayer(). That field is only essential when serializing to be able to compute the checksum correctly. When decoding it is redundant.

randstr: There's also NetworkFlow() to consider, it should use RoutingDst as destination address.

randstr:
One reason why I think the look-ahead decode strategy won't work is that the IPv6 layer would have to have an understanding of the internal format for every conceivable type of routing header out there. Source Routing, MobileIPv6, RPL, and maybe others. It'll would also not allow any of the extensibility afforded by the enum routing type map (in the same way as all the awesome extensibility afforded by gopacket for IP and others).

randstr:
I have given it more thought and I wouldn't want to change the current implementation too much.

Wouldn't making RoutingDst a private field and just implementing a getter method achieve exactly the same?

```Golang
func (i *IPv6) RoutingDst() net.IP {
    return i.routingDst
}
```

Issue #84, opened 2015-06-03
compare layers.ICMPv6TypeCode


porjo:
I'd like to be able to do this:
```Go
	for _, layerType := range decoded {
		switch layerType {
		case layers.LayerTypeIPv6:
			switch icmp.TypeCode {
			case layers.ICMPv6TypeNeighborSolicitation:
				// do something
```
However comparing `icmp.TypeCode` with `layers.ICMPv6TypeNeighborSolicitation` doesn't work. In this case,   `icmp.TypeCode` == 34560 and `layers.ICMPv6TypeNeighborSolicitation` == 135 (bigendian vs littlendian)

randstr:
Close?

porjo:
@randstr works for me, thanks!

PR #121, opened 2015-09-04
Use struct to represent ICMPv6TypeCode

randstr:
Fixes issue #84


gconnell:
This is a great idea, but unfortunately it's not backwards-compatible, so it'll break compilation for anyone currently using the library.  We strive very hard to be backwards-compatible, to avoid causing issues to our users, so I'm afraid I'm unable to let this one through.

randstr:
I understand and agree with the reasoning but do you think it is worth bumping the major version number (non-backward compatible change) to fix this small wart in the API?

Alternatively I'll submit a small helper function to help with the endianness and bit shifting for the ICMPv6 Type.

gconnell:
Feel free to add any functions to the type to help with endian-ness and bit
shifting.  I'm happy to accept those :)

PR #124, opened 2015-09-09
Add GoString() method to ICMPv6TypeCode

randstr:
Also minor improvement to String() method output format.

### DIFF
--- a/layers/icmp6.go
+++ b/layers/icmp6.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"github.com/google/gopacket"
 	"strconv"
+	"reflect"
 )
 
 type ICMPv6TypeCode uint16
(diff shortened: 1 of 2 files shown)


layers/icmp6.go
```diff
@@ -31,6 +29,47 @@ const (
...
+
+	icmpv6CodeStringMap = map[[2]uint8]string{
+		[2]uint8{1, 0}:   "NoRouteToDst",
+		[2]uint8{1, 1}:   "AdminProhibited",
+		[2]uint8{1, 2}:   "BeyondScopeOfSrc",
+		[2]uint8{1, 3}:   "AddressUnreachable",
+		[2]uint8{1, 4}:   "PortUnreachable",
+		[2]uint8{1, 5}:   "SrcAddressFailedPolicy",
+		[2]uint8{1, 6}:   "RejectRouteToDst",
+		[2]uint8{2, 0}:   "",
```
gconnell: you shouldn't need to add empty strings anywhere... a map will return an empty string if the key doesn't exist.

PR #127, opened 2015-09-09
icmp6: More convenience methods, refactoring

randstr:


layers/icmp6.go
```diff
@@ -152,6 +147,26 @@ type ICMPv6 struct {
...
+}
+
+// Code returns the ICMPv6 code field.
+func (i *ICMPv6) Code() uint8 {
+	return i.TypeCode.Code()
+}
+
+// SetType sets the ICMPv6 type field.
+func (i *ICMPv6) SetType(typ uint8) {
+	i.TypeCode = ICMPv6TypeCode(binary.BigEndian.Uint16([]byte{typ, i.Code()}))
```
gconnell: Set type:  i.TypeCode = (i.TypeCode & 0x00FF) | (uint16(typ) << 8)
Set code:  i.TypeCode =  (i.TypeCode & 0xFF00) | uint16(typ)

randstr: Will do, that definitely seems better.

randstr:
Remove SerializeTo/CreateICMPv6TypeCode or leave it in? Thoughts?

gconnell:
Let's leave it in... I like it :)

randstr:
Ah, you'd rather keep the previous code. OK...

I think this would be a nice to have though.

```Go
// Type returns the ICMPv6 type field.
func (i *ICMPv6) Type() uint8 {
	return i.TypeCode.Type()
}

// Code returns the ICMPv6 code field.
func (i *ICMPv6) Code() uint8 {
	return i.TypeCode.Code()
}
```

gconnell:
Hmm, personally that seems a little redundant, when TypeCode is exported
and you can just say i.TypeCode.Type().  It's not too much extra typing :)
 Thoughts?

randstr:
No problem. :)

randstr:
I do have to add that the reason I like those methods is exactly to avoid the visual redundancy in i.TypeCode.Type(), not really saving keystrokes, but it's not a big deal at all.

PR #128, opened 2015-09-11
fix typo in CONTRIBUTING.md

darxriggs:
Btw. the section "How To Submit Code" should be rewritten. It still refers to the workflow used at code.google.com.


darxriggs:
I signed it!

PR #134, opened 2015-09-13
Make implementation for ICMPv4 as similar as possible to ICMPv6

darxriggs:


gconnell:
Thanks for great work on this!

PR #136, opened 2015-09-21
fixup to comply to contributors guidelines

breml:
* go fmt
* copyright
* .gitignore for binaries built by ./gc


breml:
I signed it!

Issue #137, opened 2015-09-21
Support for handcrafted BPF filters


breml:
I have a use case, where I use gopacket with handcrafted BPF filters because it is not possible to express these BPF filters with the pcap filter expression.

The use case is to do 1 in n sampling of network traffic to collect statistical information. For this use case I use the rand BPF extension and the mod operator as described in https://www.kernel.org/doc/Documentation/networking/filter.txt

For this use case I have implemented the following three methods:
* SetBPFInstructionFilter
* NewBPFInstructionFilter
* CompileBPFFilter

See my pull request for implementation details.

breml:
Pull request merged. Close.

PR #138, opened 2015-09-21
Add methods to apply handcrafted BPF filters

breml:
Added methods
* SetBPFInstructionFilter
* NewBPFInstructionFilter
* CompileBPFFilter
to work with handcrafted BPF filters for cases, which could
not be expressed with the pcap filter expression.

Fixes: #137

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -109,6 +109,15 @@ import (
 
 const errorBufferSize = 256
 
+// Maximum number of BPF instructions supported (BPF_MAXINSNS),
+// taken from Linux kernel: include/uapi/linux/bpf_common.h
+//
+// https://github.com/torvalds/linux/blob/master/include/uapi/linux/bpf_common.h
+const MaxBpfInstructions = 4096
+
+// 8 bytes per instruction, max 4096 instructions
+const bpfInstructionBufferSize = 8 * MaxBpfInstructions
+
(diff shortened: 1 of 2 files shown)


breml:
I signed it!

pcap/pcap.go
```diff
@@ -430,9 +446,86 @@ func (p *Handle) SetBPFFilter(expr string) (err error) {
 	defer C.free(unsafe.Pointer(cexpr))
 
 	if -1 == C.pcap_compile(p.cptr, &bpf, cexpr, 1, C.bpf_u_int32(maskp)) {
+		return bpf, p.Error()
```
gconnell: free the bpf here and return nil, err?

pcap/pcap.go
```diff
@@ -430,9 +446,86 @@ func (p *Handle) SetBPFFilter(expr string) (err error) {
...
+	return bpf, nil
+}
+
+// CompileBPFFilter compiles and returns a BPF filter for the pcap handle.
+func (p *Handle) CompileBPFFilter(expr string) ([]BPFInstruction, error) {
+	bpf, err := p.compileBPFFilter(expr)
+	if err != nil {
+		return nil, err
+	}
+
```
gconnell: move the pcap_freecode call up here as a defer?

I'm worried that a future coder might throw a return somewhere else in this function.

defer C.pcap_freecode(&bpf)

PR #139, opened 2015-09-23
CompileBPFFilter C.pcap_freecode

breml:
@gconnell thanks for the hints.

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -446,6 +446,7 @@ func (p *Handle) compileBPFFilter(expr string) (_Ctype_struct_bpf_program, error
 	defer C.free(unsafe.Pointer(cexpr))
 
 	if -1 == C.pcap_compile(p.cptr, &bpf, cexpr, 1, C.bpf_u_int32(maskp)) {
+		C.pcap_freecode(&bpf)
 		return bpf, p.Error()
 	}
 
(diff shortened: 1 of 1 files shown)


pcap/pcap.go
```diff
@@ -455,6 +456,7 @@ func (p *Handle) compileBPFFilter(expr string) (_Ctype_struct_bpf_program, error
 // CompileBPFFilter compiles and returns a BPF filter for the pcap handle.
 func (p *Handle) CompileBPFFilter(expr string) ([]BPFInstruction, error) {
 	bpf, err := p.compileBPFFilter(expr)
+	defer C.pcap_freecode(&bpf)
```
gconnell: I think this one will fail, since compileBPFFilter may already call freecode (new line 449), and this would make it freecode twice on the same object.

breml: You are right. What is the better way to resolve this issue:

1. Remove the `C.pcap_freecode` in function compileBPFFilter on line 449 and delegate the freeing always to the calling function (remove line 449)
2. Move the `defer`statements in CompileBPFFilter and SetBPFFilter below the error checking code

gconnell: Also, the compileBPFFilter call is not exported, so library users don't have to care too much :)

gconnell: I'd lean towards option 2, but happy to accept either option.  Let's add a comment to compileBPFFilter detailing what callers are expected to do.

breml:
@gconnell Updated the code based on option 1 and added a comment to the function compileBPFFilter.

PR #140, opened 2015-09-25
Add export of stats from a socket (afpacket)

LrsK:
I have added SocketStats and SocketStatsV3 fields to the TPacket struct, to expose the statistics gathered by the socket. I decided to use the structs that already exists in if_packet.h: tpacket_stats and tpacket_stat_v3. There is an InitSocketStats() method that clears the internal socket counters. To be honest I was unsure on how to best expose the stats based on which version (1/2/3) is being used, but decided to expose both V1/V2 stats, and V3 stats in the SocketStats() method, so that the user doesn't have to know the version being used when asking for stats. This could be solved another way by making a SocketStatsV3() method.


LrsK:
I signed it!

PR #151, opened 2015-11-18
Support for OSX/BSD based kernels that need the length and frag offset part of the IPv4 header in host byte order.

### LINKED ISSUE #150: on "darwin", "dragonfly", "freebsd", "netbsd" the ip total length and frag off needs to be host byte order
https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man4/ip.4.html

go down to raw ip sockets and the line "Note that the ip_off and ip_len fields are in host byte order."

This doesn't happen in gopacket and I was very confused for a while.

In the ipv4 package it is done there

https://github.com/golang/net/blob/master/ipv4/header.go (line 75)

Its trivial to fix in your own code but it would be nice to have it in gopacket.

nhooyr:
fixes #150 

I also ran gofmt to simplify.


layers/ip4.go
```diff
@@ -110,9 +112,15 @@ func (ip *IPv4) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeO
 	}
 	bytes[0] = (ip.Version << 4) | ip.IHL
 	bytes[1] = ip.TOS
-	binary.BigEndian.PutUint16(bytes[2:], ip.Length)
+	switch runtime.GOOS {
```
gconnell: I'm quite certain that the byte order of an IP packet does not change depending on the host OS.  I believe this code is in error.  Packet format doesn't change per-OS or per-architecture... if it did, different OS's wouldn't be able to communicate with each other over the wire.

Note that SerializeTo is writing the WIRE format for packets, which for IPv4 is always big-endian.

Can you provide a concrete example where this isn't the case?  A captured packet from one of these OS's that shows the little-endian order?

nhooyr: @gconnell You're right. I overlooked it sorry. Over the wire it's in big endian. But when you're working with INET raw sockets in those OSes, the len and frag offset are in little endian. Pretty sure this is only with raw INET sockets. On the wire the byte order is big-endian. I'm not sure why, but it is.

See https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man4/ip.4.html "Note that the ip_off and ip_len fields are in host byte order."

There is also https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man4/icmp.4.html
"Incoming packets are received with the IP header and options intact.  Note that the ip_off and ip_len fields are in host byte order"

I've gotta go for now, but I'll be back in 30 minutes to discuss a solution.

nhooyr:
@gconnell alright so on those OSes when sending/receiving IP packets with raw IP sockets, the length and fragment offset are both in host byte order. Since its only for IP packets and with raw IP sockets I think we should just add a extra method to the IPv4 struct to serialize itself and other layers given as arguments. I'll add tests to see if this is working and then I think it should be fine. Should I get started or is there another way to solve this.

gconnell:
How about something like:

func (i *IPv4) RawSocketByteOrder() {
  i.Length = switchbyteorder(i.Length)
  i.FragOffset = switchbyteorder(i.FragOffset)
}

That way, it'll work if we both serialize TO:

  ip4 := ...
  ip4.RawSocketByteOrder()
  // serialize

AND if we read FROM:

  ip4 := IP4(readRawSocketPacketData())
  ip4.RawSocketByteOrder()
  ip4.Length // is now correct

nhooyr:
@gconnell Yea your right thats much better. I'm on it!

nhooyr:
@gconnell alright done, what do you think now.

nhooyr:
@gconnell there is just one problem with this approach. I like my length to be automatically be determined by goPacket.SerializeLayers but this approach doesn't really allow that to happen seamlessly. I have to find the length manually (adding the tcp length and other stuff) and then convert to host byte order and then finally serialize it. What do you think? 

I think we should add a field to the IPv4 struct to specify this instead of goPacket.SerializeOptions since this is IPv4 specific.

nhooyr:
@gconnell is this every going to be merged? its been several months.

ajvb:
:+1:

PR #152, opened 2015-11-22
Added raw IPv4 and IPv6 packet wrapping linktype

fclairamb:
These link types are perfectly supported by tcpdump & wireshark. Instead of specifying a LinkTypeEthernet and wrapping ethernet frames, you can specify a LinkTypeIPv4 and wrap IPv4 frames (skipping the ethernet wrapping).

### DIFF
--- a/layers/enums.go
+++ b/layers/enums.go
@@ -92,7 +92,7 @@ const (
 type LinkType uint8
 
 const (
-	// According to pcap-linktype(7).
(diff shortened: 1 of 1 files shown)


fclairamb:
I signed it!

fclairamb:
So ? Any issue with this ? Do you want some sample code to show you how to produce an IPv4 stream that is directly parsable by tcpdump ?

lhausermann:
That's look OK for me. 
As Graeme give me the "merge power", PR accepted ;)

PR #158, opened 2015-12-29
New example program : pcaplay

lhausermann:
Pcaplay is a simple utility program to replay a recorded pcap file, trying to get the same timing than the orign packets.

Its dumps some stats during operation, and offer a fast mode where it sends packet as fast as he can.

### DIFF
--- a/examples/pcaplay/main.go
+++ b/examples/pcaplay/main.go
@@ -24,28 +24,74 @@ import (
 
 var iface = flag.String("i", "eth0", "Interface to write packets to")
 var fname = flag.String("r", "", "Filename to read from")
+var fast = flag.Bool("f", false, "Send each packets as fast as possible")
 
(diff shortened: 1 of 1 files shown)


examples/pcaplay/main.go
```diff
@@ -0,0 +1,159 @@
...
+	"os"
+	"strings"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/examples/util"
+	"github.com/google/gopacket/pcap"
+)
+
+var iface = flag.String("i", "eth0", "Interface to write packets to")
```
gconnell: wrap all of these into a single var (?

var (
  fname = flag.String(...)
  fast = flag.Bool(...)
  etc...
)

gconnell:
Looks great, thanks!

PR #160, opened 2016-01-11
Adding additional parsing capabilities to layers/igmp.go

jesseward:
Added support for IGMPv1 and IGMPv2 types. Included here is
MembershipQuery (v[123]), MembershipReportV1, MembershipReportV2,
MembershipReportV3 and LeaveGroup (v2)

Including tests supplied in layers/igmp_test.go

### DIFF
--- a/layers/igmp.go
+++ b/layers/igmp.go
@@ -9,6 +9,7 @@ package layers
 
 import (
 	"encoding/binary"
+	"fmt"
 	"github.com/google/gopacket"
 	"net"
 	"time"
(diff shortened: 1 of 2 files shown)


jesseward:
I signed it!

layers/igmp.go
```diff
@@ -16,18 +16,210 @@ import (
...
-	SourceAddresses         []net.IP
+	Type    IGMPType  // IGMP message type
+	Message IGMPQuery // Container for message payload
+	Version uint8     // IGMP protocol version
+}
+
+type GroupRecordType uint8
+
+const (
+	IsIn  GroupRecordType = 0x01 // Type MODE_IS_INCLUDE, source addresses x
```
gconnell: The 'layers' package contains a namespace used by a whole ton of different protocols.  As such, please prefix any protocol-related objects with the name of the protocol:

IGMPGroupRecordType
IGMPIsIn

etc?

layers/igmp.go
```diff
@@ -16,18 +16,210 @@ import (
...
+}
+
+type IGMPQuery interface {
+	decodeResponse(data []byte) error
+}
+
 // IGMP is the packet structure for IGMP messages.
 type IGMP struct {
 	BaseLayer
-	Type            IGMPType
```
gconnell: This change to the IGMP type breaks backwards compatibility:  If someone's already using type IGMP and references, for example, the RobustnessValue field, this change will break their code.  We try very hard to not break existing code with new commits.

Is the current IGMP struct version-specific?  If so, maybe just create a new struct IGMPv2 or whatever and leave IGMP around so old code continues to run.  Thoughts?

layers/igmp.go
jesseward: noted @gconnell , will correct naming as to avoid name scape collisions.

layers/igmp.go
jesseward: > This change to the IGMP type breaks backwards compatibility.....  We try very hard to not break existing code with new commits.

got it and makes sense. Great stance to take.

> Is the current IGMP struct version-specific

It is not. But the current IGMP implementation only calls out a single IGMP query type. That being When [byte 0 == 0x11](https://github.com/google/gopacket/blob/master/layers/igmp.go#L53 ) ( a Membership query) . This type happens to be shared between all 3 versions of IGMP. Version inference of the membership query type is then made based on content length of the packet and contents of byte 1.

I am guessing that IGMP v1 is rarely used so it probably safe to ignore, but V2 still has some use. I will revisit this pull request and see if it can be squeezed into the current implementation.

jesseward:
@gconnell I have taken another stab at this pull request and incorporated your suggestions. Let me know how the [latest revision](https://github.com/jesseward/gopacket/commit/ae935a92bcc0626e7d99afe537023227336cc0f0) looks.

layers/igmp.go
```diff
@@ -16,22 +17,204 @@ import (
...
 	IntervalTime            time.Duration
 	SourceAddresses         []net.IP
+	NumberOfGroupRecords    uint16
+	NumberOfSources         uint16
+	GroupRecords            []IGMPv3GroupRecord
+	Version                 uint8 // IGMP protocol version
+
+}
+
+//  0                   1                   2                   3
```
gconnell: type and function comments should start with type/function name.  IE: move line 101 up here, and do the same for other places.

layers/igmp.go
```diff
@@ -44,26 +227,42 @@ func igmpTimeDecode(t uint8) time.Duration {
...
+	i.Type = IGMPType(data[0])
+
+	switch i.Type {
+
+	case IGMPMembershipQuery:
+		i.decodeIGMPv3MembershipQuery(data)
+
+	case IGMPMembershipReportV3:
+		i.decodeIGMPv3MembershipReport(data)
 	}
```
gconnell: default to returning an error ("unsupported IGMP type", or something like that).

layers/igmp.go
```diff
@@ -16,22 +17,204 @@ import (
...
+
+//  0                   1                   2                   3
+//  0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |      Type     | Max Resp Time |           Checksum            |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// |                         Group Address                         |
+// +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+// IGMPv1Orv2 stores header details for an IGMPv1 or IGMPv2 packet.
+type IGMPv1Orv2 struct {
```
gconnell: This name causes me trouble, v1Orv2 looks like v10rv2, which is nonsensical.  Maybe just IGMPv1 or v1or2?  I'm okay with breaking the camelcase rules for readability.

layers/igmp.go
```diff
@@ -44,26 +227,42 @@ func igmpTimeDecode(t uint8) time.Duration {
...
+}
+
+// DecodeFromBytes decodes the given bytes into this layer.
+func (i *IGMP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+
+	// common IGMP header values between versions 1..3 of IGMP specification..
+	i.Type = IGMPType(data[0])
+
+	switch i.Type {
+
```
gconnell: I'd reduce the switch statement whitespace here and elsewhere.  Just this should be fine:

    switch x {
    case A:
      blah
    case B:
      blah
    }

layers/igmp.go
```diff
@@ -77,7 +276,44 @@ func (i *IGMP) NextLayerType() gopacket.LayerType {
...
+	// byte 0 contains IGMP message type.
+	switch IGMPType(data[0]) {
+
+	case IGMPMembershipQuery:
+
+		// IGMPv3 Membership Query payload is >= 12
+		if len(data) >= 12 {
+			i := &IGMP{Version: 3}
+			return decodingLayerDecoder(i, data, p)
+
```
gconnell: remove extra whitespace here, too, and line 299

layers/igmp_test.go
```diff
@@ -0,0 +1,144 @@
...
+	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x04, 0x52, 0x00, 0x00, 0x00, 0x00,
+}
+
+func TestIGMPv1MembershipReportPacket(t *testing.T) {
+	p := gopacket.NewPacket(igmpv1MembershipReportPacket, LinkTypeEthernet, gopacket.Default)
+	if p.ErrorLayer() != nil {
+		t.Error("Failed to decode packet:", p.ErrorLayer().Error())
+	}
+	checkLayers(p, []gopacket.LayerType{LayerTypeEthernet, LayerTypeIPv4, LayerTypeIGMP}, t)
+}
```
gconnell: grab out the IGMP layers and make sure they look like you expect:

    igmp := p.Layer(LayerTypeIGMP).(*IGMP)
    // test fields

layers/igmp_test.go
```diff
@@ -0,0 +1,144 @@
...
+	0x01, 0x3c, 0x12, 0x00, 0x0c, 0xc3, 0xe0, 0x00, 0x01, 0x3c, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+	0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0x04, 0x52, 0x00, 0x00, 0x00, 0x00,
+}
+
+func TestIGMPv1MembershipReportPacket(t *testing.T) {
+	p := gopacket.NewPacket(igmpv1MembershipReportPacket, LinkTypeEthernet, gopacket.Default)
+	if p.ErrorLayer() != nil {
+		t.Error("Failed to decode packet:", p.ErrorLayer().Error())
+	}
+	checkLayers(p, []gopacket.LayerType{LayerTypeEthernet, LayerTypeIPv4, LayerTypeIGMP}, t)
```
gconnell: should this be LayerTypeIGMPv1orv2, instead of LayerTypeIGMP?

gconnell:
Should this still be open?

jesseward:
@gconnell , no i will close this out.

PR #168, opened 2016-02-18
167.fix bsd bpf.0

david415:
Hi. What do you think of this solution to the bug in the BPF sniffer?

This also includes a change to make ``pickBpfDevice`` panic if no BPF device was acquired.

### DIFF
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -196,7 +196,7 @@ func (b *BPFSniffer) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
 			CaptureLength: 0,
 			Length:        0,
 		}
-		return nil, captureInfo, nil
+		return nil, captureInfo, fmt.Errorf("BPF captured frame received with corrupted BpfHdr struct.")
 	}
 
 	rawFrame := b.readBuffer[frameStart : frameStart+int(hdr.Caplen)]
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -187,6 +187,16 @@ func (b *BPFSniffer) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
 	hdr := (*unix.BpfHdr)(unsafe.Pointer(&b.readBuffer[b.readBytesConsumed]))
 	frameStart := b.readBytesConsumed + int(hdr.Hdrlen)
 	b.readBytesConsumed += bpfWordAlign(int(hdr.Hdrlen) + int(hdr.Caplen))
+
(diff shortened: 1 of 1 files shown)


bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -171,6 +172,9 @@ func (b *BPFSniffer) pickBpfDevice() {
 			break
 		}
 	}
+	if b.options.BPFDeviceName == "" {
```
gconnell: I don't believe this check does what you want it to.  Note that after the first iteration of the loop, BPFDeviceName is no longer empty, due to line 169.  So this if statement will never run.

Instead, how about changing the 'break' at line 172 to a 'return', then just panicing if you get past the for loop?

Or, better yet, make this function return an error, and check it at line 100, returning the error if it's non-nil.

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -187,6 +191,16 @@ func (b *BPFSniffer) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
...
 	frameStart := b.readBytesConsumed + int(hdr.Hdrlen)
 	b.readBytesConsumed += bpfWordAlign(int(hdr.Hdrlen) + int(hdr.Caplen))
+
+	if frameStart+int(hdr.Caplen) > len(b.readBuffer) {
+		captureInfo := gopacket.CaptureInfo{
+			Timestamp:     time.Unix(int64(hdr.Tstamp.Sec), int64(hdr.Tstamp.Usec)*1000),
+			CaptureLength: 0,
+			Length:        0,
+		}
+		return nil, captureInfo, nil
```
gconnell: should this error be nil?  Or should another non-nil error detailing what's going on here be returned?

david415: Unclear. Why is this error happening? Perhaps a bug in the OpenBSD kernel. If we were to return an error then i'm not sure exactly what we'd say; perhaps that a BPF frame's header claims the frame is bigger than it is.

bsdbpf/bsd_bpf_sniffer.go
david415: OK... I've corrected this mistake.

bsdbpf/bsd_bpf_sniffer.go
david415: I decided to return an error. Oh but we should specify our own error type so that the user of this API can choose to ignore this class of error.

david415:
Unless you'd like me to make a change I'm happy with the current state of this pull request.

Right now it returns an error via fmt.Errorf :
```
return nil, captureInfo, fmt.Errorf("BPF captured frame received with corrupted BpfHdr struct.")
```

Please merge this code. =-)

PR #176, opened 2016-02-28
Adding a VXLAN layer that parses RFC 7348 and vxlan group policy

hyakuhei:
https://tools.ietf.org/html/rfc7348
https://tools.ietf.org/html/draft-smith-vxlan-group-policy-00

This is my first PR against gopacket, I hope you don't mind it being made unsolicited - I needed vxlan parsing for a project I'm working on at the moment.

I'm extremely new to Go, happy to make changes.

-Rob


hyakuhei:
Signed the Google CLA as suggested by the friendly bot.

layers/vxlan.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2016 Robert Clark. All rights reserved.
```
gconnell: Please use the standard Copyright:


    // Copyright 2016 Google, Inc. All rights reserved.
    //
    // tree.

layers/vxlan.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Policy ID as per the group policy draft
+	vx.GBPGroupPolicyID = binary.BigEndian.Uint16(data[2:4])
+
+	vxlanLength := 8
+	vx.Contents = data[:vxlanLength]
+	vx.Payload = data[vxlanLength:]
+
+	p.AddLayer(vx)
+	return p.NextDecoder(LinkType(1))
```
gconnell: Pass in `LinkTypeEthernet` instead of `LinkType(1)`

layers/vxlan_test.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2016 Robert Clark. All rights reserved.
```
gconnell: Fix copyright here too

layers/vxlan.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// 'A' bit - indicates that the group policy has already been applied to this packet.
+	if (data[1] & 0x80) > 0 {
+		vx.GBPApplied = true
+	}
+
+	// Policy ID as per the group policy draft
+	vx.GBPGroupPolicyID = binary.BigEndian.Uint16(data[2:4])
+
+	vxlanLength := 8
```
gconnell: const vxlanLength = 8

layers/vxlan.go
```diff
@@ -0,0 +1,74 @@
...
+	// 'I' bit per RFC7348
+	if (data[0] & 0x08) > 0 {
+		vx.ValidIDFlag = true
+	}
+
+	// VNI - VXLAN Network Identifier per RFC7348
+	vx.VNI = binary.BigEndian.Uint32(append([]byte{0x0}, data[4:7]...)) //Uint32 wants 4 bytes in a slice
+
+	// 'G' bit per the group policy draft
+	if (data[0] & 0x80) > 0 {
```
gconnell: To save on vertical space, maybe write:

    vx.ValidIDFlag = data[0] & 0x08 > 0  // 'I' bit per RFC7348
    vx.GBPExtension = data[0] & 0x80 > 0  // 'G' bit per the group policy draft
    vx.GBPDontLearn = data[1] & 0x40 > 0  // 'D' bit - the egress VTEP MUST NOT learn the source address of the encapsulated frame.
    vx.GBPApplied = data[1] & 0x80 > 0  // 'A' bit - indicates that the group policy has already been applied to this packet.

layers/vxlan.go
```diff
@@ -0,0 +1,74 @@
...
+func decodeVXLAN(data []byte, p gopacket.PacketBuilder) error {
+	vx := &VXLAN{}
+
+	// 'I' bit per RFC7348
+	if (data[0] & 0x08) > 0 {
+		vx.ValidIDFlag = true
+	}
+
+	// VNI - VXLAN Network Identifier per RFC7348
+	vx.VNI = binary.BigEndian.Uint32(append([]byte{0x0}, data[4:7]...)) //Uint32 wants 4 bytes in a slice
```
gconnell: This append actually forces a malloc(), which is slow and, here, easily avoided.  Try this:

    var buf [4]byte  // Stack-allocated, so we avoid malloc.  Also, pre-zero'd.
    copy(buf[1:], data[4:7])  // copy data bytes we care about into the rest of the buffer
    vx.VNI = binary.BigEndian.Uint32(buf[:])  // now, we can pass the stack buffer into this function

gconnell: Alternatively:

    vx.VNI = binary.BigEndian.Uint32(data[3:7]) & 0xFFFFFF

gconnell:
Super nice work, thanks!

I've added a few comments, mostly small things.  Once they're done, I think we should be ready to merge.

Re: copyright, see item 2 here:  https://cla.developers.google.com/about/google-corporate

hyakuhei:
Thanks for all the great feedback!

Another commit on the way. I've altered the layout vertically as per your advice. I went with allocating a few bytes on the stack for the BigEndian.Uint32 issue as I thought it was a lot clearer to read.

hyakuhei:
Hi Graeme,

Thank you for the comments, they're really appreciated!

I pushed a new version up a few days ago which I believe addressed your
comments, could you do another review when you have time please?

-Rob

layers/vxlan.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// RFC 7348 https://tools.ietf.org/html/rfc7348
+	vx.ValidIDFlag = data[0] & 0x08 > 0		// 'I' bit per RFC7348
+	vx.VNI = binary.BigEndian.Uint32(buf[:]) // VXLAN Network Identifier per RFC7348
+
+	// Group Based Policy https://tools.ietf.org/html/draft-smith-vxlan-group-policy-00
+	vx.GBPExtension = data[0] & 0x80 > 0	// 'G' bit per the group policy draft
+	vx.GBPDontLearn = data[1] & 0x40 > 0  // 'D' bit - the egress VTEP MUST NOT learn the source address of the encapsulated frame.
+	vx.GBPApplied = data[1] & 0x80 > 0		// 'A' bit - indicates that the group policy has already been applied to this packet.
+	vx.GBPGroupPolicyID = binary.BigEndian.Uint16(data[2:4])	// Policy ID as per the group policy draft
```
gconnell: I think you might need to run gofmt... your comment spacing is a little weird.

Try running "./gc -a" from the top-level, that'll run all tests, formatting, etc, to get your contribution looking squeaky clean.

gconnell:
Looks good except for gofmt issues.  Fix those up and ping again, and I think we should be good to go.

hyakuhei:
Thanks once again for your comments - new version should be ready now

-Rob

PR #177, opened 2016-03-03
Add bounds checking for pcapgo read

david415:
make sure slice end is within buffer size

### DIFF
--- a/pcapgo/read.go
+++ b/pcapgo/read.go
@@ -102,6 +102,7 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
 
 	var n int
 	if 16+ci.CaptureLength > len(r.buf) {
+		err = gopacket.ErrMalformedPacket
 		return
 	}
 	data = r.buf[16 : 16+ci.CaptureLength]
--- a/pcapgo/read.go
+++ b/pcapgo/read.go
@@ -102,7 +102,6 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
 
 	var n int
 	if 16+ci.CaptureLength > len(r.buf) {
-		err = gopacket.ErrMalformedPacket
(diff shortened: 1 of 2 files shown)


pcapgo/read.go
```diff
@@ -101,6 +101,9 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
 	}
 
 	var n int
+	if 16+ci.CaptureLength > len(r.buf) {
+		return
```
gconnell: Should return an error.  Note that if this happens, it's a malformed packet file, since the file said the snaplen was X and is now offering up a packet that's X+N.  Maybe have the error say that.

hyakuhei: +1 to providing that information in the error.

There are times that gopacket telling me that a packet is invalid is perfectly normal.

david415: OK I addressed this issue in the next commit. Please review my solution.

pcapgo/read.go
```diff
@@ -101,6 +101,10 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
 	}
 
 	var n int
+	if 16+ci.CaptureLength > len(r.buf) {
+		err = gopacket.ErrMalformedPacket
```
gconnell: For now, maybe just a regular error with more context:

err = fmt.Errorf("capture length with header exceeds buffer size: %d > %d", 16+ci.CaptureLength, len(r.buf))

If you're dead set on having an error up in packet.go, please add a comment to it:

// ErrMalformedPacket is returned when ...
var ErrMalformedPacket ...

david415:
this latest commit should satisfy your last comment. please review

david415:
Can we get this merged now that I fixed the code change?

gconnell:
Hey, sorry about the wait, I missed the email for your last commit.

PR #179, opened 2016-03-03
Add bounds checking for reading pcap packets

david415:
return zero capture info struct and zero packet buffer if
size mismatch detected

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -362,6 +362,11 @@ func (p *Handle) ZeroCopyReadPacketData() (data []byte, ci gopacket.CaptureInfo,
 		slice.Cap = ci.CaptureLength
 	}
 	p.mu.Unlock()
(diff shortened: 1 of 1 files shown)


pcap/pcap.go
```diff
@@ -362,6 +362,11 @@ func (p *Handle) ZeroCopyReadPacketData() (data []byte, ci gopacket.CaptureInfo,
 		slice.Cap = ci.CaptureLength
 	}
 	p.mu.Unlock()
+	if slice.Len != len(slice.Data) {
```
gconnell: I'm guessing this doesn't compile?  The 'slice' variable shouldn't be referencable here.  Also, note that slice.Data is NOT a slice.  It's a uintptr, so len(slice.Data) makes no sense.

david415:
yes this makes no sense... sorry for the confusion

Issue #104, opened 2015-08-10
AF_PACKET: need to expose network interface from which packet was captured


rshriram:
Context: Host with several docker containers
Capture mode: AF_PACKET, all interfaces [i.e., bound to pseudo interface, capturing packets from all interfaces]

When running gopacket based app on a machine with multiple docker containers, its often essential to identify the interface from which the packet was captured, so that it can be tied back to the target container. Looking at the metadata information (or even the ethernet layer code), the network interface information is not exposed.

In the absence of this information, it is necessary to run the packet capture process on each virtual interface, which is highly inefficient. (https://groups.google.com/forum/m/#!topic/linux.kernel/3T8aWc_gLCc)

With AF_PACKET, its possible to monitor packets arriving on all interfaces using ifindex 0, and after the packet arrives at the userpspace, its possible to identify the specific interface using the sockaddr_ll.
Exposing this information in either the metadata or as part of some layer would be very helpful.

LrsK:
I think this is possible by looking at the sockaddr_ll section of the tpacket frame. 
In https://www.kernel.org/doc/Documentation/networking/packet_mmap.txt it says:

```
   Frame structure:
   - Start. Frame must be aligned to TPACKET_ALIGNMENT=16
   - struct tpacket_hdr
   - pad to TPACKET_ALIGNMENT=16
   - struct sockaddr_ll
   - Gap, chosen so that packet data (Start+tp_net) aligns to TPACKET_ALIGNMENT=16
   - Start+tp_mac: [ Optional MAC header ]
   - Start+tp_net: Packet data, aligned to TPACKET_ALIGNMENT=16.
   - Pad to align to TPACKET_ALIGNMENT=16
```

The sockaddr.ll section could perhaps be extracted by performing:

```
// For V3 headers
var ll *C.struct_sockaddr_ll
//     Start at w.packet                           +  aligned size of w.packet
d := uintptr(unsafe.Pointer(w.packet)) + uintptr(tpacketAlign(int(unsafe.Sizeof(w.packet))))
ll = (*C.struct_sockaddr_ll)(unsafe.Pointer(&d))
```

sockaddr_ll.sll_ifindex will give the index of the interface, which can be mapped to a string at init-time by using ioctl with SIOCGIFNAME.

gconnell:
I believe this is now done.

PR #183, opened 2016-03-22
Added ability to listen on all interfaces with afpacket.

LrsK:
As discussed in #104, exposing the port on which the packet came in is useful in certain situations. I have added the ability to extract this information from the packet header, with the index of the port now stored in the CaptureInfo metadata field of the packet. This index can later be translated to a name by invoking e.g. net.InterfaceByIndex().

Note: To perform the proper tpacket alignment calculations, I had to add a different alignment function "tpAlign" which seems to work, where the included "tpacketAlign" did not. I see that "tpacketAlign" is used for processing of the packets in the blocks in v3-headers, but I'm not sure how that is working if the alignment is incorrect. I tested using my "tpAlign" function in place of "tpacketAlign", which seemed to work fine, but I didn't want to make the change in this PR. Since "tpacketAlign" is only used one place, it would be an easy fix in the future.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -91,14 +91,20 @@ type TPacket struct {
 
 // bindToInterface binds the TPacket socket to a particular named interface.
 func (h *TPacket) bindToInterface(ifaceName string) error {
-	iface, err := net.InterfaceByName(ifaceName)
-	if err != nil {
-		return fmt.Errorf("InterfaceByName: %v", err)
+	ifIndex := 0
+	// An empty string here means to listen to all interfaces
+	if ifaceName != "" {
+		iface, err := net.InterfaceByName(ifaceName)
+		if err != nil {
+			return fmt.Errorf("InterfaceByName: %v", err)
+		}
+		ifIndex = iface.Index
 	}
+
 	var ll C.struct_sockaddr_ll
 	ll.sll_family = C.AF_PACKET
 	ll.sll_protocol = C.__be16(C.htons(C.ETH_P_ALL))
-	ll.sll_ifindex = C.int(iface.Index)
+	ll.sll_ifindex = C.int(ifIndex)
 	if _, err := C.bind(h.fd, (*C.struct_sockaddr)(unsafe.Pointer(&ll)), C.socklen_t(unsafe.Sizeof(ll))); err != nil {
 		return fmt.Errorf("bindToInterface: %v", err)
 	}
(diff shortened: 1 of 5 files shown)


gconnell:
I'm somewhat concerned about the addition of InterfaceIndex to CaptureInfo, specifically the fact that it's only currently set in afpacket.  pcap, for example, doesn't set it right now.  Are we sure an interface index is the correct way to export things?  What will be set, for example, when reading from a pcap file?  Is 0 a valid interface index, and if so, how do we differentiate between "not set" and "set, valid, and zero"

LrsK:
This is a good point. To think it though:

- I am pretty sure that 0 (zero) is never a valid index value in Linux, given the behavior of many of the network functions such as if_nametoindex() which returns 0 with an error.
- CaptureInfo.InterfaceIndex will (as of this writing) be 0 in every case where pcap or pfring is used. This could be left as-is for now, or we can try to add similar functionality for pcap and pfring.
- To add this for pfring, we can add a reference to the [pfring_extended_pkthdr](http://www.ntop.org/pfring_api/structpfring__extended__pkthdr.html) which has a similar if_index field. Pfring is set to capture from all interfaces by setting the device string to "any".
- Adding this for pcap may be difficult because there is no interface information in the pcap_pkthdr, in fact when you listen to all devices with pcap, [you don't get real layer 2 headers at all](http://man7.org/linux/man-pages/man3/pcap_next_ex.3pcap.html). However, [you are still able to listen in all devices by passing in "any" or NULL](http://man7.org/linux/man-pages/man3/pcap_open_live.3pcap.html). Here an index of 0 may be appropriate for the "any" case.
- We could try to add the functionality in cases where the device is set, so that it is not 0 when it clearly shouldn't be.
    - For pcap, do a reverse-lookup of the handle device name to the index and add to the handle struct
    - For pfring, do a reverse-lookup from device to the index and add the index to the Ring struct
    - In above cases just plug in this index to CaptureInfo.InterfaceIndex 
- For pcap files, there is by definition no device, so an index of 0 is appropriate. Of course it is possible/probable that the capture is from a real device, but given the information above this will have to be known a priori anyway.

I'm not sure which approach is the best here, but it seems like we at least should add the proper index for pfring captures, since the index is available. Adding the index number to the Ring/Handle is an option too, but may seem unnecessary since in the cases outlined above, the interface name is already known to the user.

LrsK:
I have added more code to enable pcap and pfring captures to add interface index to their packets' metadata.

The way to works now is:
    - for pcap, the interface index is added when the handle is created. If the device does not exist, the index is set to 0, in case the device string is still accepted, which is the case for "any".
    - for pfring, the interface index is added when the ring is created. If the flag pfring.FlagLongHeader is set, the ring will remember this and use whatever is in the extended packet header. If it is not set, we use the interface name passed in by the user. In the case of pfring, you are supposed to be able to pass in "any", and the if_index field of the extended header will show the correct interface where the packet came in, but I was unable to successfully capture packets by passing in the "any" device name. If this works on other machines or is fixed in the future, the added solution will just work as expected.
    - For pcap files, the interfaceIndex will always be 0
    - For afpacket, the interfaceIndex will always be correct.

LrsK:
Any feedback on the latest commit @gconnell ?

gconnell:
Sorry for the slow reply on this, slipped my mind.  Merged.

PR #188, opened 2016-04-25
[pcap][workaround] Avoid deadlocking on waiting on packets

nplanel:
As ReadPacketData() is called usually in loop, goroutine scheduler seems
to be not very fair when running on a 1 core CPU (GOMAXPROCS = 1 as NumCPU
= 1)
As the p.mu lock is re-acquired (on the next loop), a call to Close() have
poor chance to acquire the lock.
So the workaround is to give a chance to other goroutine waiting on a same
Mutex if pcap handler return timeout.


nplanel:
I signed it!

PR #189, opened 2016-05-04
Fix PF_RING WritePacketData rv error

camathieu:
C.pfring_send return the number of bytes sent.
https://github.com/ntop/PF_RING/blob/6.2.0-stable/userland/lib/pfring.h#L630

Without this fix I always get error like this : 
2016/05/04 16:34:47 Unable to send packet data, got error code 60
60 is not an error code but the number of bytes sent.


camathieu:
I signed it!

PR #193, opened 2016-05-10
Add GRE packet decoding and update tests and benchmarks.

larit:


gconnell:
Could you add a test that:

starts with a set of layers
serializes it, with a checksum
deserializes it, checks checksum against known-good value

I believe all the current tests skip checksumming in GRE due to not having ChecksumPresent set.

gconnell:
Also, please sign CLA.

larit:
Added some tests for checksum. Please take another look.

layers/gre_test.go
```diff
@@ -253,3 +254,138 @@ func BenchmarkEncodePacketEthernetOverGRE(b *testing.B) {
...
+	}
+}
+
+func setNetworkLayer(layers []gopacket.SerializableLayer) error {
+	type setNetworkLayerForChecksum interface {
+		SetNetworkLayerForChecksum(gopacket.NetworkLayer) error
+	}
+	var l gopacket.NetworkLayer
+	for _, layer := range layers {
+		n, ok := layer.(gopacket.NetworkLayer)
```
gconnell: for the record, in the future you can say:

```go
if n, ok := ...; ok {
  do something with n
}
```

same with line 380

layers/gre.go
```diff
@@ -154,11 +156,8 @@ func (g *GRE) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOpt
 		// zero out checksum bytes
 		buf[4] = 0
```
gconnell: I expect we want to zero out regardless, then maybe overwrite?

```go
buf[4] = 0
buf[5] = 0
if computeChecksum {
  overwrite buf[4] and buf[5]
}

File: layers/gre_test.go [hunk 3]
```diff
@@ -253,3 +254,138 @@ func BenchmarkEncodePacketEthernetOverGRE(b *testing.B) {
 		buf.Clear()
 	}
 }
+
+var testGREChecksum = map[uint16][]gopacket.SerializableLayer{
+	0x77ff: {
```
gconnell: has at least one of these come from an in-the-wild GRE packet?  If not, maybe we can see if we can rustle one up somewhere.

Just want to make absolutely sure that the checksums we're generating actually match in-the-wild checksums.

layers/gre.go
gconnell: Actually, I think the logic we want here should be the following, to allow users to add in their own checksums if they'd like:

```go
if g.ChecksumPresent {
  if opts.ComputeChecksums {
    g.Checksum = actuallyComptue()
  }
  // put in checksum regardless, allowing users to set their own checksum if they so desire.
  binary.BigEndian.PutUint16(buf[4:6], g.Checksum)
} else {
  // if not present, zero it out
  buf[4], buf[5] = 0, 0
}
```

layers/gre.go
```diff
@@ -106,6 +105,9 @@ func (g *GRE) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOpt
 	if err != nil {
 		return err
 	}
+	// Reset any potentially dirty memory in the first 2 bytes, as these use OR to set flags.
```
gconnell: ooo, great catch!

layers/gre_test.go
larit: Yes, the second one was from my desktop, though I changed the outer headers that don't affect the GRE checksum.

Because the checksums in the lower level packets cause them to cancel out, only the GRE header actually affect the value so there are a limited number of values it can take, basically just the protocol field affect it in these packets. Trying to capture wild packets with other options is hard.

gconnell: Good enough for me :)

larit:
I think I've addressed everything, can you have another look?

PR #198, opened 2016-05-25
Add serialization support for the DNS layer

tzneal:


layers/ip4.go
```diff
@@ -10,9 +10,10 @@ package layers
 import (
 	"encoding/binary"
 	"fmt"
-	"github.com/google/gopacket"
```
tzneal: goimports did this reordering, I didn't bother to undo it.

gconnell:
Thanks!

PR #199, opened 2016-05-28
Fix packet size calculation and record serialization

tzneal:
I noticed issues when capturing serialized packets with wireshark.
I added a test for the additional records that weren't being serialized
and added support for NS records to implement the test.

There is no test for the packet size calculation, but it would panic
if we were not reserving enough space. Previously we were reserving
a few bytes per record too much.


lhausermann:
And with the fix of PR#203 ?

tzneal:
Sorry, I never saw #203.  It looks like this is a subset of those fixes, I'll close it.

PR #200, opened 2016-06-01
Add support for DHCPv4

tzneal:
Started by Vasiliy Tolstov <v.tolstov@selfip.ru>

### DIFF
--- a/layers/dhcpv4.go
+++ b/layers/dhcpv4.go
@@ -1,0 +1,560 @@
+// Copyright 2016 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"net"
+
+	"github.com/google/gopacket"
+)
+
+// DHCPOp rerprents a bootp operation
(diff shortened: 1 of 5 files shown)


vtolstov:
Thanks!

tzneal:
Thanks, I didn't mean for those to be exported, fixed now.

tzneal:
Is there something else needed here?

lhausermann:
Sorry for the delay of answer. The only point I see is why do we need to allocate a buffer for decoding DHCP Options. Please explain or fix it using a slice. 
I'll merge right away.

layers/dhcpv4.go
```diff
@@ -0,0 +1,563 @@
...
+			if err := o.encode(data[offset:]); err != nil {
+				return err
+			}
+			offset += 2 + len(o.Data)
+		}
+		optend := NewDHCPOption(DHCPOptEnd, nil)
+		if err := optend.encode(data[offset:]); err != nil {
+			return err
+		}
+	}
```
tzneal: Unless I'm missing something, I already fixed that here.  The options encode directly to the data buffer.

lhausermann: I was not talking about Encode() but DecodeFromBytes() which is still making a new buffer and write in it.

tzneal: Sorry, I missed that, fixed now.

PR #205, opened 2016-06-11
Write a "NULL" SRE to terminate GRE routing field

kaorimatz:
The GRE decoding implementation in gopacket discards the NULL SRE so it would be better to terminate the routing field with a NULL SRE when serializing.


lhausermann:
I am not an expert of GRE. 
For the history, is it possible to add RFC référence (with adhoc section) to make very clear why we need that.
Thanks

kaorimatz:
Sorry for not putting the reference in the description. This is explained in RFC 1701 page 5.

https://tools.ietf.org/html/rfc1701#page-5

> The routing field is terminated with a "NULL" SRE containing an address family of type 0x0000 and a length of 0.

PR #206, opened 2016-06-16
add human readable TCP option display

tzneal:
Adds a readable version of TCP Option as well as constants for various
common TCP Option kinds.

### DIFF
--- a/layers/tcp.go
+++ b/layers/tcp.go
@@ -9,8 +9,10 @@ package layers
 
 import (
 	"encoding/binary"
+	"encoding/hex"
 	"errors"
 	"fmt"
+
 	"github.com/google/gopacket"
 )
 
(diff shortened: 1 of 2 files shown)


tzneal:
Ping.

lhausermann:
I would suggest to name it "TCPOptionType" and not "Kind" which is not very accurate.

tzneal:
I can change it, but I had named it from RFC 793 terminology:

    Options:  variable

    Options may occupy space at the end of the TCP header and are a
    multiple of 8 bits in length.  All options are included in the
    checksum.  An option may begin on any octet boundary.  There are two
    cases for the format of an option:

      Case 1:  A single octet of option-kind.

      Case 2:  An octet of option-kind, an octet of option-length, and
               the actual option-data octets.

    The option-length counts the two octets of option-kind and
    option-length as well as the option-data octets.

PR #212, opened 2016-07-06
implement FlushNoCloseOlderThan for better control over idle empty connections

postwait:
For sessions that have long lulls, it makes sense to keep them around in the pool as long as they aren't consuming too much packet buffer space.  This allows us to flush (without closing empty connections) with a tighter age control and the close them after a long time.  We typically don't see out-of-order packets arrive 5 seconds later (especially on a local segment), so we want to get the packets going if they are actually missing or we've picked up mid-stream.   But if we have a stream going and we've processed all the outstanding contiguous segments, we should be able to hold open the connection in wait for future packets.

e.g.

We can call FlushNoCloseOlderThan(t - 5 seconds) and FlushOlderThan(t - 5 minutes) to achieve this.


tcpassembly/assembly.go
```diff
@@ -256,6 +256,38 @@ func (a *Assembler) FlushOlderThan(t time.Time) (flushed, closed int) {
 	return flushes, closes
 }
 
+// This is much like FlushOlderThan, but doesn't close empty streams...
+// This is useful to get stream moving where don't expect long delays on
+// out-of-order packets, but you do expect long idle sessions or significant
+// time to elapse between a request and response between two correlated streams.
+func (a *Assembler) FlushNoCloseOlderThan(t time.Time) (flushed, closed int) {
```
gconnell: This is a great idea, but not too extensible.  How about the following:

```
type FlushOptions struct {
  T time.Time  // If nonzero, only connections with data older than T are flushed
  CloseAll bool  // If true, ALL connections are closed post flush, not just those that correctly see FIN/RST.
}

func (a *Assembler) Flush() { a.FlushWithOptions(FlushOptions{CloseAll: true}) }
func (a *Assembler) FlushOlderThan(t time.Time) { a.FlushWithOptions(FlushOptions{CloseAll: true, T: t}) }
```

Then, instead of a new FlushNoCloseOlderThan, you could just say `a.FlushWithOptions(FlushOptions{T: t})`

postwait:
Updated to use a FlushWithOptions approach as suggested.

tcpassembly/assembly.go
postwait: Addressed.

gconnell:
Sorry, spaced on this for a while.

PR #215, opened 2016-07-10
Add support for VRRPv2 decoding

jesseward:
Addition of a VRRP layer that provides decoding for VRRPv2 messages (rfc3768)


lhausermann:
Nice patch !
Everything looks ok for me. Merging it right away.

Thanks Jesse.

PR #216, opened 2016-07-10
Added additional records types to be formatted in DNSResourceRecord.String()

bramp:


lhausermann:
Great idea!
I'll merge it right away.

PR #217, opened 2016-07-28
Implement DecodingLayer interface for SCTP

kor44:
Now can use SCTP layer in NewDecodingLayerParser like TCP, UDP

### DIFF
--- a/layers/sctp.go
+++ b/layers/sctp.go
@@ -9,8 +9,9 @@ package layers
 import (
 	"encoding/binary"
 	"fmt"
-	"github.com/google/gopacket"
 	"hash/crc32"
+
+	"github.com/google/gopacket"
 )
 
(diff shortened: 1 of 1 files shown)


kor44:
I signed it!

lhausermann:
Hi. Thanks for your contribution. Looks like you added a test for payload length less than 12. Is it a normal case ? If so, please add a unit test. If not, please trigger an error.

Laurent

kor44:
Hi. Can you clarify line number: 71 or 80?

lhausermann:
To be honest, I didn't spot the line 71... which is great. In fact, line 80 is useless. Please just do the safety check first and next consider you are above 12.

kor44:
You are right. I removed this verificaction

PR #220, opened 2016-08-11
TSO support fix

noricube:
If using TSO(TCP Segmentation Offload), the packet length will be zero.
But gopacket considers the packet as an invalid packet.

This PR will check the TSO packets, and change to the actual data length.

### DIFF
--- a/layers/ip4.go
+++ b/layers/ip4.go
@@ -188,6 +188,8 @@ func (ip *IPv4) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 	// Set up an initial guess for contents/payload... we'll reset these soon.
 	ip.BaseLayer = BaseLayer{Contents: data}
 
+	// This code is added for the following enviroment:
+	// * Windows 10 with TSO option activated. ( tested on Hyper-V, RealTek ethernet driver )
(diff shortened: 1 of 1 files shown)


noricube:
I signed it!

lhausermann:
Hi 

Thanks for your patch.
Can you please provide the command you are using to setup TSO (or check it's enabled)?
What kind of ethernet driver are you using?

I am a bit surprised that the driver do not put the right size, and I do not want to remove a safety check if it's not required.

Laurent

noricube:
Thanks for reply.

I’m using Windows 10 and winpcap.
TSO supports both Hyper-v Ethernet driver and Realtek drivers. (below images)

image

image

Also TSO option is enabled by default.

On Wireshark, header length seems to zero.

image

Thanks

lhausermann:
Hi

For the clarity of the source code, I would ask you to add a comment for this before merging. I suggest you add a reference to your setup (Windows 10 + WinPCAP + Hyper-V + RealTek + TSO)
It will ease understanding for next review, as I assurme other platform such as Linux does put the right size even when TSO is setup

Laurent

noricube:
I added it.

PR #221, opened 2016-08-12
Have packet read return new packet buffers.

gconnell:
Returning a shared buffer makes handling multiple packets from a single
file in parallel impossible, and it breaks the ReadPacketData contract.
We could alternatively build a ZeroCopyReadPacketData for file reading,
should this change be too slow for some folks.

### DIFF
--- a/pcapgo/read.go
+++ b/pcapgo/read.go
@@ -121,7 +121,7 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
 		return
 	}
 	if ci.CaptureLength > int(r.snaplen) {
-		err = fmt.Errorf("capture length exceeds snap length: %d > %d", 16+ci.CaptureLength, len(r.buf))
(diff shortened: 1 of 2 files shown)


gconnell:
I signed it.

pcapgo/read.go
```diff
@@ -121,27 +120,17 @@ func (r *Reader) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
...
-
-	var n int
-	if 16+ci.CaptureLength > len(r.buf) {
-		err = fmt.Errorf("capture length with header exceeds buffer size: %d > %d", 16+ci.CaptureLength, len(r.buf))
-		return
-	}
-	data = r.buf[16 : 16+ci.CaptureLength]
-	if n, err = io.ReadFull(r.r, data); err != nil {
+	if ci.CaptureLength > int(r.snaplen) {
+		err = fmt.Errorf("capture length exceeds snap length: %d > %d", 16+ci.CaptureLength, len(r.buf))
```
mspielberg: len(r.buf) is always 16. Should be r.snaplen.

mspielberg:
Other than the incorrect error message, this LGTM.

PR #227, opened 2016-08-30
some sctp changes

tzneal:


lhausermann:
Looks very good patch. Thanks for contributing to GoPacket.

PR #228, opened 2016-09-15
Function to use an open file for pcap input

shane-kerr:
I recently needed to use stdin as a pcap source. The underlying libpcap supports this, so I added a function to do this to gopacket.


gconnell:
Instead of this, consider using pcapgo/NewReader?

https://godoc.org/github.com/google/gopacket/pcapgo#NewReader

shane-kerr:
Okay, I've refactored the code to use pcapgo/NewReader.

Issue #231, opened 2016-09-21
ip header checksum calculation is wrong


nathandfox:
https://github.com/google/gopacket/blob/master/layers/ip4.go#L150

the current code is wrong in the case of addition of the carry

https://play.golang.org/p/z8jcO1yUC4

nathandfox:
Submitted pull request

https://github.com/google/gopacket/pull/232

gconnell:
I believe this issue has been fully fixed.

PR #232, opened 2016-09-22
Fix ip header checksum bug

nathandfox:
https://github.com/google/gopacket/issues/231

For some header the original calculation was wrong

https://play.golang.org/p/z8jcO1yUC4


nathandfox:
I signed it!

gconnell:
Code looks good.  Could you also add a test for this?

PR #241, opened 2016-10-19
check length of data in decodeName

hkparker:
When parsing malformed DNS, possible index out of range here.  Discovered this trying to parse an iodine session.

### DIFF
--- a/layers/dns.go
+++ b/layers/dns.go
@@ -436,6 +436,9 @@ func decodeName(data []byte, offset int, buffer *[]byte, level int) ([]byte, int
 		return nil, 0, errMaxRecursion
 	}
 	start := len(*buffer)
+	if len(data) < offset {
+		return nil, 0, fmt.Errorf("data length %d less than offset %d", len(data), offset)
+	}
 	index := offset
 	if data[index] == 0x00 {
 		return nil, index + 1, nil


hkparker:
I am not interested in signing a license agreement so I can open a pull request.  Holy crap.

gconnell:
Sorry that didn't work out for you.  Unfortunately, to receive any code, we
need to have a CLA that allows us to use said code.

hkparker:
@gconnell Is that really a legal requirement?  How are other open sourced projects able to operate without CLAs if thats true?  Is there any way I can just slap and MIT license on this diff so there is no doubt its free to use?  If I understand it correctly merging this would be receiving code from another 3-clause BSD licensed fork of the same codebase, so Google has 0 obligation to use it in any particular way.

I'm genuinely curious about why this is a requirement, not trying to be argumentative, so please answer my questions if you have the time.  But I'm not really interested in trying to get this merged, I'll just merge future changes into my fork and continue to use mine.

gconnell:
IANAL, but here's my understanding of things.  Hope it helps :)

When you've authored your code snippet, you've done a creative act.  If you've done it under company time of any sort, that grants copyright of the snippet to your employer.  If not, it grants one to you.  I believe the latter is the case, here.  By owning the copyright to the material you've sent, you have the basis to make demands on what's done with it.

Transfering that copyright could be difficult and hard to automate (I'm not sure exactly how it would work).  So instead, Google's CLA says "keep your copyright, just please tell us that we can use the code as we see fit."  You can read over https://cla.developers.google.com/about/google-individual if you want, but in short, it never removes your copyright from your code... it remains yours.  It DOES, however, grant Google the right to use that code as it wants, including redistributing it, etc.  For example, section 2 gives us a copyright license (https://en.wikipedia.org/wiki/Public_copyright_license).  It doesn't give us the copyright, but it gives us legal authority to use the code.

In short, we live in a world where, were we to accept code without this, you MIGHT have legal standing to come back later and make demands on how we use it.  And that's really scary.

As to the question of why it's not a requirement for pure open-source stuff, I don't know.  My guess is that, in those cases, you're not providing code to a legal entity (a corporation).  Instead, you're becoming part of a group, which collectively decides what to do with the code.  Again, IANAL and that's just a guess.  Note that as open source projects get larger, some of them DO start using CLAs, see for example Apache's at https://www.apache.org/licenses/icla.txt.

Unfortunately, I really can't accept non-CLA'd code.  But I've read over Google's CLA a few times, and I like what I see there.  We don't take ownership or copyright of code, we simply request that you allow us to use it like we want later on.  And I think that's a pretty good way to go about things :)

Issue #165, opened 2016-02-04
init() in layers eat too much memory


hilyjiang:
I found in `init()` a lot of slice initialization operations.
https://github.com/google/gopacket/blob/487c0cc416754a86af4a35006a8d0e0174f75629/layers/enums.go#L379

is it necessary to init so much slice?
Or can it be done when need?

lhausermann:
Hi

The decoding flow is based on this slice : for a EthernetType we use it to find the right function.
As there is a lot of potential EthernetType, we are initializing all the potential ones to trigger a right error message.

Would you explain how you reach 20MB memory for this? It looks a very large amount of memory.

hilyjiang:
This my test code:
```golang
package main

import (
        _ "github.com/google/gopacket/layers"
)

func main() {
        ch := make(chan bool)
        <- ch
}
```

Build it and run, the memory usage as follow:

>   PID USER      PR  NI  VIRT  RES  SHR S %CPU %MEM    TIME+  COMMAND                                                                                                                     
> 23035 root      20   0  161m  26m 1892 S  0.0  2.7   0:00.27 gopacket 
>

lhausermann:
Well, I understand your concern but on another way golang itself was never designed to be low in memory consumption.

Is there a specific usage you are trying to make working? So low ressources board?

hilyjiang:
We are doing a small agent installed on a large amount of small memory size machine.
So the memory usage is indeed a problem.

I am now temporarily fork and remove the init() when I only use the layers package.
Is it possible to initialize the big slices when I really need it?

lhausermann:
What you can do is to write you own Decode/LayerType/String method for EthernetType and PPPType.
The slide is only used in layers/enums.go if I am not wrong, and is a way for a thrid party dev to add it's own layers in the stack.  

In your use case, you probably want to change that to have a fix set of layers.

bronze1man:
relative to #101

lukedirtwalker:
This can probably be closed. Since https://github.com/google/gopacket/commit/8c5cfc5ae25c08c58287b22231251f3f9c155ec6 the memory usage dropped considerably.

notti:
I'm not completely sure if this is actually resolved. 8c5cfc5ae25c08c58287b22231251f3f9c155ec6 moves the stuff from the heap to bss and text (and maybe even use more due to this layout - not completely sure how to check that quickly now).
[...]

PR #242, opened 2016-10-20
Allow to register UDP underlaying layertypes

eonpatapon:
Add a simple interface to register layertypes for a particular src or dst UDP port

### DIFF
--- a/layers/ports.go
+++ b/layers/ports.go
@@ -61,20 +61,20 @@ func (a UDPPort) String() string {
 //
 // Returns gopacket.LayerTypePayload for unknown/unsupported port numbers.
 func (a UDPPort) LayerType() gopacket.LayerType {
-	switch a {
-	case 53:
-		return LayerTypeDNS
-	case 123:
-		return LayerTypeNTP
-	case 4789:
-		return LayerTypeVXLAN
-	case 67, 68:
-		return LayerTypeDHCPv4
-	case 6343:
-		return LayerTypeSFlow
-	default:
-		return gopacket.LayerTypePayload
+	lt := udpPortLayerType[uint16(a)]
+	if lt != 0 {
+		return lt
 	}
+	return gopacket.LayerTypePayload
+}
+
+var udpPortLayerType [65536]gopacket.LayerType
+
+// RegisterUDPPortLayerType create a new mapping between an UDPPort
+// and an underlaying LayerType.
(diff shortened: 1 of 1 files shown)


eonpatapon:
I signed it!

vtolstov:
Nice! +1

lhausermann:
Great idea (as I am the orignal author of a switch/case which has grown)

layers/ports.go
```diff
@@ -61,20 +61,20 @@ func (a UDPPort) String() string {
...
+		return lt
 	}
+	return gopacket.LayerTypePayload
+}
+
+var udpPortLayerType [65536]gopacket.LayerType
+
+// RegisterUDPPortLayerType create a new mapping between an UDPPort
+// and an underlaying LayerType.
+func RegisterUDPPortLayerType(portNum uint16, layerType gopacket.LayerType) uint16 {
```
gconnell: Please accept a layers.UDPPort instead of a uint16 here.

I'd also maybe remove the return value... doesn't seem too necessary/useful.

layers/ports.go
```diff
@@ -106,3 +106,12 @@ func (a UDPLitePort) String() string {
 	}
 	return strconv.Itoa(int(a))
 }
+
+func init() {
```
gconnell: instead of doing this in init, maybe build udpPortLayerType with this already in place?

```go
var dupPortLayerType = [65536]gopacket.LayerType{
  53: LayerTypeDNS,
  123: LayerTypeNTP,
  ...
}
```

layers/ports.go
```diff
@@ -61,20 +61,20 @@ func (a UDPPort) String() string {
...
-	default:
-		return gopacket.LayerTypePayload
+	lt := udpPortLayerType[uint16(a)]
+	if lt != 0 {
+		return lt
 	}
+	return gopacket.LayerTypePayload
+}
+
+var udpPortLayerType [65536]gopacket.LayerType
```
gconnell: Note, there's already some complaints about init taking up too much memory (https://github.com/google/gopacket/issues/165).  I'm considering the performance hit of making everything maps instead of arrays.  I'd say this is a fine way to do things for now, though, just something to keep in mind.

eonpatapon:
Thanks for your input, I will send a new patch to improve things

PR #243, opened 2016-10-20
Improve UDPPort LayerType mapping code

eonpatapon:
Follow up patch on #242 comments


gconnell:
Looks great!

For future reference, instead of deleting the old branch and creating a new one, just keep adding commits to the branch associated with the pull request, and they'll show up on the initial PR as subsequent changes.

gconnell:
Oh, never mind, I didn't see that Laurent had already merged that first PR.  My bad :)

PR #247, opened 2016-10-28
DHCPv4 fix option length

jerome-laforge:
According the RFC below.
Option has length of 0 or 255 bytes.
```rfc
2. BOOTP Extension/DHCP Option Field Format


   DHCP options have the same format as the BOOTP 'vendor extensions'
   defined in RFC 1497 [2].  Options may be fixed length or variable
   length.  All options begin with a tag octet, which uniquely
   identifies the option.  Fixed-length options without data consist of
   only a tag octet.  Only options 0 and 255 are fixed length.  All
   other options are variable-length with a length octet following the
   tag octet.  The value of the length octet does not include the two
   octets specifying the tag and length.  The length octet is followed
   by "length" octets of data.  Options containing NVT ASCII data SHOULD
   NOT include a trailing NULL; however, the receiver of such options
   MUST be prepared to delete trailing nulls if they exist.  The
   receiver MUST NOT require that a trailing null be included in the
   data.  In the case of some variable-length options the length field
   is a constant but must still be specified.

   Any options defined subsequent to this document MUST contain a length
   octet even if the length is fixed or zero.
```


jerome-laforge:
Bad pull request.

PR #248, opened 2016-10-28
DHCPv4 fix option length

jerome-laforge:
According the RFC below.
Option has length of 0 or 255 bytes.
```
2. BOOTP Extension/DHCP Option Field Format


   DHCP options have the same format as the BOOTP 'vendor extensions'
   defined in RFC 1497 [2].  Options may be fixed length or variable
   length.  All options begin with a tag octet, which uniquely
   identifies the option.  Fixed-length options without data consist of
   only a tag octet.  Only options 0 and 255 are fixed length.  All
   other options are variable-length with a length octet following the
   tag octet.  The value of the length octet does not include the two
   octets specifying the tag and length.  The length octet is followed
   by "length" octets of data.  Options containing NVT ASCII data SHOULD
   NOT include a trailing NULL; however, the receiver of such options
   MUST be prepared to delete trailing nulls if they exist.  The
   receiver MUST NOT require that a trailing null be included in the
   data.  In the case of some variable-length options the length field
   is a constant but must still be specified.

   Any options defined subsequent to this document MUST contain a length
   octet even if the length is fixed or zero.
```


layers/dhcpv4.go
```diff
@@ -540,21 +536,38 @@ func (o *DHCPOption) encode(b []byte) error {
...
+
+// DHCPv4Error is used for constant errors for DHCPv4. It is needed for test asserts.
+type DHCPv4Error string
+
+// DHCPv4Error implements error interface.
+func (d DHCPv4Error) Error() string {
+	return string(d)
+}
+
+const (
```
lhausermann: I have two comments on this : 
1 - As the whole namespace "layers" is quite crowded, I would suggest to rename these errors with "DHCP" prefix.
2 - Adding errors as const will allocate them in any case. I know that @gconnell is looking to keep the whole footprint of gopacket low, so maybe we should not do it. @gconnell any idea on that ?

lhausermann:
See my comment on the diff

layers/dhcpv4.go
jerome-laforge: 1 - Ok, for the prefix.
2 - I am waiting for the suggest/directive from @gconnell and I'll propose a new PR.

PR #259, opened 2016-12-02
Add string function for DNSClass, DNSType and DNSOpCode

denisvm:
Just added string functions for DNSClass, DNSType and DNSOpCode.


denisvm:
I signed it!

lhausermann:
Thanks Denis ! Merged

PR #266, opened 2016-12-09
ip4defrag: flush fragment list and ignore duplicates

sentryo:
This commit fixes two major issues.

The first bug was that the fragment list per IPv4 flow was *never*
flushed. We are now flushing it after each successful packet rebuild.

The second one was that when we received a duplicate fragment, we were
considering it as a valid one. We are now ignoring any new fragment at
the same offset as an earlier one.

Also, include some cosmetic changes from our internal fork.

### DIFF
--- a/ip4defrag/defrag.go
+++ b/ip4defrag/defrag.go
@@ -206,6 +206,18 @@ func (f *fragmentList) insert(in *layers.IPv4) (*layers.IPv4, error) {
 		for e := f.List.Front(); e != nil; e = e.Next() {
 			frag, _ := e.Value.(*layers.IPv4)
 			if in.FragOffset == frag.FragOffset {
+				// TODO: what if we receive a fragment
+				// that begins with duplicate data but
+				// *also* has new data? For example:
+				//
+				// AAAA
+				//     BB
+				//     BBCC
+				//         DDDD
+				//
+				// In this situation we completely
+				// ignore CC and the complete packet can
+				// never be reassembled.
 				debug.Printf("defrag: ignoring frag %d as we already have it (duplicate?)\n",
 					fragOffset)
 				return nil, nil
--- a/ip4defrag/defrag.go
+++ b/ip4defrag/defrag.go
@@ -21,7 +21,7 @@ import (
 
 // Quick and Easy to use debug code to trace
 // how defrag works.
-var debug debugging = false // or flip to false
+var debug debugging = false // or flip to true
 type debugging bool
 
 func (d debugging) Printf(format string, args ...interface{}) {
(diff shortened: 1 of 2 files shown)


ip4defrag/defrag.go
```diff
@@ -188,15 +198,20 @@ type fragmentList struct {
...
+	// or not (ie the packet source is reliable) ? -> depends on Lazy / last packet
 	fragOffset := in.FragOffset * 8
 	if fragOffset >= f.Highest {
 		f.List.PushBack(in)
 	} else {
 		for e := f.List.Front(); e != nil; e = e.Next() {
 			frag, _ := e.Value.(*layers.IPv4)
-			if in.FragOffset <= frag.FragOffset {
-				debug.Printf("defrag: inserting frag %d before existing frag %d \n",
+			if in.FragOffset == frag.FragOffset {
```
gconnell: I'm not sure if we actually want to do this.  Consider this set of fragments:

```
AAAA
    BB
    BBCC
        DDDD
```

With the new logic, we'll never see bytes CC, and we'll never be able to reconstruct the packet.  Thoughts?

lhausermann: Well, I think there is many different strategies on this. See http://www.sans.org/reading-room/whitepapers/detection/ip-fragment-reassembly-scapy-33969 I've already referenced in the code.

Our original issue we had is that the duplicates segment were not correctly taken into account (see the added test).

Moreover, in this patch there is the flush of the list, which is a big bug

gconnell: It's important to note that I'm not talking about overlaps here.  I'm talking about two packets with the same initial data, one of which is longer than the other.  Note that BB is the same in both, so regardless of the overlap strategy used (figure 2:5 in the linked doc), the "correct" output is AAAABBCCDDDD.  I just worry that this code will result in AAAABBxxDDDD, where x is "missing and unrecoverable".

Also, is your "which is a big bug" pointing out a bug in the current code, or a fix in that code?  Sorry, I'm just unclear.  Line number?

lhausermann: Line 121, add a call to the new flush method.

Well, you are right on the case you mention. It would imply to re-read the list to replace the first segment with the newer...

NormanDenayer: Hello,
I'm "blocked" in my current development because of the "flushing bug", solved in this patch.
Can I help in anyway to move on with this pending PR?

Thank you

gconnell: Maybe add a //BUG or //TODO here detailing the case I mentioned, and we can submit as is?  Or were you planning on doing a fix for that?

NormanDenayer: @sentryo, Can you add a //BUG with remark from @gconnell  to unblock this PR?
Or do you prefer I create a separate PR to address the flushing bug, and keep here the debate about the reconstruction order?

lhausermann: Let us (ie @sentryo team) have a look next Monday/Tuesday

NormanDenayer: thank you @sentryo @gconnell  :-)

PR #270, opened 2016-12-14
Add Geneve tunneling protocol support

safchain:
Introduce Geneve tunneling protocol
described in :

https://tools.ietf.org/html/draft-ietf-nvo3-geneve-03


layers/geneve.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2016 Red Hat, Inc. All rights reserved.
```
gconnell: Please use copyright from the rest of the files (IE: google, inc)

layers/geneve_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2016 Red Hat, Inc. All rights reserved.
```
gconnell: ditto here

safchain:
I changed the copyright lines, thanks

PR #272, opened 2016-12-15
Fix possible IP comparison error

karalabe:
A net.IP may be represented by both by a 4 as well as a 16 byte long byte slice. Because of this, it is not safe to compare IP addresses using bytes.Equal as the same IP address using a different internal representation will produce mismatches.

*Note, this PR was not initiated by my use of this package and noticing a bug, rather by creating an extension proposal for `go vet` and that signalling the possible misuse in this library.*

Cheers :)


gconnell:
Thanks :)

PR #282, opened 2017-01-09
Add IPv6 defrag function

songlinjian:
I would like to submit this pull request to add ipv6 defragment function to gopacket.  Basically, it will work like drag.go in ip4drag except it will gopacket.Packet as input to get both IPv6 layer and IPv6Fragment layer.

Note that my colleague RunxiaWan did the initial pull request but his Github account is not listed in the CLA I signed. So I re-do the pull request.


ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
+// Package ip6defrag implements a IPv6 defragmenter
+package main
```
gconnell: package should be 'ip6defrag', not 'main'

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
+// Package ip6defrag implements a IPv6 defragmenter
```
gconnell: Please add standard Copyright boilerplate to top of file:
```
// Copyright 2013 Google, Inc. All rights reserved.
//
// tree.
```

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+package main
+
+import (
+	"container/list"
+	"fmt"
+	"github.com/google/gopacket"
```
gconnell: Please separate standard-lib imports from gopacket imports, see https://github.com/google/gopacket/blob/master/ip4defrag/defrag.go for an example

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/google/gopacket/layers"
+	"log"
+	"sync"
+	"time"
+)
+
+//need to check
```
gconnell: Spurious comment?  This comment will actually show up in godoc, since it's at the start of a const block, so please fix.  Also, please actually do check :)

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+	ipFlows map[ipv6]*fragmentList
+}
+
+type ipv6 struct {
+	ip6 gopacket.Flow
+	id  uint32
+}
+
+// newIPv6 returns a new initialized IPv6 Flow
+func newIpv6(packet gopacket.Packet) ipv6 {
```
gconnell: newIPv6

when one part of an acronym is capitalized, all of it should be.

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+	frag := packet.Layer(layers.LayerTypeIPv6Fragment).(*layers.IPv6Fragment)
+	ip := packet.Layer(layers.LayerTypeIPv6).(*layers.IPv6)
+	return ipv6{
+		ip6: ip.NetworkFlow(),
+		id:  frag.Identification,
+	}
+}
+
+func (d *IPv6Defragmenter) DefragIPv6(in gopacket.Packet) (gopacket.Packet, error) {
```
gconnell: Please add godoc here

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+	d.Lock()
+	fl, exist = d.ipFlows[ipf]
+	if !exist {
+		debug.Printf("defrag: creating a new flow\n")
+		fl = new(fragmentList)
+		d.ipFlows[ipf] = fl
+	}
+	d.Unlock()
+	// insert, and if final build it
+	out, err2 := fl.insert(in)
```
gconnell: pass in v6frag here so you don't need to grab it again?

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+		f.FinalReceived = true
+	}
+	// Ready to try defrag ?
+	if f.FinalReceived && f.Highest == f.Current {
+		return f.build(in)
+	}
+	return nil, nil
+}
+
+//need work
```
gconnell: remove comment... do work?  :D

ip6defrag/defrag.go
```diff
@@ -0,0 +1,227 @@
...
+			// Houston - we have an hole !
+			debug.Printf("defrag: hole found while building, " +
+				"stopping the defrag process\n")
+			return nil, fmt.Errorf("defrag: building - hole found")
+		}
+		debug.Printf("defrag: building - next is %d\n", currentOffset)
+	}
+	final_ipv6 := in.Layer(layers.LayerTypeIPv6).(*layers.IPv6)
+	final_frag := in.Layer(layers.LayerTypeIPv6Fragment).(*layers.IPv6Fragment)
+	// TODO recompute IP Checksum
```
gconnell: I think this is done, line 214?

RunxiaWan: Hi, I am the code author. The reason I choose to build a packet and return it is that I use packet as input because I need get both IPv6 layer and IPv6 Fragmentation layer so I think it is more symmetrical to return the same data format as input. But if you think just return IPv6 layer is better, then I can change the code

RunxiaWan: Hi, I am the code author. The reason I choose to build a packet and return it is that I use packet as input because I need get both IPv6 layer and IPv6 Fragmentation layer so I think it is more symmetrical to return the same data format as input. But if you think just return IPv6 layer is better, then I can change the code

lhausermann:
Closing this PR has no progress since long time.

PR #283, opened 2017-01-13
avoid call to header.next after poll error

safchain:
In case of a poll error it is a mistake to call
next to the header. This patch introduces a
variable that protect this call.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -86,6 +86,8 @@ type TPacket struct {
 	pollset unix.PollFd
 	// shouldReleasePacket is set to true whenever we return packet data, to make sure we remember to release that data back to the kernel.
 	shouldReleasePacket bool
+	// headerNextNeeded is set to true when header need to move to the next packet. No need to move it case of poll error.
+	headerNextNeeded bool
 	// stats is simple statistics on TPacket's run.
 	stats Stats
 	// socketStats contains stats from the socket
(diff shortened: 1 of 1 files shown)


afpacket/afpacket.go
```diff
@@ -248,16 +250,17 @@ func (h *TPacket) releaseCurrentPacket() error {
 func (h *TPacket) ZeroCopyReadPacketData() (data []byte, ci gopacket.CaptureInfo, err error) {
 	h.mu.Lock()
 retry:
-	if h.current == nil || !h.current.next() {
+	if h.current == nil || !h.headerNextNeeded || !h.current.next() {
 		if h.shouldReleasePacket {
 			h.releaseCurrentPacket()
 		}
 		h.current = h.getTPacketHeader()
```
gconnell: I think you've definitely found some code that needs some work, but I'm not sure if this solution actually addresses the main concern I have with it, after looking closely.

From what I can see, the problem isn't really with the .next() call (althought that is an issue), it's actually with the getTPacketHeader() call, which advances h.current to the next block.  Should poll error out, there's a chance that we'll miss packets, since h.getTPacketHeader() will be called twice, advancing _past_ a block that could have packets in it.

I'm going to go back and take a look at this code myself, and maybe together we can rejigger things to handle poll errors better.  Thanks for bringing this up!

safchain: @gconnell thx, you're right code around getTPacketHeader needs to be reworked too. I'll try to have a look too.

safchain: @gconnell it seems to me that getTPacketHeader returns a new position depending on h.offset and h.offset is moved forward by releaseCurrentPacket which is protected by shouldReleasePacket. shouldReleasePacket is set to true only when on poll success ?

gconnell: Hmm, maybe past-me isn't as dumb as he looks.  Let me take another close look at this change, and I'll try to get back to you in a day or two.

safchain: Hi @gconnell, any news on this ? have you had a chance to take a look ?

lebauce:
Hi @gconnell. We are heavily using gopacket in @skydive-project. Could you please have an other look at the PR ? Thanks !

gconnell:
Sorry, I totally spaced on this.  You're good to go!

gconnell:
Also, I"ve added you two as collaborators.  Great power, great responsibility, all that (in short, don't make backwards-incompatible changes).  But that should allow you to bugfix faster while keeping in the master branch.  Skydive looks awesome, btw :D

safchain:
@gconnell Great ! thanks for merging and for adding us as collaborators, we will use our new power carefully :)

PR #292, opened 2017-02-09
This commit removes a lot of the CGo used by AF_PACKET

levigross:
It does this by enabling afpacket to make raw set/getsockopt syscalls and
making use of the golang.org/x/unix syscall interface.

Signed-off-by: Levi Gross <levi@levigross.com>

### DIFF
--- a/afpacket/sockopt_linux.go
+++ b/afpacket/sockopt_linux.go
@@ -5,6 +5,7 @@
 // tree.
 
 // +build linux
+
 package afpacket
 
 import (
--- a/afpacket/sockopt_linux.go
+++ b/afpacket/sockopt_linux.go
@@ -1,0 +1,50 @@
+package afpacket
+
+import (
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+// setsockopt provides access to the setsockopt syscall.
+func setsockopt(fd, level, name int, val unsafe.Pointer, vallen uintptr) error {
+	_, _, errno := unix.Syscall6(
+		unix.SYS_SETSOCKOPT,
+		uintptr(fd),
+		uintptr(level),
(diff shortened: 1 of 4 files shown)


afpacket/sockopt_linux.go
```diff
@@ -0,0 +1,50 @@
+package afpacket
```
gconnell: need copyright here

afpacket/sockopt_linux_386.go
```diff
@@ -0,0 +1,55 @@
+// Copyright 2014 The Go Authors. All rights reserved.
```
gconnell: Please copy copyright from other gopacket files.

afpacket/sockopt_linux_386.s
```diff
@@ -0,0 +1,6 @@
+// Copyright 2014 The Go Authors. All rights reserved.
```
gconnell: Please copy from other files.

afpacket/sockopt_linux.go
```diff
@@ -1,3 +1,10 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+// +build linux
 package afpacket
```
gconnell: sorry, golint is requesting a space between the build comment and the package line. :)  https://travis-ci.org/google/gopacket/builds/200166979

levigross:
@gconnell I have updated the copyright. Please review and confirm that everything is kosher. As for removing CGo, I would like to not pay the CGo penalty so where I can get away with it I am happy to :)

gconnell:
Merged, thanks!

PR #293, opened 2017-02-10
add ability to compile pcap filters offline

tzneal:
Use pcap_open_dead to create a handle and then compile the filter.
According to the docs this is what pcap_compile_nopcap does internally,
but this method allows extracting an error message which
pcap_compile_nopcap swallows.


gconnell:
Thanks!

PR #295, opened 2017-02-10
Revert "This commit removes a lot of the CGo used by AF_PACKET"

levigross:
I am really sorry about this, but after doing some testing on the new AF_PACKET modifications I have decided to revert the work due to bugs introduced by my PR yesterday. I will run more rigorous tests on the code before resubmitting the PR.

Thank you for your understand and patience,
Levi

This reverts commit 5527bfcadcaabe78fa8db887e997d3a71171d425.

PR #296, opened 2017-02-10
Fixed some bugs in the new AF_PACKET implementation

levigross:
I found some bugs in the AF_PACKET code that I introduced yesterday. I have
fixed and tested these fixes.

Signed-off-by: Levi Gross <levi@levigross.com>

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -22,17 +22,23 @@ import (
 	"time"
 	"unsafe"
 
-	"github.com/google/gopacket"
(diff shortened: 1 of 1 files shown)


levigross:
This is a follow up from https://github.com/google/gopacket/pull/295. I have fixed the issues with this update. I am really sorry for the PR headache

gconnell:
No worries, happens to the best of us :)

If you have the inclination, we should probably set up an integration test for afpacket, something like https://github.com/google/stenographer/blob/master/integration_test/test.sh

PR #302, opened 2017-03-19
Register decodeDot11 for LinkTypeMetadata

Hjdskes:
In a (private) project of mine I have this code:

```go
	dot11Layer := packet.Layer(layers.LayerTypeDot11)
	if dot11Layer == nil {
		return
	}

	dot11 := &layers.Dot11{}
	data := dot11Layer.LayerPayload()
	dot11.DecodeFromBytes(data, gopacket.NilDecodeFeedback)
```

which failed to decode properly: `Packet decoding error: Unable to decode link type 105`. This change fixes that issue.

If the issue is a result of a mistake on my side and this commit is the wrong way around it, please do tell me.


Hjdskes:
I signed it!

PR #303, opened 2017-03-22
vxlan: add SerializeTo method

sirspinach:


sirspinach:
signed

gconnell:
Travis seems to think the code needs to be formatted... could you run 'go fmt'?

https://travis-ci.org/google/gopacket/builds/214010279

layers/vxlan.go
```diff
@@ -61,3 +62,33 @@ func decodeVXLAN(data []byte, p gopacket.PacketBuilder) error {
 	p.AddLayer(vx)
 	return p.NextDecoder(LinkTypeEthernet)
 }
+
+// SerializeTo writes the serialized form of this layer into the
+// SerializationBuffer, implementing gopacket.SerializableLayer.
+// See the docs for gopacket.SerializableLayer for more info.
+func (vx *VXLAN) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
+	bytes, err := b.PrependBytes(64)
```
gconnell: Note that this is PrependBYTES, not BITS.  It looks like you're only using 8 bytes here?  I'd expect:  b.PrependBytes(8)

I'll have to look back at the vxlan protocol's frame size to be sure, though...

sirspinach: that's right, it should be 8

PR #305, opened 2017-04-18
Determine endpoint IPv4/IPv6 type based on structure rather than byte-length

zcahana:
This PR modifies `layers.NewIPEndpoint(net.IP)` to properly inspect whether the input IP is an IPv4 or an IPv6, based on its structure/content, rather than by its `[]byte` slice length.
Prior to this PR, passing a 16-bytes IPv4 (e.g., `net.IPv4(192,168,0,1)`) created a `gopacket.Endpoint` of type IPv6.

According to the [`net.IP` documentation](https://golang.org/pkg/net/#IP), IPv4 addresses may be represented using 16-bytes:
> Note that in this documentation, referring to an IP address as an IPv4 address or an IPv6 address is a semantic property of the address, not just the length of the byte slice: a 16-byte slice can still be an IPv4 address.

The fix currently utilizes `net.IP`'s `.To4()`/`.To16()` methods to determine the type of the IP address. This can be micro-optimized by using `bytes.Compare(a[0:12], 0x0000000000ff)`, which is how 16-bytes IPv4 is represented by the `net` package. This is an internal and undocumented property, but is probably unlikely to change...

### DIFF
--- a/layers/endpoints.go
+++ b/layers/endpoints.go
@@ -50,12 +50,16 @@ var (
 // NewIPEndpoint creates a new IP (v4 or v6) endpoint from a net.IP address.
 // It returns gopacket.InvalidEndpoint if the IP address is invalid.
 func NewIPEndpoint(a net.IP) gopacket.Endpoint {
-	switch len(a) {
-	case 4:
-		return gopacket.NewEndpoint(EndpointIPv4, []byte(a))
-	case 16:
-		return gopacket.NewEndpoint(EndpointIPv6, []byte(a))
+	ipv4 := a.To4()
+	if ipv4 != nil {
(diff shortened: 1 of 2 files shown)


gconnell:
I'm okay with this change, and happy to see it go through.

Note:  If you have an IPv6 object, and you call:
```go
  flow := ip6.ENetworkFlow()
```
then call either `flow.Src()` or `flow.Dst()`, you'll _always_ get an EndpointIPv6 endpoint, even if the IPv6 address is in the https://tools.ietf.org/html/rfc5156#section-2.2 address space, because the flow is created with EndpointIPv6.

Since https://tools.ietf.org/html/rfc5156#section-2.2 addresses are not externally routable, I expect this isn't a huge issue.  It could cause some problems with equality checking/hashing, though.

PR #306, opened 2017-04-20
Avoid the double free of the bpf object

rouzier:
SetBPFFilter C.pcap_freecode may be called twice sometimes.


gconnell:
Thanks!

PR #307, opened 2017-04-22
Dhcp pad

rouzier:
Ensure the Pad option is treated as a single byte as according to https://tools.ietf.org/html/rfc2132


gconnell:
Thanks!

PR #308, opened 2017-04-25
af_packet SocketStats() stuck for waiting mutex

xuzhaokui:
Since ZeroCopyReadPacketData() always acquire the *TPacket.mu and then do the polling which is a blocking syscall，go scheduler may never give the chance to *TPacket.SocketStats() to get this lock.

So I add another mutex for protecting socket stats data fields.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -19,6 +19,7 @@ import (
 	"net"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


afpacket/afpacket.go
```diff
@@ -293,8 +295,8 @@ retry:
 
 // Stats returns statistics on the packets the TPacket has seen so far.
 func (h *TPacket) Stats() (Stats, error) {
-	h.mu.Lock()
-	defer h.mu.Unlock()
+	h.statsMu.Lock()
```
gconnell: You still want mu here instead of statsMu, I think, due to line 290 and 427 increment of the # of packets/polls.  Or you could consider replacing those increments with sync/atomic.AddInt64.

afpacket/afpacket.go
```diff
@@ -326,8 +328,8 @@ func (h *TPacket) InitSocketStats() error {
 
 // SocketStats saves stats from the socket to the TPacket instance.
 func (h *TPacket) SocketStats() (SocketStats, SocketStatsV3, error) {
-	h.mu.Lock()
-	defer h.mu.Unlock()
+	h.statsMu.Lock()
```
gconnell: Great idea!

xuzhaokui:
@gconnell I fixed it in atomic way.

BTW, I don't think its a good idea to spend another dozens of nanoseconds on this high frequency path, the performance of packet reading is more important than the precision of stats. :P

PR #310, opened 2017-05-03
Implement SerializeTo() for NTP layer

nogoegst:


layers/ntp.go
```diff
@@ -347,6 +347,43 @@ func (d *NTP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Pack the first few fields into the first 32 bits.
+	h := uint32(0)
+	h |= (uint32(d.LeapIndicator) << 30) & 0xC0000000
+	h |= (uint32(d.Version) << 27) & 0x38000000
+	h |= (uint32(d.Mode) << 24) & 0x07000000
+	h |= (uint32(d.Stratum) << 16) & 0x00FF0000
```
gconnell: for bytes 1-3, seems like you could just say:

```go
data[0] = ... byte packed from leap, version, mode ...
data[1] =d.Stratum
data[2] = d.Poll
data[3] = d.Precision
```

nogoegst: Heh, yes... Thanks! See #311.

gconnell:
Thanks again for the submission!  Really appreciate the work, and the code
is nice, clean, and well-tested.

PR #313, opened 2017-05-08
Implement SerializePacket

nogoegst:
I'm not sure whether `writer.go` is the right place for this.


gconnell:
Looks like as good a place as any to me ;)

PR #323, opened 2017-06-05
Reset IPv4 Options slice during new data decoding

miros:
Existing IPv4 slice might be reused to avoid new allocations (with DecodingLayerParser or standalone), therefore Options slice should be reset during DecodeFromBytes, otherwise previous ip options will remain in Options slice.


miros:
I signed it!

2017-06-05 13:15 GMT+03:00 googlebot <notifications@github.com>:

gconnell:
Thanks!

PR #324, opened 2017-06-06
Add raw layertype for google

zecke:
For Elasticsearch's packetbeat it is necessary to expose a LayerTypeRaw. Correct the LinkTypeRaw for mainstream platforms and add a LayerTpeRaw. Update the tests that use LinkTypeRaw right now.

### DIFF
--- a/layers/enums.go
+++ b/layers/enums.go
@@ -374,18 +374,6 @@ func (a Dot11Type) LayerType() gopacket.LayerType {
 	return Dot11TypeMetadata[a].LayerType
 }
 
-// Decode a raw v4 or v6 IP packet.
-func decodeIPv4or6(data []byte, p gopacket.PacketBuilder) error {
-	version := data[0] >> 4
-	switch version {
-	case 4:
-		return decodeIPv4(data, p)
-	case 6:
-		return decodeIPv6(data, p)
-	}
-	return fmt.Errorf("Invalid IP packet version %v", version)
-}
-
 func init() {
 	// Here we link up all enumerations with their respective names and decoders.
 	for i := 0; i < 65536; i++ {
@@ -501,9 +489,9 @@ func init() {
 	LinkTypeMetadata[LinkTypeIEEE802_11] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeDot11), Name: "Dot11"}
 	LinkTypeMetadata[LinkTypeLoop] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeLoopback), Name: "Loop"}
 	LinkTypeMetadata[LinkTypeIEEE802_11] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeDot11), Name: "802.11"}
-	LinkTypeMetadata[LinkTypeRaw] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeIPv4or6), Name: "Raw"}
-	LinkTypeMetadata[LinkTypeRawLinux] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeIPv4or6), Name: "Raw"}
-	LinkTypeMetadata[LinkTypeRawOpenBSD] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeIPv4or6), Name: "Raw"}
+	LinkTypeMetadata[LinkTypeRaw] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeRawIP), Name: "Raw"}
+	LinkTypeMetadata[LinkTypeRawLinux] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeRawIP), Name: "Raw"}
+	LinkTypeMetadata[LinkTypeRawOpenBSD] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeRawIP), Name: "Raw"}
 	LinkTypeMetadata[LinkTypePFLog] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodePFLog), Name: "PFLog"}
 	LinkTypeMetadata[LinkTypeIEEE80211Radio] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeRadioTap), Name: "RadioTap"}
 	LinkTypeMetadata[LinkTypeLinuxUSB] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeUSB), Name: "USB"}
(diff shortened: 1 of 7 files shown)


zecke:
> On 6. Jun 2017, at 23:44, googlebot <notifications@github.com> wrote:
> 

I signed it!

layers/raw.go
```diff
@@ -0,0 +1,68 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+// Copyright 2009-2011 Andreas Krennmair. All rights reserved.
```
gconnell: this second copyright notice isn't necessary on new files, just the Google line will do.

layers/raw_test.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2017 Holger Hans Peter Freyther, Google, Inc. All rights reserved.
```
gconnell: Please copyright to Google as in other files.

layers/raw_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+// testPacketPacket0 is the packet:
+//   12:13:21.217957 IP 10.8.254.8 > 8.8.8.8: ICMP echo request, id 20348, seq 1, length 64
+//   	0x0000:  4500 0054 1530 4000 4001 0d59 0a08 fe08  E..T.0@.@..Y....
+//   	0x0010:  0808 0808 0800 d81d 4f7c 0001 e1e5 3059  ........O|....0Y
+//   	0x0020:  0000 0000 fc52 0300 0000 0000 1011 1213  .....R..........
+//   	0x0030:  1415 1617 1819 1a1b 1c1d 1e1f 2021 2223  .............!"#
+//   	0x0040:  2425 2627 2829 2a2b 2c2d 2e2f 3031 3233  $%&'()*+,-./0123
+//   	0x0050:  3435 3637                                4567
+var testPacketPacket0 = []byte{
```
gconnell: please rename to `testPacketRawICMP4` or something like that.

layers/enums.go
```diff
@@ -102,10 +102,10 @@ const (
 	LinkTypeSLIP           LinkType = 8
 	LinkTypePPP            LinkType = 9
 	LinkTypeFDDI           LinkType = 10
+	LinkTypeRaw            LinkType = 12
```
gconnell: I wonder if we should keep Raw at 101 and add Raw2 or something for 12?

layers/raw.go
```diff
@@ -0,0 +1,68 @@
...
+
+package layers
+
+import (
+	"fmt"
+	"github.com/google/gopacket"
+)
+
+// RawIP (DLT_RAW) contains no header and we start with the IP header
+type Raw struct {
```
gconnell: This is great code, but I'm not 100% sure its specific purpose... why would this be useful instead of just having a packet with either `IPv4` or `IPv6` as the first layer?

layers/enums.go
```diff
@@ -93,7 +93,7 @@ const (
 type LinkType uint8
 
 const (
-	// According to pcap-linktype(7) and http://www.tcpdump.org/linktypes.html
+	// According to pcap-linktype(7) and http://www.tcpdump.org/linktypes.html with fixes from pcap/bpf.h
```
gconnell: Sadly, it appears that raw can be 12 or 14 (https://opensource.apple.com/source/libpcap/libpcap-18/libpcap/pcap-bpf.h) or 101 https://github.com/the-tcpdump-group/libpcap/blob/1a6b088a88886eac782008f37a7219a32b86da45/pcap-common.c#L159.  Note that 101 is canonical, since it's how libpcap says "this could be one of a few values".  I wonder if we should do:

```go
  LinkTypeRaw 101
  LinkTypeRawBSD 14
  LinkTypeRawLinux 12
```

Thoughts?  Without 101 pointing at raw, pcap files from raw interfaces will no longer work.

layers/raw.go
zecke: I looked at the loop.go and decodeIPv4or6 and as I don't know who wrote it in the end copied the above copyright. If you think that Andreas's copyright don't apply then maybe Googles doesn't and I should claim it?

layers/raw.go
zecke: It would behave similar to how wireshark is showing the stacking for raw frames, maybe if the encapsulation layer gains a way to write pcap files having a separate interface for the encapsulating layer helps?
Other than that we don't really need to add this layer right now.

layers/raw_test.go
zecke: Can you please elaborate? Which part do you think is copyrighted by Google?

layers/raw_test.go
zecke: will do.

layers/enums.go
zecke: It is worth. On BSD it is 12 but on OpenBSD it is 14. On OpenBSD 12 is used for something else.

layers/enums.go
zecke: I need to see/understand how things will work out if there is not a 1:1 mapping from LinkType* to LayerType. Do you know how it works?

layers/raw_test.go
gconnell: Looking through the CLA, you're absolutely right.  The CLA allows us as Google to use the copyright code, but doesn't transfer the copyright to us as part of the contribution.

Would you mind, for this file, using the following language instead, and adding your name to AUTHORS?:

```
// Copyright 2014 The GoPacket Authors
```

I should probably standardize that on all the other files as well.

layers/enums.go
gconnell: I believe it should be fine... LinkType implements a Decoder, but there's nothing against different link types decoding into the same layer type.

layers/raw.go
gconnell: As mentioned elsewhere, let's do "The GoPacket Authors" and make sure you, Andreas, and I at least are in the AUTHORS file.

lhausermann:
@gconnell @zecke What's the status on this ?

zecke:
I think I had addressed all review comments.

gadkrumholz:
I'd like to add that until I merged this PR locally I couldn't get the https://github.com/google/gopacket/blob/master/examples/arpscan/arpscan.go example to receive any ARP packets in Linux (ubuntu 16.04 x64 with libpcap 1.7.4-2) ... I didn't have to change any code in the arpscan example.

Is this going to get merged soon?

Thanks.

notti:
Sorry didn't see this pull request, when I added 0d76b763fa06849d9c0eca7f98644c7dd4d19c09
Closing, since it is already fixed by this commit.

Issue #249, opened 2016-11-07
Why not add Packet Body Length field to EAPOL frame?


artificerpi:
I am learning IEEE 802.1X protocol with gopacket. Currently, there are only three fields in [EAPOL](https://github.com/google/gopacket/blob/master/layers/eapol.go) interface.
I know that when no further information is needed and the packet body length is set to 0 (and the body is omitted). However, if there is a packet body, such as an EAP message, its length and data are added on as appropriate.
* So why there is no Length field in `EAPOL` type?  since sometimes it decodes EAPOL packet with errors  
### example:
* packet (Get from wireshark)
```
----
# Packet 0 from /tmp/wireshark_eth0_20161108163321_MDBpM6.pcapng
- 154
- 6.941372676
- Hangzhou_91:eb:ec
- Nearest
- EAP
- 64
- Request, Identity

0000   01 80 c2 00 00 03 50 da 00 91 eb ec 88 8e 01 00  ......P.........
0010   00 05 01 01 00 05 01 00 00 00 00 00 00 00 00 00  ................
0020   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00  ................
0030   00 00 00 00 00 00 00 00 00 00 00 00 29 70 35 65  ............)p5e
```

* Code:
```golang
import (
	"fmt"
	"log"
	"time"

	"github.com/google/gopacket"
	"github.com/google/gopacket/layers"
	"github.com/google/gopacket/pcap"
)

var (
	device      string = "eth0"
	snapshotLen int32  = 1024
	promiscuous bool   = false
	err         error
	timeout     time.Duration = -1 * time.Second
	handle      *pcap.Handle
)

func main() {
	// Open device
	handle, err = pcap.OpenLive(device, snapshotLen, promiscuous, timeout)
	var filter string = "ether proto 0x888e || udp port 61440"
	err = handle.SetBPFFilter(filter)
	if err != nil {
		log.Fatal(err)
	}
	defer handle.Close()

	packetSource := gopacket.NewPacketSource(handle, handle.LinkType())
	for packet := range packetSource.Packets() {
		printPacketInfo(packet)
	}
}

func printPacketInfo(packet gopacket.Packet) {
	eapolP := gopacket.NewPacket(packet.Data(), layers.LayerTypeEthernet, gopacket.Default)
	fmt.Println(eapolP)
}
```

* Running log
```
PACKET: 64 bytes
[...]

PR #325, opened 2017-06-27
Fix eap packet parsing

christophert:
The current version of the EAPOL layer does not parse EAPOL packets with length properly. It ignores the Length parameter and attempts to parse the data starting at the Length parameter. This change found in issue #249 fixes the issue and parses EAPOL/EAP packets properly.

### DIFF
--- a/layers/eapol.go
+++ b/layers/eapol.go
@@ -7,6 +7,7 @@
 package layers
 
 import (
+	"encoding/binary"
 	"github.com/google/gopacket"
 )
 
--- a/layers/eapol.go
+++ b/layers/eapol.go
@@ -16,7 +16,7 @@ type EAPOL struct {
 	BaseLayer
 	Version uint8
 	Type    EAPOLType
-    Length  uint16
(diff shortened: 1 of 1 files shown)


christophert:
I signed it!

layers/eapol.go
```diff
@@ -24,7 +26,18 @@ func (e *EAPOL) LayerType() gopacket.LayerType { return LayerTypeEAPOL }
 func (e *EAPOL) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 	e.Version = data[0]
 	e.Type = EAPOLType(data[1])
-	e.BaseLayer = BaseLayer{data[:2], data[2:]}
+	e.Length = binary.BigEndian.Uint16(data[2:4])
+	e.BaseLayer = BaseLayer{data[:4], data[4:]}
```
gconnell: should this be:
```go
e.BaseLayer = BaseLayer{data[:4], data[4:e.Length+4]}
```

PR #327, opened 2017-07-07
[Proposal] Implementation of SIP layer

bodji:
Hi,

I would like to propose an implementation of the SIP layer in gopacket.

It's my first contribution to this project. 
Feel free to make remarks about the code (format, implementation, methods, ...).

Thanks
Mathieu


bodji:
I signed it!

layers/sip.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// NewSIP instantiates a new empty SIP object
+func NewSIP() *SIP {
+	s := new(SIP)
+	s.Headers = make(map[string][]string)
+	return s
+}
+
+// LayerType returns gopacket.LayerTypeDNS.
```
lhausermann: Mmm, this comment must be updated

layers/sip.go
```diff
@@ -0,0 +1,360 @@
+// Copyright 2017 Google, Inc. All rights reserved.
```
lhausermann: Did you gofmt this file ? 
All other code are passed to gofmt before being merged to master.

layers/sip.go
```diff
@@ -0,0 +1,360 @@
...
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
```
lhausermann: Imports should be first the golang packages, a blank line and the gopacket package

layers/sip.go
```diff
@@ -0,0 +1,360 @@
...
+	var line []byte
+	var err error
+
+	// Clean leading new line
+	data = bytes.Trim(data, "\n")
+
+	// Iterate on all lines of the SIP Headers
+	// and stop when we reach the SDP (aka when the new line
+	// is at index 0 of the remaining packet)
+	buffer := bytes.NewBuffer(data)
```
lhausermann: Well, I am not sure is the bytes package could be used here. I would have say yes, as it does not do allocation or copy. SIP is a text based protocols so using the golang bytes API is natural (ie as you do to find end of lines, etc..).
Other layers are parsing binary protocols and use the binary package.

@gconnell what's your point of view about this ?

layers/sip.go
```diff
@@ -0,0 +1,360 @@
...
+			}
+		}
+
+		// Trim the new line delimiters
+		line = bytes.Trim(line, "\r\n")
+
+		// Empty line, we hit Body
+		// Putting packet remain in Paypload
+		if len(line) == 0 {
+			s.BaseLayer.Payload = buffer.Bytes()
```
lhausermann: No, the Payload is normally what is encapsulted and will be passed to another layer. Like TCP/UDP/GRE/SCTP etc.. Here Payload for SIP is NULL as there is no extra layer to parse it.

layers/sip.go
bodji: Yes.

layers/sip.go
bodji: I do not understand. I based the format on dns layer

layers/sip.go
bodji: Yes, doing

layers/sip.go
bodji: SIP often contains SDP in the Payload, so I put the rest (boy) in the payload no ?

bodji:
Hi everyone !

I just made some improvments on my implementation as you suggested.

Here are the changes : 
- A little bit of refactor
- Handling multi-line headers (as defined in RFC 3261 - 7.3.1)
- Add some useful getters for common SIP Headers
- Add tests 


Can someone review my code ? :)
Thanks !

lhausermann:
Sorry for the delay.

@bodji 1) I am sorry but there is remaining conflicts. Can you fix these ? 2) regarding the unit test, we would prefer to have full packet inside the test.. it will test also the layers sequence which is better. can you provide those?


@gconnell  I suggest to merge this branch after

bodji:
Thanks @lhausermann, I just did some fixes : 
- Rebasing branch
- Resolved some lint errors
- Resolved some variable declaration lint errors

The Travis build is now successfull.

Tonight, I will add a full packet in the test file.

layers/sip.go
lhausermann: all right

lhausermann:
@bodji did you add the full packet test ?

bodji:
Hi @lhausermann, I just pushed the new test file with complete 
packets of a request/response SIP REGISTER session :)

Is it okay for you ?

lhausermann:
But now you made it, I think we can merge.

Ok @gconnell ?

PR #332, opened 2017-08-07
[proposal] ospf layer

florianl:
Hi,
I've implemented the ospf layer for gopacket and I'm looking forward to answer your comments.

Kind regards,
 Florian


layers/ospf.go
```diff
@@ -0,0 +1,120 @@
...
+package layers
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/gopacket"
+)
+
+type OSPFType uint8
```
gconnell: Please add comments for all new types, and for exported functions (even if they are simply "DecodeFromBytes implements gopacket.DecodingLayer" or the like).

layers/ospf.go
```diff
@@ -0,0 +1,120 @@
...
+// OSPFv3 extend the OSPF head with version 3 specific fields
+type OSPFv3 struct {
+	Instance uint8
+	Reserved uint8
+}
+
+// OSPF is a basic OSPF packet header with common fields of Version 2 and Version 3.
+type OSPF struct {
+	BaseLayer
+	OSPFv2
```
gconnell: Interesting use of embedded types here.  I'm not sure if I prefer your approach:

```go
type OSPF struct {
  OSPFv2
  OSPFv3
  ...
}
```

or this approach:

```go
type OSPF {
  ...  common fields ...
}

type OSPFv2 {
  OSPF
  ... v2 specific fields
}

type OSPFv3 {
  OSPF
  ... v3 specific fields ...
}
```

I'd be interested to hear your thoughts on each.  A question this brings up is whether an OSPFv2 packet should be a different layer type than an OSPFv3 one.

layers/ospf.go
```diff
@@ -0,0 +1,120 @@
...
+	return gopacket.LayerTypePayload
+}
+
+// CanDecode returns the set of layer types that this DecodingLayer can decode.
+func (ospf *OSPF) CanDecode() gopacket.LayerClass {
+	return LayerTypeOSPF
+}
+
+func decodeOSPF(data []byte, p gopacket.PacketBuilder) error {
+
```
gconnell: spurious empty line ;)

florianl:
I signed it!

layers/ospf.go
florianl: My approach might cost a few more bytes in memory, but keeps it as simple as possible for the user. You just have to handle with one struct.

florianl:
I've changed the code according the advice of @gconnell.

florianl:
Hi, 
I've extended the PR, so not only the header of the OSPF packet is handled but also the whole packet as it is described in RFC 5340. So I'm looking forward your comments.

florianl:
If there are problems or questions, please let me know @gconnell .

florianl:
I've resolved conflicts because of recent changes.

gconnell:
Could you merge layers/layertypes and resubmit?

gconnell:
Thanks for your great work on this!

PR #336, opened 2017-08-17
adding support for discarding on captured data

jzaeske:
I'm using gopacket to analyse pcap files, containing some fragmented ip packets.
Due to using `time.Now()` in `insert` was unable to use a sliding window based on the actual capture time.

This pull request provies an additional function `DefragIPv4WithTimestamp` which can be used to specify the acutal capture time. As before, DefragIPv4 uses the default `time.Now()`

### DIFF
--- a/ip4defrag/defrag.go
+++ b/ip4defrag/defrag.go
@@ -73,6 +73,16 @@ const (
 //}
 //
 func (d *IPv4Defragmenter) DefragIPv4(in *layers.IPv4) (*layers.IPv4, error) {
+	return d.DefragIPv4WithTimestamp(in, time.Now())
+}
+
+// DefragIPv4WithTimestamp provides functionality of DefragIPv4 with
+// an additional timestamp parameter which is used for discarding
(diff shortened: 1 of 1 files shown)


jzaeske:
I signed it!

gconnell:
Thanks!

lhausermann:
@jzaeske : as the author of this original code and a daily user of this, I am curious to know in which case the older imp does not work ?

jzaeske:
@lhausermann maybe my usecase was a bit special. Your original code does a well job in nearly all cases. Thanks for that.

I did defragmentation on some pcap files using a sliding window, also interested in the amount of uncompleted fragments. With the old code a sliding window would have been possible by discarding fragments older then `time.Now() - windowWidth`. That works, but the result depends on the machine the code is running on. Also other processes which are using cpu time can affect the result. 
To have reproducible results, independend on the runtime I wanted to use the actual measurement time of the framgment. 
The branch name might have been misleading. Sorry for that. It's more an improvement than a fix.

PR #337, opened 2017-08-17
fixing memory leak in reassembly

jzaeske:
Using reassmbly to reassemble tcp connections in large pcap files I noticed an increasing usage of memory. I'm not sure why, but it seems that `(p *StreamPool) remove()` is executed multiple times for the same connection, when `ReassemblyComplete()` returns `true`.

In this case deleting the connection from `p.conns` does nothing, but the connection is added multiple times to `p.free`.

This pull request provides functionality to avoid this duplicates in `p.free` by checking if the given connection is available in `p.conns`.

### DIFF
--- a/reassembly/memory.go
+++ b/reassembly/memory.go
@@ -170,8 +170,10 @@ func (p *StreamPool) Dump() {
 
 func (p *StreamPool) remove(conn *connection) {
 	p.mu.Lock()
-	delete(p.conns, conn.key)
-	p.free = append(p.free, conn)
+	if _, ok := p.conns[conn.key]; ok {
+		delete(p.conns, conn.key)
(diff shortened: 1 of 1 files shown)


jzaeske:
I signed it!

rfrancoise:
Note: the fix makes sense, but it would be interesting to know why remove() seems to be called multiple times for the same connection.

jzaeske:
@rfrancoise I agree with you. Unfortunately I did not completely understand all the internal behaviour to name the source of that issue. This fix does at least reduce the impact of it.

rfrancoise:
@jzaeske can you perhaps share more information about your use-case? Are you using `reassemblydump`, or your own code that uses the `reassembly` package?

PR #342, opened 2017-09-06
Fix: sflow Extended Router Flow Record skipping IP type field

jthurman42:
The sflow decodeExtendedRouterFlowRecord was missing address type detection, and decoding was off by 4 bytes if ExtendedRouterFlowRecord was not the last record.

The test packets have been updated to reflect the reference below (validated through pcap), and an additional IPv6 test was added to verify the detection was working as expected.

Packet format reference: http://sflow.org/developers/diagrams/sFlowV5FlowData.pdf


jthurman42:
I signed it!

lhausermann:
Hi! Thanks for contributing .. 

Being not a sflow expert is @safchain or @actaeon would be able to review that code ?

jthurman42:
@safchain I'll take a look at SFlowExtendedGatewayFlowRecord, and open another PR for it shortly.  Thanks!

lhausermann:
OK, so as everybody agrees here, I am merging this one.

PR #344, opened 2017-09-09
Improve own decoder example

florianl:

Improve example in the documentation so that it can be copied and used directly.


gconnell:
Thanks!

Issue #345, opened 2017-09-12
pcap.BlockForever never yields packets on Ubuntu 16.04


shabbyrobe:
Even just using the simple examples in the `examples/` directory (as well as with my own code), I can not get gopacket to yield any packets via the `PacketSource`, it just blocks forever. The examples herein use `httpassembly`, but `bidirectional` exhibits the same behaviour.

Using both my workstation and a fresh installation of Ubuntu Server 16.04 in a virtual machine, I was able to reproduce consistently.

```
# assuming go is set up and configured:
sudo apt install apache2 build-essential libpcap-dev
go get -u github.com/google/gopacket
sudo systemctl start apache2
go install github.com/google/gopacket/examples/httpassembly
sudo "$GOPATH/bin/httpassembly" -i lo

# in another terminal:
curl -sI localhost
```

If I modify the `httpassembly` example to change `pcap.BlockForever` to an actual timeout (5 seconds in the below case), I see packets after the timeout expires, but not when they actually occur.

```
$ go install github.com/google/gopacket/examples/httpassembly && sudo "$GOPATH/bin/httpassembly" -i lo                       
2017/09/12 11:34:23 Starting capture on interface "lo"
2017/09/12 11:34:23 reading in packets
2017/09/12 11:34:33 Received request from stream ::1->::1 45846->80 : &{GET / HTTP/1.1 1 1 map[User-Agent:[curl/7.47.0] Accept:[*/*]] {} <nil> 0 [] false localhost map[] map[] <nil> map[]  / <nil> <nil> <nil> <nil>} with 0 bytes in request body
2017/09/12 11:34:33 Received request from stream ::1->::1 45848->80 : &{GET / HTTP/1.1 1 1 map[User-Agent:[curl/7.47.0] Accept:[*/*]] {} <nil> 0 [] false localhost map[] map[] <nil> map[]  / <nil> <nil> <nil> <nil>} with 0 bytes in request body
2017/09/12 11:34:33 Received request from stream ::1->::1 45850->80 : &{GET / HTTP/1.1 1 1 map[User-Agent:[curl/7.47.0] Accept:[*/*]] {} <nil> 0 [] false localhost map[] map[] <nil> map[]  / <nil> <nil> <nil> <nil>} with 0 bytes in request body
```

```
bl@ubuntu:~$ for i in {1..3}; do date && curl -s localhost > /dev/null ; sleep 1; done
Tuesday 12 September  11:34:29 AEST 2017
[...]

PR #346, opened 2017-09-12
Fix pcap's timeout for Unix.

gconnell:
For PCAP, we should:

1) if we have a timeout, return an error when that timeout has been hit
and no packet received.
2) if we have a negative timeout, loop forever waiting for a packet,
checking every -timeout.


lhausermann:
Not sure you are seeking for review, but that seems a far better way to handle this test (ie switch case increase readability)

gconnell:
Thanks for the look, Laurent :)

gconnell:
Fixes #345

PR #348, opened 2017-09-16
More elegant solution for decodeExtendedRouterFlowRecord

jthurman42:
PR #342 fixed a bug in layers/sflow.go `decodeExtendedRouterFlowRecord` that missed decoding the address.  The initial patch was accepted, but was not very elegant.  In that PR, @safchain requested that I also fix the issue in for decoding `SFlowExtendedGatewayFlowRecord`.  Upon examination, the in-code documentation did not show the field, but the actual decode block did the correct thing.  That is also where the more elegant solution below came from (Thanks @actaeon).

This PR:
* Changes the address type detection to follow existing methods (using `SFlowIPType`)
* Updates the documented Extended gateway records structure to match the implementation


actaeon:
LGTM. 👍

PR #352, opened 2017-09-19
add a simple STP decoder

nplanel:


layers/stp.go
```diff
@@ -0,0 +1,26 @@
...
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"github.com/google/gopacket"
+)
+
+// Spanning Tree Protocol is on top of LLC to transport BPDU (bridge protocol data unit) message.
```
gconnell: Godoc comments should start with the name of the thing they're commenting on.  For here:

```go
// STP decodes spanning tree protocol packets to ...
```

layers/llc.go
```diff
@@ -62,6 +62,9 @@ func decodeLLC(data []byte, p gopacket.PacketBuilder) error {
 	if l.DSAP == 0xAA && l.SSAP == 0xAA {
```
gconnell: might be better to make this a switch:

```go
switch {
  case l.DSAP == 0xAA && l.SSAP == 0xAA:
    ...
  case l.DSAP == 0x42 && l.SSAP == 0x42:
    ...
}
```

I expect we'll add things here later on, and a switch seems more useful for doing that.

layers/stp.go
```diff
@@ -0,0 +1,26 @@
...
+type STP struct {
+	BaseLayer
```
gconnell: I'm assuming the idea is to fill this in with actual STP fields at a later date.  Maybe add a TODO here?
```go
// TODO:  parse the STP protocol into actual subfields.
```

layers/stp.go
nplanel: done

layers/stp.go
nplanel: done

layers/llc.go
nplanel: done

gconnell:
Cool, thanks!  Can't wait to see it expanded to decode protocol fields :)

PR #357, opened 2017-09-28
fix nfds error in select statement

tzneal:
From man page for select "nfds should be set to the highest-numbered
file descriptor in any of the three sets, plus 1". This was causing
one of our apps to never capture packets as the fd returned by
pcap_get_selectable_fd was 8 and then never checked.


tzneal:
Any chance you can tag a release after this as well?

gconnell:
Pushed v1.1.14

tzneal:
Thanks!

PR #359, opened 2017-10-01
add unknown file format test

florianl:

Adds a basic `unknown file format` test.

### DIFF
--- a/pcap/pcap_test.go
+++ b/pcap/pcap_test.go
@@ -9,7 +9,9 @@ package pcap
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
+	"os"
 	"testing"
 
 	"github.com/google/gopacket"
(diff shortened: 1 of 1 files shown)


pcap/pcap_test.go
```diff
@@ -29,42 +31,69 @@ func TestPcapNonexistentFile(t *testing.T) {
...
-			t.Fatal(err)
+			if file.err != "" {
+				if err.Error() != file.err {
+					t.Errorf("expected message %q; got %q", file.err, err.Error())
+				}
+			} else {
+				t.Fatal(err)
+			}
 		} else {
 			packetSource := gopacket.NewPacketSource(handle, handle.LinkType())
```
gconnell: I think you also need to check inside the else for file.err:

if handle, err := ...; err != nil {
} else if file.err != "" {
  t.Fatalf("Expected error, got none")
} else {
  ...
}

florianl:
thanks @gconnell for this good point. I've added it.

Issue #347, opened 2017-09-15
Wrong assumptions on SCTP data chunk and design issue..


zecke:
In one SCTP package we can have _multiple_ SCTPData chunks. The current code doesn't support it and will provide all other chunks as payload. I removed the special case of decodeSCTPChunk and then put a Payload into the SCTPData struct. One issue is that the test data is kind of invalid...

The fundamental issue is that the structure of SCTP is like this.

```
   IP
  SCTP
  SCTPDataChunk
     Payload
  SCTPDataChunk
     Payload
```

While p.NextDecoder will just be linear. We would need to have some kind of special "Payload" so we can jump to the next SCTPDataChunk. Alternatively be always eager for SCTP...

gconnell:
Would it make sense for SCTPDataChunk to just have a []byte to its data, then it could return the next SCTPDataChunk immediately after?

zecke:
Yes, that would be the easiest (and how I patched gopacket) locally. But I don't know if "Payload" has significance for other users gopacket? E.g. if code should work transparently for tcp and sctp then someone might have written something like the below code? Should this continue to work?

```
for _, layer := ranges packet.Layers() {
    if payload, ok := packet.(Payload); !ok {
      continue
    }
}
```

PR #360, opened 2017-10-01
Fix to handle multiple chunks

florianl:

This is a fix, to handle multiple chunks as defined in [RFC 4960 - Section 3](https://tools.ietf.org/html/rfc4960#section-3).


florianl:
As @zecke said in https://github.com/google/gopacket/issues/347 

> One issue is that the test data is kind of invalid...

I'll fixing that as well.

zecke:
I think you want to remove the special casing of SCTP Data chunk as well? E.g. chunk.BaseLayer defines what is parsed next?

florianl:
As you pointed out, the test cases are invalid ...
At first, I will correct them.

PR #362, opened 2017-10-04
Fix handling of SCTP type data

florianl:

This patch fixes the design issue mentioned in https://github.com/google/gopacket/issues/347. It allows the nesting of SCTP data chunks. The payload of a data chunk is stored inside a []byte within the SCTPData struct.

### DIFF
--- a/layers/sctp.go
+++ b/layers/sctp.go
@@ -120,19 +120,12 @@ func decodeSCTPChunk(data []byte) (SCTPChunk, error) {
 	actual := roundUpToNearest4(int(length))
 	ct := SCTPChunkType(data[0])
 
-	// For SCTP Data, use a separate layer for the payload
-	delta := 0
-	if ct == SCTPChunkTypeData {
-		delta = int(actual) - int(length)
-		actual = 16
-	}
-
 	return SCTPChunk{
 		Type:         ct,
 		Flags:        data[1],
 		Length:       length,
 		ActualLength: actual,
-		BaseLayer:    BaseLayer{data[:actual], data[actual : len(data)-delta]},
+		BaseLayer:    BaseLayer{data[:actual], data[actual:len(data)]},
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


layers/sctp.go
```diff
@@ -304,19 +298,25 @@ func decodeSCTPData(data []byte, p gopacket.PacketBuilder) error {
 		StreamSequence:  binary.BigEndian.Uint16(data[10:12]),
 		PayloadProtocol: SCTPPayloadProtocol(binary.BigEndian.Uint32(data[12:16])),
 	}
+
+	if sc.PayloadProtocol == 0 {
+		sc.Payload = make([]byte, sc.SCTPChunk.Length-16)
```
gconnell: don't alloc/copy the data here, just set `sc.Payload = data[16:sc.blahblah]`.

if https://godoc.org/github.com/google/gopacket#DecodeOptions `NoCopy` is false (the default), gopacket will have already have copied this data into a new buffer.  If it's true, then the user has assumed responsibility for the passed-in buffer and will make sure it stays live.

layers/sctp.go
```diff
@@ -304,19 +298,25 @@ func decodeSCTPData(data []byte, p gopacket.PacketBuilder) error {
...
+		copy(sc.Payload, data[16:sc.SCTPChunk.Length+1])
+		// Length is the length in bytes of the data, INCLUDING the 16-byte header.
+		p.AddLayer(sc)
```
gconnell: Make sure you call p.SetApplicationLayer(sc) here, to set the ApplicationLayer() pointer to the first application layer seen in the packet.

florianl: I think,this isn't the right place for `p.SetApplicationLayer(sc)`. The payload that is contained inside the SCTPData chunk isn't decoded. And since it can be anything, there will be no decoder for it. The layer for the protocol is set at a different place https://github.com/google/gopacket/blob/master/layers/sctp.go#L34

PR #365, opened 2017-10-08
Handling nil case of RawSockaddr

aslanbekirov:
Nil case of RawSockaddr should be checked otherwise it gets `nil preference

PR #366, opened 2017-10-08
Add check if value exists

florianl:

This patch adds a check if the received pointer is not `nil`.

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -775,6 +775,10 @@ func findalladdresses(addresses *_Ctype_struct_pcap_addr) (retval []InterfaceAdd
 }
 
 func sockaddrToIP(rsa *syscall.RawSockaddr) (IP []byte, err error) {
+	if unsafe.Pointer(rsa) == nil {
+		err = errors.New("Value not set")
+		return
+	}
 	switch rsa.Family {
 	case syscall.AF_INET:
 		pp := (*syscall.RawSockaddrInet4)(unsafe.Pointer(rsa))


pcap/pcap.go
```diff
@@ -775,6 +775,10 @@ func findalladdresses(addresses *_Ctype_struct_pcap_addr) (retval []InterfaceAdd
 }
 
 func sockaddrToIP(rsa *syscall.RawSockaddr) (IP []byte, err error) {
+	if unsafe.Pointer(rsa) == nil {
```
gconnell: any reason why unsafe.Pointer is necessary here?  I'd assume that a simpler `if rsa == nil` would work just as well?

aslanbekirov: did we get the answer for question above?

gconnell: Florian responded via email:

```
Hi!

I needed to cast rsa otherwise it paniced and crashed for me.
As I don't have access to github during office hours, I'm writting this way.

Cheers,
   Florian
```

aslanbekirov: I a sorry why we should cast it? I did it without casting in #365  , worked for me and passed the tests.

gconnell:
Florian, do you have a stack trace you could provide?

florianl:
Sorry for my delayed answer.

The lines of code I used to reproduce the problem:
`package main

import (
	"fmt"
	"github.com/google/gopacket/pcap"
)

func main() {
	devices, err := pcap.FindAllDevs()
	if err != nil {
		fmt.Errorf("%v\n", err)
		return
	}

	for _, device := range devices {
		fmt.Println("\nName: ", device.Name)
		fmt.Printf("\nFlags: %x\n", device.Flags)
		for _, address := range device.Addresses {
			fmt.Println("- IP address: ", address.IP)
			fmt.Println("- Subnet mask: ", address.Netmask)
			fmt.Println("- Broadaddr mask: ", address.Broadaddr)
			fmt.Println("- P2P mask: ", address.P2P)
		}
	}
}`

And the stack trace, if I run it without the casting:
`$ ./devices 
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x4ff416]

goroutine 1 [running]:
github.com/google/gopacket/pcap.sockaddrToIP(0x0, 0xc4210d46d8, 0x4, 0x4, 0x0, 0x0)
	/home/flo/go/src/github.com/google/gopacket/pcap/pcap.go:778 +0x26
github.com/google/gopacket/pcap.findalladdresses(0x134e770, 0x0, 0x0, 0xc420012700)
	/home/flo/go/src/github.com/google/gopacket/pcap/pcap.go:769 +0x1f9
github.com/google/gopacket/pcap.FindAllDevs(0xc420012700, 0xd, 0xd, 0x0, 0x0)
	/home/flo/go/src/github.com/google/gopacket/pcap/pcap.go:734 +0x231
main.main()
	/home/flo/devices.go:9 +0x55
`

aslanbekirov:
can you please also past code of `github.com/google/gopacket/pcap.sockaddrToIP` that used for this example?

florianl:
The trace was generated with ```if rsa == nil {```.

aslanbekirov:
I run exact same code with only change `if rsa == nil`  in `github.com/google/gopacket/pcap.sockaddrToIP` and it worked for me. that is interesting. wonder why.

Issue #376, opened 2017-10-21
Enum data takes 18MB of memory


dlespiau:
The EnumMetadata objects take about 18MB of memory as about 133k strings and functions are created in https://github.com/google/gopacket/blob/master/layers/enums.go#L388.

It'd be nice to turn some of this into functions instead of heap data.

![screenshot from 2017-10-21 23-51-47](https://user-images.githubusercontent.com/7986/31856538-486512f8-b6bb-11e7-8fe3-31d413e4aa41.png)

gconnell:
It's been a request in the past, but unfortunately it seems impossible to handle this without seriously breaking backwards compatibility.  If you have any thoughts on the best way to transition this, though, I'd be happy to hear them.

gconnell:
Actually, it seems eminently possible to make this data at least DATA-section instead of heap, by making the strings all constant.  Could save on startup time, too... let me take a quick look.

dlespiau:
Maybe the EnumMetadata entries, in say EthernetTypeMetadata to make things concrete, for ether types without a handler could remain uninitialised. Then, `EthernetType.Decode()` could test whether this EthernetTypeMetadata[a].DecodeWith is nil and provide the default behaviour that is currently assigned to every entry in `init()`.

Of course, strictly speaking, this break compatibility as `EthernetTypeMetadata[i]` can now contain uninitialized data.

A good thing would be to know how awful it would be if EthernetTypeMetadata were to be hidden in future versions, provided there's a new way for people to override the decoders.

gconnell:
See #377 to see if that helps things out.

dlespiau:
With #377 merged we can close this, thanks for the extremely quick PR!

PR #377, opened 2017-10-23
Utilize some trickiness to reduce init memory usage in enums.

gconnell:
Using some prebuilt arrays that don't require heap creation and some lazy string creation for errors, avoid doing a ton of fmt.Sprintf in enums.go's init function, which should speed up initiation and reduce heap mem usage.

### DIFF
--- a/layers/enums.go
+++ b/layers/enums.go
@@ -272,107 +272,6 @@ const (
 	Dot11TypeDataQOSCFAckPollNoData Dot11Type = 0x3e
 )
 
-var (
-	// Each of the following arrays contains mappings of how to handle enum
-	// values for various enum types in gopacket/layers.
-	//
-	// So, EthernetTypeMetadata[2] contains information on how to handle EthernetType
-	// 2, including which name to give it and which decoder to use to decode
-	// packet data of that type.  These arrays are filled by default with all of the
-	// protocols gopacket/layers knows how to handle, but users of the library can
-	// add new decoders or override existing ones.  For example, if you write a better
-	// TCP decoder, you can override IPProtocolMetadata[IPProtocolTCP].DecodeWith
-	// with your new decoder, and all gopacket/layers decoding will use your new
-	// decoder whenever they encounter that IPProtocol.
-	EthernetTypeMetadata     [65536]EnumMetadata
-	IPProtocolMetadata       [256]EnumMetadata
-	SCTPChunkTypeMetadata    [256]EnumMetadata
-	PPPTypeMetadata          [65536]EnumMetadata
-	PPPoECodeMetadata        [256]EnumMetadata
-	LinkTypeMetadata         [256]EnumMetadata
-	FDDIFrameControlMetadata [256]EnumMetadata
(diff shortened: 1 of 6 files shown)


gconnell:
Attempting to fix #376

dlespiau:
So, I've done some testing:

First, the static impact:

screenshot from 2017-10-24 12-21-57

As you can see the big impact there is the `.noptrbss` section that gains 1MB. From the section name, I'm assuming that's a section the garbage collector doesn't look at, which is good news. That prompted me to wonder if the allocation of all those objects isn't putting some strain on the garbage collection. So far I haven't found any proof of that.

Secondly the dynamic impact. There the picture looks quite good. The big heap data consumers in the profile of bug #376 don't show up anymore. The *big* win appears in the RSS memory used:

before:
```
USER       PID %CPU %MEM    VSZ    RSS TTY      STAT START   TIME COMMAND
root     12077  6.2  0.8 738788 145076 pts/8   Sl+  12:23   0:07              |   |       \_ ./scope.gopacket.master -mode=probe -probe.http.listen=:4039
```
after:
```
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root     13446  6.0  0.6 638464 111608 pts/8    Sl+  12:26   0:02              |   |       \_ ./scope.gopacket.pr-377 -mode=probe -probe.http.listen=:4039
```

So, if those numbers are correct, we're saving a lot more than 18MB RSS. I'm guessing this has to do with how the go runtime manages memory, but unfortunately not in the know of the fine details there.

As far as I'm concerned, +1MB to the binary is a good trade-off to save that much RSS space, especially as I'm hopefully not all the pages backing that `.noptrbss` section need to be in memory if untouched during the program life time.

layers/enums.go
```diff
@@ -558,7 +426,7 @@ func init() {
 	Dot11TypeMetadata[Dot11TypeDataQOSCFPollNoData] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeDot11DataQOSCFPollNoData), Name: "DataQOSCFPollNoData", LayerType: LayerTypeDot11DataQOSCFPollNoData}
 	Dot11TypeMetadata[Dot11TypeDataQOSCFAckPollNoData] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeDot11DataQOSCFAckPollNoData), Name: "DataQOSCFAckPollNoData", LayerType: LayerTypeDot11DataQOSCFAckPollNoData}
 
-	USBTypeMetadata[USBTransportTypeInterrupt] = EnumMetadata{DecodeWith: gopacket.DecodeFunc(decodeUSBInterrupt), Name: "Interrupt", LayerType: LayerTypeUSBInterrupt}
```
dlespiau: This seems to be an API change?

gconnell:
Things seem good, so I'm going ahead with this.  Thanks for the testing!

dlespiau:
Oh btw, it's +1MB to a **BSS** section, so it doesn't use +1MB more disk space, just ~50kb. \o/

PR #380, opened 2017-10-27
802.11-related fixes

sknaumov:
I had to make these changes to be able to use gopacket to strip data payload from packets while retaining all 802.11 headers.

It looks like originally gopacket was suitable for my task until this commit: https://github.com/google/gopacket/commit/d23ebfbfc434b2f371699e7d59a30d096de2cde1

It would be good to know what goal that commit was trying to solve. Probably it is OK to use Dot11WEP instead of Dot11Data, but so far I do not understand how to propagate context from Dot11 to Dot11Data via Dot11DataQoSData.

I tested my changes on sniffed Wi-Fi packets - both encrypted (not only data, but and also management - see 802.11w) or not. Everything is OK if to stop iterating over layers at Dot11Data or at Radiotap if bad FCS encountered.

### DIFF
--- a/layers/dot11.go
+++ b/layers/dot11.go
@@ -625,16 +625,15 @@ type Dot11DataQOS struct {
 }
 
 func (m *Dot11DataQOS) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
-	if len(data) < 4 {
+	if len(data) < 2 {
 		df.SetTruncated()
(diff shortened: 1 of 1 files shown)


rstrand:
Pull request #436 which has been merged fixed these issues

notti:
Closing, because the issue was resolved.

PR #384, opened 2017-11-04
Activate(): handle PCAP_WARNING, provide proper error message

rschmied:
When activating an InactiveHandle on Linux with a buffer size that is too small, the returned error is '1' but this case is not handled in the code (probably there are more situations where this can occur, the above happened to me). The man page says

> **PCAP_WARNING**
              Another warning condition occurred; pcap_geterr() or pcap_perror() may be called with p as an  argument  to  fetch  or  display  a  message describing the warning condition.

Didn't have this issue on my Mac and had to modify the gopacket code to figure out what was actually going on behind the scenes (in my case `can't mmap rx ring: Invalid argument` which was caused by the inappropriate buffer size). Thought it would be nice if gopacket could tell this apart and report it properly out-of-the-box.

There's some code replication for the `(*InactiveHandle) Error()` which is identical to `(*Handle) Error()` and also the variable which holds the content... Since I'm relatively new to Golang there probably is a better way... works for me, however.

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -341,6 +341,7 @@ const (
 	aeNoSuchDevice = C.PCAP_ERROR_NO_SUCH_DEVICE
 	aeDenied       = C.PCAP_ERROR_PERM_DENIED
 	aeNotUp        = C.PCAP_ERROR_IFACE_NOT_UP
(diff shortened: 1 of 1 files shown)


rschmied:
I signed it!

gconnell:
Thanks!

PR #387, opened 2017-11-27
Remove costly allocation from writePacketHeader by moving buffer into struct

nu11ptr:
My production code uses pcapgo.WritePacket to write header and packet data to a byte slice in the real time capture tight loop. While profiling this code, I discovered pcapgo.writePacketHeader's byte array seems to get allocated on the heap. I found by moving this into the Writer struct this allocation is not made saving approx. 30-40ns per call on my machine. While this won't matter much for post-capture processing, it does make a large impact during the capture itself. Moving this will make these write routines safe to call only from a single goroutine, but I can't see what use case could be had by calling from multiple anyway.

Before:
---------
    scott@saruman:~/go/src/github.com/google/gopacket/pcapgo$ go test -v ./... -run=^$ -bench=.
    goos: linux
    goarch: amd64
    pkg: github.com/google/gopacket/pcapgo
    BenchmarkWritePacket-8   	20000000	       116 ns/op
    PASS
    ok  	github.com/google/gopacket/pcapgo	2.510s

After:
-------
    scott@saruman:~/go/src/github.com/google/gopacket/pcapgo$ go test -v ./... -run=^$ -bench=.
    goos: linux
    goarch: amd64
    pkg: github.com/google/gopacket/pcapgo
    BenchmarkWritePacket-8   	20000000	        73.6 ns/op
    PASS
    ok  	github.com/google/gopacket/pcapgo	2.726s

### DIFF
--- a/pcapgo/write.go
+++ b/pcapgo/write.go
@@ -26,6 +26,9 @@ import (
 // timestamp resolution and little-endian encoding.
 type Writer struct {
 	w io.Writer
+
+	// Moving this into the struct seems to save an allocation for each call to writePacketHeader
+	buf [16]byte
 }
 
 const magicMicroseconds = 0xA1B2C3D4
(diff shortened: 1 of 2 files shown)


gconnell:
Great change, thanks!

PR #388, opened 2017-11-27
Fix #385 by truncating DHCP option list.

### LINKED ISSUE #385: repeatedly calling gopacket.NewDecodingLayerParser.DecodeLayers to decode DHCP packets causes DHCPOptions to be appended instead of reset
something like the following code will cause dhcp options to get appended each time a DHCP packet is processed

var (
               dhcp layers.DHCPv4
)
parser := gopacket.NewDecodingLayerParser(layers.LayerTypeEthernet, &dhcp)
decoded := []gopacket.LayerType{}
for packet := range packetSource.Packets() {
        _ = parser.DecodeLayers(packet.Data(), &decoded)
	for _, layerType := range decoded {
		switch layerType {
		case layers.LayerTypeDHCPv4:
                      // second time here dhcp.Options will contain data from both packets
                      fmt.Println(dhcp.Options)
                 }
       }
}

gconnell:


vtolstov:
Thanks!

Issue #389, opened 2017-11-28
tcpassembly can incorrectly indicate data loss


mspielberg:
When a page is returned to the page cache, it may have its Skip field set.

The Skip field is not reset when the page is retrieved from the cache.
https://github.com/google/gopacket/blob/a1c3065f20e5974a7be1a89a6daeecf9e81b9adc/tcpassembly/assembly.go#L137

The Skip field is only written in addNextFromConn() if a gap in sequence numbers is detected.
https://github.com/google/gopacket/blob/a1c3065f20e5974a7be1a89a6daeecf9e81b9adc/tcpassembly/assembly.go#L765

The upshot is that a Stream may receive a Reassembly with a non-zero Skip indicating data loss, where no such loss occurred.

The simplest fix would be to add an `else` to addNextFromConn() to reset Skip to 0 for the normal happy-path case. I have a test case in my application which I can minimize if desired.

gconnell:
Please try out https://github.com/google/gopacket/pull/390 at your convenience.  If you wouldn't mind sending a pull request with the test case, I'd greatly appreciate it.

mspielberg:
Tested #390 and it is working. I have the test case minimized and will submit a PR tomorrow.

gconnell:
#390 rebased/merged, thanks for the report!

PR #392, opened 2017-11-29
Add test for #389 Skip having non-zero value on reused pages

mspielberg:


gconnell:
Would you be able to accept the CLA to allow this to be pulled in?

notti:
@mspielberg ping

PR #401, opened 2018-01-02
Fixing SFlowIpv6Record SFlowIpv4Record does not resolve the problem o…

lflxp:
When I parsed the sFlow V5 with gopacket, I found that the length of the SFlowIpv4Record interception was wrong, and it was found that the lack of SFlowBaseFlowRecord resulted in the lack of SFlowBaseFlowRecord

### DIFF
--- a/layers/sflow.go
+++ b/layers/sflow.go
@@ -1613,6 +1613,7 @@ func decodeExtendedUserFlow(data *[]byte) (SFlowExtendedUserFlow, error) {
 //  |                      TOS                      |
 //  +--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+--+
 type SFlowIpv4Record struct {
+	SFlowBaseFlowRecord
 	// The length of the IP packet excluding ower layer encapsulations
 	Length uint32
 	// IP Protocol type (for example, TCP = 6, UDP = 17)
@@ -1633,6 +1634,11 @@ type SFlowIpv4Record struct {
 
 func decodeSFlowIpv4Record(data *[]byte) (SFlowIpv4Record, error) {
 	si := SFlowIpv4Record{}
+	var ir SFlowFlowDataFormat
+
+	*data, ir = (*data)[4:], SFlowFlowDataFormat(binary.BigEndian.Uint32((*data)[:4]))
+	si.EnterpriseID, si.Format = ir.decode()
+	*data, si.FlowDataLength = (*data)[4:], binary.BigEndian.Uint32((*data)[:4])
 
 	*data, si.Length = (*data)[4:], binary.BigEndian.Uint32((*data)[:4])
 	*data, si.Protocol = (*data)[4:], binary.BigEndian.Uint32((*data)[:4])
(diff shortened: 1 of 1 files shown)


lflxp:
I signed it!

ezak:
i signed it!

gconnell:
I believe the CLA bot is specifically looking for a CLA from "李学坪"

Also, currently Travis is complaining about this test failure:

```
$ go test github.com/google/gopacket/layers
--- FAIL: TestDecodeExtendedIpv4TunnelIngressFlow (0.00s)
	sflow_test.go:1097: Failed to decode packet: runtime error: slice bounds out of range
	base_test.go:25: Checking packet layers, want [SFlow]
	base_test.go:27:   Got layer DecodeFailure, 108 bytes, payload of 0 bytes
	base_test.go:30: PACKET: 108 bytes
		- Layer 1 (108 bytes) = DecodeFailure	Packet decoding error: runtime error: slice bounds out of range
		
	base_test.go:38:   Layer 0 mismatch: got DecodeFailure want SFlow
panic: interface conversion: gopacket.ApplicationLayer is nil, not *layers.SFlowDatagram [recovered]
	panic: interface conversion: gopacket.ApplicationLayer is nil, not *layers.SFlowDatagram
```

layers/sflow.go
```diff
@@ -1633,6 +1636,11 @@ type SFlowIpv4Record struct {
 
 func decodeSFlowIpv4Record(data *[]byte) (SFlowIpv4Record, error) {
 	si := SFlowIpv4Record{}
+	var ir SFlowFlowDataFormat
+
+	*data, ir = (*data)[4:], SFlowFlowDataFormat(binary.BigEndian.Uint32((*data)[:4]))
```
safchain: I did test it and it seems to be not working for tunneling record. I tested it with Openswitch. Did you see any issue with tunneling record or was it with another kind of record ?

gconnell: safchain, do you have a test case that could be added to the tests in this pull?

safchain: @gconnell there is already a test that this PR modifies. According to my test with Openvswitch, it doesn't seem to be working properly with this patch. 

@lflxp Maybe the IPV4/IPV6 records are used somewhere else by sFlow protocol needing SFlowBaseFlowRecord. In that case we will need to use a king of decodeSFlowIpv4Data used by decodeSFlowIpv4Record and decodeExtendedIpv4TunnelEgress.


Below two records from a lab, first one without the patch second one with it. In the second one the SFlowIpv4Record is incorrect.

```
{SFlowBaseFlowRecord:{EnterpriseID:Standard SFlow Format:Extended IPv4 Tunnel Ingress Record FlowDataLength:32} SFlowIpv4Record:{Length:0 Protocol:47 IPSrc:10.41.20.10 IPDst:10.41.20.14 PortSrc:0 PortDst:0 TCPFlags:0 TOS:0}}

{SFlowBaseFlowRecord:{EnterpriseID:Standard SFlow Format:Extended IPv4 Tunnel Ingress Record FlowDataLength:32} SFlowIpv4Record:{SFlowBaseFlowRecord:{EnterpriseID:Standard SFlow Format: FlowDataLength:47} Length:170464266 Protocol:170464270 IPSrc:0.0.0.0 IPDst:0.0.0.0 PortSrc:0 PortDst:0 TCPFlags:3 TOS:1001}}
```

PR #402, opened 2018-01-02
layers: Improve OSPFv2 decoder

florianl:

This PR improves the OSPFv2 decoder.

### DIFF
--- a/layers/ospf.go
+++ b/layers/ospf.go
@@ -27,19 +27,19 @@ const (
 
 // LSA Function Codes for LSAheader.LSType
 const (
-	RouterLSAtypeV2          = 0x1
-	RouterLSAtypeV3          = 0x2001
-	NetworkLSAtypeV2         = 0x2
-	NetworkLSAtypeV3         = 0x2002
-	SummaryLSANetworktypeV2  = 0x3
-	InterAreaPrefixLSAtypeV3 = 0x2003
-	SummaryLSAASBRtypeV2     = 0x4
-	InterAreaRouterLSAtypeV3 = 0x2004
-	ASExternalLSAtypeV2      = 0x5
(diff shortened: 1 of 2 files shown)


gconnell:
What Laurent said :)

I try really hard to make sure that any changes to GoPacket don't break (IE: stop the compilation of) existing code that uses this library.  Given that, I tend to discourage any changes to constant/struct/method/variable names, as references to them from external code would cause said code to blow up.

For the constants, this could maybe be worked around with:

```go
const (
  NewVarname = 1
  OldVarname = NewVarname
)
```

For the struct renames, we might be able to do the same thing with type aliases, but I also worry that using a Go 1.9 feature could cause folks using older compilers to have some problems.  Of course, we'll use 1.9 features eventually, but ATM there's still somewhat new.

For now, could we keep all the V3 stuff with the old names, and just add the new V2 consts/structs?  Then folks using V3 stuff in their current code won't be broken, but people will be able to utilize V2 if they'd like.

Thoughts/comments welcome.

florianl:
Thanks for your comments. And I see your point.
Best solutions would be, to keep the V3 stuff with the old names. I will update the patch.

vtolstov:
@gconnell may be this is time to split layers to dedicated packages? So users can easy vendor needed layers and not think about breakage?
Also this helps to not clush namespace with something like DNSXXX DHCPXXX NTPXXX and so...

gconnell:
I've definitely considered it, but holy heck, talk about backwards incompatible ;)

We actually hit issues with godoc not being able to show all of our constants/structs that required we be special-cased, I think.

But yeah, my general strategy so far is to curl up in a ball and pretend this isn't a problem until it magically fixes itself.  So far it's going great.

PR #403, opened 2018-01-06
New APIs for setting pfring's poll_watermark and poll_duration parameters

gkodali-zededa:
PF_RING has default values of 128 packets and 500ms for poll_watermark and poll_duration respectively. It means the pf_ring's packet poll call would return only when either there are 128 packets queued in it's ring buffer or 500ms time has elapsed since the call to pfring_poll. This is affecting the performance in our use case. I have added these APIs to make these parameters flexible for change to programmers needs.

### DIFF
--- a/pfring/pfring.go
+++ b/pfring/pfring.go
@@ -222,7 +222,7 @@ func (r *Ring) SetSamplingRate(rate int) error {
 	}
 	return nil
 }
-
+// SetPollWatermark sets the pfring's poll watermark packet count
 func (r *Ring) SetPollWatermark(count uint16) error {
 	if rv := C.pfring_set_poll_watermark(r.cptr, C.u_int16_t(count)); rv != 0 {
 		return fmt.Errorf("Unable to set poll watermark, got error code %d", rv)
@@ -230,10 +230,12 @@ func (r *Ring) SetPollWatermark(count uint16) error {
 	return nil
 }
 
+// SetPriority sets the pfring poll threads CPU usage limit
 func (r *Ring) SetPriority(cpu uint16) {
 	C.pfring_config(C.u_short(cpu))
 }
 
+// SetPollDuration sets the pfring's poll duration before it yields/returns
 func (r *Ring) SetPollDuration(duration uint) error {
 	if rv := C.pfring_set_poll_duration(r.cptr, C.u_int(duration)); rv != 0 {
 		return fmt.Errorf("Unable to set poll duration, got error code %d", rv)
--- a/pfring/pfring.go
+++ b/pfring/pfring.go
@@ -223,6 +223,24 @@ func (r *Ring) SetSamplingRate(rate int) error {
 	return nil
 }
 
+func (r *Ring) SetPollWatermark(count uint16) error {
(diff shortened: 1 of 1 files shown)


gkodali-zededa:
Travis build failure is shown for my commit that only has source code comments.

gconnell:
I believe the Travis build is saying that your files need to be gofmt'd.  Could you try that and re-submit the pull?

gconnell:
Specifically, it probably wants a newline between lines 224 and 225

pfring/pfring.go
```diff
@@ -222,6 +222,26 @@ func (r *Ring) SetSamplingRate(rate int) error {
...
+	return nil
+}
+
+// SetPriority sets the pfring poll threads CPU usage limit
+func (r *Ring) SetPriority(cpu uint16) {
+	C.pfring_config(C.u_short(cpu))
+}
+
+// SetPollDuration sets the pfring's poll duration before it yields/returns
+func (r *Ring) SetPollDuration(duration uint) error {
```
gconnell: would we be able to pass in a time.Duration here and convert it into a u_int under the hood?

gkodali-zededa:
Hi Graeme,

Thanks for your response. I have submitted the changes required. Travis
build passes now.
Can you please take a look?

Thanks
Gopi

pfring/pfring.go
gkodali-zededa: I have just checked time.Duration and it is uint64 type representing nano second difference. pfring_set_poll_duration treats the value passed in as time in milliseconds. So there is no benefit in passing nano seconds and then immediately converting the value to milli seconds. Please let me know if I did not understand your point completely.

gconnell: The point is simply that, in Go world, there is a canonical type for Duration, and we should use that where possible.  My preference would be something like:

func (r *Ring) SetPollDuration(d time.Duration) {
  millis := <convert d to millis>
  ...call with millis...
}

gconnell: Alternatively, if you're wedded to passing in an integer of some sort, please rename the passed-in arg to durationMillis or something like that that makes it clear what time unit you're expecting.

gkodali-zededa: Yeah, I'll do that. Shall submit new code soon.

pfring/pfring.go
```diff
@@ -223,6 +223,27 @@ func (r *Ring) SetSamplingRate(rate int) error {
...
+	return nil
+}
+
+// SetPriority sets the pfring poll threads CPU usage limit
+func (r *Ring) SetPriority(cpu uint16) {
+	C.pfring_config(C.u_short(cpu))
+}
+
+// SetPollDuration sets the pfring's poll duration before it yields/returns
+func (r *Ring) SetPollDuration(durationMillis time.Duration) error {
```
gconnell: sorry, if this is millis, it should be:

```
durationMillis uint
```

not

```
durationMillis time.Duration
```

gkodali-zededa: Done. Submitted changes.

gconnell:
Thanks!

PR #413, opened 2018-01-30
Fix: stopped dhcpv4 option length increasing indefinitely 

rsabhilash:
While decoding dhcp4 using  DecodingLayerParser, as same object is reused, the `Options` field got appended with options of packets indefinitely. The `Options` field is cleared, initialized to zero length, to avoid this problem

### DIFF
--- a/layers/dhcpv4.go
+++ b/layers/dhcpv4.go
@@ -150,6 +150,7 @@ func (d *DHCPv4) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error 
 	}
 
 	options := data[240:]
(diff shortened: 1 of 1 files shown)


rsabhilash:
I signed it!

gconnell:
Thanks for the submission!

However, I believe that https://github.com/google/gopacket/commit/b8bab68a4d8cb61dc32887af5095121b2b1829b4 already fixed this issue by adding line 128.  Given that, I think this pull is redundant?  Let me know if I'm missing something :)

rsabhilash:
:), my bad, it's redundant

PR #418, opened 2018-02-06
[afpacket] return err when empty block

sharang:
Patch a12eb5bd add retry when empty block received,
but the retry loop will disable POLL_TIMEOUT.

Signed-off-by: Yang XIANG <xiangyang@yunshan.net.cn>


sharang:
I signed it!

PR #428, opened 2018-03-19
SetBPFFilter for af_packet

gkodali-zededa:
Currently af_packet only has SetBPF that requires us to pass compiled program. But, pcap has compile API's for converting match instructions in plain english strings to compiled programs.

This code is originally taken from the following commit to packet-beat
https://github.com/packetb-old/gopacket/commit/44554145590f42204f5f78cb2aeb45ba08cd2c2f

We use this in our code. That's why I'm pushing it to master. Please review.


gkodali-zededa:
Pull request does not look right. I'll raise another.

nogoegst:
@gkodali-zededa you could just squash/force push to your branch to make it look good.

PR #429, opened 2018-03-19
Current af_packet code does not have SetBPFFilter similar to pf_ring/…

gkodali-zededa:
Current af_packet does not have SetBPFFilter similar to pf_ring/pcap. SetBPF in af_packet needs us to pass compiled BPF program. SetBPFFilter uses pcap's compile functions to do the compilation for us. This code is originally commited to packetbeat as part of this commit. https://github.com/packetb-old/gopacket/commit/44554145590f42204f5f78cb2aeb45ba08cd2c2f

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -27,6 +27,8 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcap"
 )
 
 /*
@@ -61,6 +63,8 @@ type Stats struct {
 	// This should always be <= Packets, since with TPacket one syscall
 	// can (and often does) return many results.
 	Polls int64
+
+	PollRetries int64
 }
 
 // SocketStats is a struct where socket stats are stored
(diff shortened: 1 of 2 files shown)


gkodali-zededa:
I do not understand what travis build is reporting. Can someone please help?

afpacket/afpacket.go
```diff
@@ -27,6 +27,8 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcap"
```
gconnell: The reason we don't have this yet is that I'd like afpacket to compile if folks don't have libpcap installed... this creates a dependency on libpcap for afpacket-based binaries, which I don't think is ideal.

jcrussell:
I looked into this a while back and found this article [0] that talks about the limitations of the `pcap-filter` language and it mentions `pflua` [1] as an `libpcap`-independent way to compile `pcap-filters`. It would be great to have something like `pflua` written in Go.

[0] https://www.snellman.net/blog/archive/2015-05-18-whats-wrong-with-pcap-filters/
[1] https://github.com/Igalia/pflua

PR #434, opened 2018-04-17
Lack of support for data frames

Galadros:
As far as I can tell, gopacket doesn't support processing the frames with subtype "Data" (That is, frames with type 10 and subtype 0000).  I've added support to process said frames symmetrical to how other data frames are processed.

Frankly, I suspect I may have missed or misunderstood something- it doesn't make sense to me that something which seems very obvious to me would simply lack support in gopacket.  Have I made some mistaken assumptions?

### DIFF
--- a/layers/dot11.go
+++ b/layers/dot11.go
@@ -516,13 +516,11 @@ func decodeDot11DataData(data []byte, p gopacket.PacketBuilder) error {
 	return decodingLayerDecoder(d, data, p)
 }
 
-func (m *Dot11DataData) LayerType() gopacket.LayerType {return LayerTypeDot11DataData}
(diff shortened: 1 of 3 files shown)


Galadros:
Frankly, this is somewhat of a formatting update, not a true feature lack.  There is no support for parsing the "Data" subtype in 802.11.  While I recognize that you can simply use the "Data" type instead, that doesn't fit with the structure of the rest of gopacket- you can't assert a frame directly as "Dot11Mgmt," for instance.

Galadros:
Awkwardly, the written update is incompatible with the significant changes made to dot11.go within the past two weeks or so.  Oh well.

Issue #435, opened 2018-04-17
bug - gopacket, IPv6 Hop-by-Hop, and ICMPv6 Type


richard-jp-leguen:
I've been using gopacket to examine ICMPv6 messages. I've discovered that a packet which includes an IPv6 Hop-by-Hop doesn't seem to get decoded properly. It looks to me like `layers.IPv6` includes the Hop By Hop Options in the bytesto be decoded as part of the "next layer".

Example
=======

Here's my packet:

```
33 33 00 00 00 16 1e c3 e3 b7 c4 d5 86 dd 60 00
00 00 00 88 00 01 00 00 00 00 00 00 00 00 00 00
00 00 00 00 00 00 ff 02 00 00 00 00 00 00 00 00
00 00 00 00 00 16 3a 00 05 02 00 00 01 00 8f 00
9e ed 00 00 00 06 03 00 00 00 ff 02 00 00 00 00 
00 00 00 00 00 01 ff b7 c4 d5 03 00 00 00 ff 02
00 00 00 00 00 00 00 00 00 01 ff 00 00 00 04 00
00 00 ff 02 00 00 00 00 00 00 00 00 00 01 ff 11
00 79 04 00 00 00 ff 02 00 00 00 00 00 00 00 00
00 01 ff 00 00 01 04 00 00 00 ff 05 00 00 00 00
00 00 00 00 00 00 00 00 00 02 04 00 00 00 ff 02
00 00 00 00 00 00 00 00 00 00 00 00 00 02
```
And the go code I'm using to parse it:

```
    package main

    import (
        "fmt"
        "github.com/google/gopacket"
        "github.com/google/gopacket/layers"
    )

    func main() {
        data := []byte {
            // Ethernet layer
             0x33, 0x33, 0x00, 0x00, 0x00, 0x16, // destination
             0x1e, 0xc3, 0xe3, 0xb7, 0xc4, 0xd5, //  source
             0x86, 0xdd, // type IPv6

             // IPv6 layer
             0x60, 0x00, 0x00, 0x00, // version; traffic class; flow label
             0x00, 0x88, // payload length?
             0x00, // Next Header
             0x01, // Hop Limit
             0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, // source
[...]

PR #438, opened 2018-04-24
Adjusting beginning of payload for IPv6 packets which include HopByHop options

richard-jp-leguen:
Without this fix the bytes for the HopByHop options end up beind decoded as part of the next layer

https://github.com/google/gopacket/issues/435


gconnell:
Happy to merge, just need the CLA handled.

gconnell:
The page isn't out of date... that's internal documentation for Googlers,
so all the links and emails have been redacted as part of open-sourcing the
docs, I think.

cla.developers.google.com should have any information you may need... let
me know if it doesn't!

gconnell:
Thanks!

PR #445, opened 2018-05-04
Update ip4defrag with correct limits from RFC791

gustafj:
This patch updates the limits set on the ip4defrag Layer to follow
the definition in RFC791.
Fragment offset according to the RFC:
    This format allows 2**13 = 8192 fragments of 8 octets each for a
    total of 65,536 octets. Every internet module must be able to
    forward a datagram of 68 octets without further fragmentation.
    This is because an internet header may be up to 60 octets, and the
    minimum fragment is 8 octets.

### DIFF
--- a/ip4defrag/defrag.go
+++ b/ip4defrag/defrag.go
@@ -31,11 +31,12 @@ func (d debugging) Printf(format string, args ...interface{}) {
 }
 
 // Constants determining how to handle fragments.
+// Reference RFC 791, page 25
 const (
-	IPv4MinimumFragmentSize    = 576   // Minimum size of a single fragment
+	IPv4MinimumFragmentSize    = 8     // Minimum size of a single fragment
 	IPv4MaximumSize            = 65535 // Maximum size of a fragment (2^16)
-	IPv4MaximumFragmentOffset  = 8189  // Maximum offset of a fragment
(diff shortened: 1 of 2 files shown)


gustafj:
I signed it!

lhausermann:
Hey @gustafj thanks for patching our IPv4 Defrag. Always good to have new contributor!

As said in my review, I don't really get why you are removing a unit test. I would appreciate to extend / grow the test list, it would ease the merge.. because we are sure the behavior is still the same based on the unit test

gustafj:
Hi, @lhausermann, thanks for the feedback.

The reason is that I am unable to trigger the scenario that was tested with that specific test without extensive mocking.
The test case fails prior to the tested value `IPv4MaximumFragmentListLen` due to triggering the `securityChecks` of `IPv4MaximumFragmentOffset`.
Reason for this is that the ip4defrag discards duplicated fragments, and thus fragment offset needs to change for each fragment, but even with offsets of 1 it will trigger `IPv4MaximumFragmentOffset` check since that one is less that `IPv4MaximumFragmentListLen` (8183 < 8192).
(And I did not want to remove the length check altogether).

gconnell:
I'll admit I straight up didn't believe you that the RFC actually said it allowed fragments as small as 8 bytes... but yeah it does.  Geez, that's tiny :D

PR #446, opened 2018-05-04
(layers) Added support for LCM protocol

TacoVox:
This patch features support for LCM layers. The code is based on the LCM
protocol definition (https://lcm-proj.github.io/).
LCM is a lightweight communications protocol that is used e.g. in the
automotive industry for testing and developing purposes.

The submitted patch allows for decoding LCM headers plus the registration
of underlying decoders (which should be used in combination with an
incoming patch towards the LCM GitHub repository that will emit gopacket
code based on LCM defintions).

### DIFF
--- a/layers/lcm.go
+++ b/layers/lcm.go
@@ -1,0 +1,193 @@
+// Copyright 2018 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/gopacket"
+)
+
+const (
+	// LCMShortHeaderMagic is the LCM small message header magic number
+	LCMShortHeaderMagic uint32 = 0x4c433032
+	// LCMFragmentedHeaderMagic is the LCM fragmented message header magic number
+	LCMFragmentedHeaderMagic uint32 = 0x4c433033
+)
+
+// LCM (Lightweight Communications and Marshalling) is a set of libraries and
+// tools for message passing and data marshalling, targeted at real-time systems
+// where high-bandwidth and low latency are critical. It provides a
+// publish/subscribe message passing model and automatic
+// marshalling/unmarshalling code generation with bindings for applications in a
+// variety of programming languages.
+//
+// References
+//   https://lcm-proj.github.io/
+//   https://github.com/lcm-proj/lcm
+type LCM struct {
+	// Common (short & fragmented header) fields
+	Magic          uint32
+	SequenceNumber uint32
+	// Fragmented header only fields
+	PayloadSize    uint32
+	FragmentOffset uint32
+	FragmentNumber uint16
+	TotalFragments uint16
+	// Common field
+	ChannelName string
+	// Gopacket helper fields
+	Fragmented  bool
+	fingerprint uint64
+	contents    []byte
+	payload     []byte
+}
+
+var (
+	lcmLayerTypes  = map[uint64]gopacket.LayerType{}
+	layerTypeIndex = 1001
+)
+
+// RegisterLCMLayerType allows users to register decoders for the underlying
+// LCM payload. This is done based on the fingerprint that every LCM message
+// contains and which identifies it uniquely. If num is not the zero value it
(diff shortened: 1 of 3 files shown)


TacoVox:
I signed it!

lhausermann:
Hi @TacoVox . Great to have a new contrib.
I am not a LCM expert, so hard to review the code on the dissection part. 
A small comment : looks like you did not used gofmt ?
https://github.com/google/gopacket/blob/master/CONTRIBUTING.md specifies a few rules, we try to enforce before merging PR. Would you please go through them ?

Thanks again !

TacoVox:
Hi @lhausermann!
Yes, everthing is formatted using gofmt. It has to, due to `.travis.gofmt.sh`.

So which lines are you referring to?

layers/lcm.go
```diff
@@ -0,0 +1,190 @@
...
+package layers
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/google/gopacket"
+)
+
+// Lightweight Communications and Marshalling (LCM)
```
gconnell: Move this comment down to be the godoc comment for the LCM struct, so it shows up in Godoc?

layers/lcm.go
```diff
@@ -0,0 +1,190 @@
...
+
+	if lcm.Magic != LCMShortHeaderMagic && lcm.Magic != LCMFragmentedHeaderMagic {
+		return fmt.Errorf("Received LCM header magic %v does not match know "+
+			"LCM magic numbers. Dropping packet.", lcm.Magic)
+	}
+
+	lcm.SequenceNumber = binary.BigEndian.Uint32(data[offset:8])
+	offset += 4
+
+	if lcm.Magic == LCMFragmentedHeaderMagic {
```
gconnell: I'm not at all familiar with fragmentation in LCM, but if it DOES contain fragments of a packet within it, take a look at the toplevel gopacket.Fragment layer type and see if that might be a good thing to return as a next layer type in these cases.

layers/lcm.go
```diff
@@ -0,0 +1,190 @@
...
+func (lcm LCM) CanDecode() gopacket.LayerClass {
+	return LayerTypeLCM
+}
+
+// NextLayerType specifies the LCM payload layer type following this header.
+// As LCM packets are serialized structs with uniq fingerprints for each uniq
+// combination of data types, lookup of correct layer type is based on that
+// fingerprint.
+func (lcm LCM) NextLayerType() gopacket.LayerType {
+	return getLCMLayerType(lcm.fingerprint)
```
gconnell: As mentioned in another comment, for fragmented packets consider returning gopacket.LayerTypeFragment at least for fragments after the first one, so that gopacket doesn't try to decode a fragment of a layer that it has no hope of decoding correctly.

lhausermann:
I was thinking first that the way the main struct fields are misaligned was a non formatting issue.. but I realize that is due to the comments in the middle. 
Just ignore my initial remark.

layers/lcm_test.go
```diff
@@ -0,0 +1,156 @@
...
+func testFragmentedLCM(t *testing.T) {
+	lcm := LCM{}
+
+	err := lcm.DecodeFromBytes(fragmentedPacket, gopacket.NilDecodeFeedback)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if lcm.Magic != LCMFragmentedHeaderMagic {
+		t.Fatalf("Expected LCM Magic %x, but decoded %x.\n",
```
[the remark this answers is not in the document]
TacoVox: I will do it the Errorf-way, as I have dealt with some libraries that do it the DeepEqual-way. With the latter, I always found it harder to read and find the mismatching parts. I hope that works with you :)

layers/lcm.go
TacoVox: Good call! I will add this.

TacoVox:
Hi @gconnell!
Thanks for your nice comments. I improved the code based on that.
The particularly liked the call for returning `gopacket.LayerTypeFragment` in case of a fragmented message. I actually haven\t told you that I open another PR after this one is in for a defragmenter of LCM messages that will then do the defragmentation :sunglasses: 

@lhausermann Nooo problem :)

gconnell:
Thanks for the help!

Issue #441, opened 2018-04-28
af_packet: zero copy is broken


styleex:
In commit 9d0c2435c1e2a38eef8f8d67e0377212388f8d0d was added code that creates a new buffer in memory for each package:

https://github.com/google/gopacket/commit/9d0c2435c1e2a38eef8f8d67e0377212388f8d0d#diff-4b6810f6883b99dcc5dcebf00351c417R69

styleex:
I now checked it on my virtual machine: when rx (tx)-vlan-offloading is disabled linux gives a full ethernet package, including information about vlan.

Accordingly, the question: why did you choose to implement such a hack, instead of disabling the unnecessary function of the network card. In this case, this behavior of linux is documented.

My suggestion: instead of this hack document the need to disable vlan-offloading on the network card.

From the pros: for vlan traffic will also be zero-copy reading packets.

gconnell:
Happy to accept a fix for this, if you'd like to send a pull :)

safchain:
sorry for the delay, I was on vacation.

I do understand your concern about the performances but I think that is a bit weird to change the configuration of something you want to monitor/troubleshoot, moreover the kernel gives the information so I think it's worth it to provide it. Why not capturing the untagged interfaces ? I would rather add an option to enable vlan header insertion or adding the vlan info as part of the capture info ?

gconnell:
Hey, Антон and Sylvain,

The overall point of this library is to do the right thing as much as possible, even if/when things aren't otherwise configured to exact specifications.  That said, we also want fidelity of packets to be a high priority, and it seems like disabling rx-vlan-offloading might aid in that (and keep performance better by avoiding a packet-copy).  Therefore, from my understanding of things, I think the best approach is:

1) keep adding in the vlan header if it's not captured but is available in the AF_PACKET info, as we do right now
2) explore auto-disabling rx/tx-vlan-offloading, and if it has no other impact on performance, switch it off by default and disable the code from 1
3) regardless, in cases where rx/tx-vlan-offloading is turned off, make sure that we don't double-add VLAN headers

Does that sound good to folks?

  --Graeme

PS:  Please avoid using words like "hack" when describing any code that isn't your own.  While I'm sure it wasn't meant to be derogatory, it can be read that way.  Everyone's contributing code here because they want things to be improved, and they're all volunteers. :)

styleex:
[...]

PR #447, opened 2018-05-04
AF_PACKET:  Adding back VLAN header optional.

gconnell:
Makes the adding back of the VLAN header for AF_PACKET packet data
optional and defaulted to off.  Especially for ZeroCopy, the addition of
VLAN headers is a performance hit (I clocked a 10x slowdown on a fast,
VLAN'd network).  While unfortunately it's counter-intuitive that the
kernel (or NIC, or whatever) removes the VLAN header, it's also
counter-intuitive that we return something to the user that the kernel
didn't return to us.  So, let's go for the thing that's at least fast,
while allowing for users to request the VLAN header if they'd like.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -283,7 +283,7 @@ retry:
 			goto retry
 		}
 	}
-	data = h.current.getData()
+	data = h.current.getData(&h.opts)
 	ci.Timestamp = h.current.getTime()
 	ci.CaptureLength = len(data)
 	ci.Length = h.current.getLength()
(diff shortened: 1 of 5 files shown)


gconnell:
Fixes #441

gconnell:
Here's the performance numbers I referenced.  This is gathering 2.5M packets off a relatively fast network, with no additional processing (using the new example script).  Note that `getData` takes 190ms of CPU time without VLAN addition and 1.66s of CPU time with it.

pprof002
pprof001

gconnell:
Having heard no objections to this, I'm putting it in place.

PR #449, opened 2018-05-08
LCM Defragmentation Support

TacoVox:
Hello again!

This patch includes support for defragmentation of LCM packages, as specified in the LCM protocol description (https://lcm-proj.github.io/udp_multicast_protocol.html).

As my PR before, this patch is intended to be used in combination with an open PR (https://github.com/lcm-proj/lcm/pull/220) towards the GitHub LCM repository that adds golang + gopacket support.

As you can see, I put the defragmenter into a package called `defrag`. My suggestion is that the IPv4 defragmenter is also moved there as well and is renamed to `ipv4defrag.go` (and `ipv4defrag_test.go` respectively). IMHO this would be a bit more cleaner in case new defragmenters should come in. If you agree, I would open another PR for that.

Thanks for your openness towards new protocols :+1:  :sunglasses:

### DIFF
--- a/.travis.golint.sh
+++ b/.travis.golint.sh
@@ -3,7 +3,7 @@
 cd "$(dirname $0)"
 
 go get github.com/golang/lint/golint
-DIRS=". tcpassembly tcpassembly/tcpreader ip4defrag reassembly macs pcapgo pcap afpacket pfring routing"
+DIRS=". tcpassembly tcpassembly/tcpreader ip4defrag reassembly macs pcapgo pcap afpacket pfring routing defrag"
 # Add subdirectories here as we clean up golint on each.
 for subdir in $DIRS; do
   pushd $subdir
--- a/.travis.golint.sh
+++ b/.travis.golint.sh
@@ -3,7 +3,7 @@
 cd "$(dirname $0)"
 
 go get github.com/golang/lint/golint
-DIRS=". tcpassembly tcpassembly/tcpreader ip4defrag reassembly macs pcapgo pcap afpacket pfring routing defrag"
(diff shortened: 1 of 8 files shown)


TacoVox:
Hello @gconnell!
I just wanted to kindly ask, if we somehow can get this in? :) Right now, we are using this already in production and it works fine :) But obviously, we do not want to have the dependency to my personal clone, but rather to this repository. Thanks!

gconnell:
Hey, Taco, sorry for the slow response on this; review just slipped my mind.

The biggest sticking point for me right now is the `defrag` subdirectory... what would you think of putting this code into something like `defrag/lcmdefrag`, so we don't have all the future defrag stuff in a single directory and run into a problem like we have now with `layers` being a catch-all?

TacoVox:
No problem!

Yeah, that sounds like a good idea. I guess that this would then also apply for the ip4defrag packet (which IMHO should be under defrag as well)?

Let me amend on Monday! Have a nice weekend!

gconnell:
yep, sounds good.  let's not do the ip4defrag switch as part of this pull (we'll want to do some silly type aliasing and stuff to not break compatibility), but I think long-term it's a fine idea.

TacoVox:
Yeah, I wasn't talking about this PR. Should definetely be a seperate one!

TacoVox:
Hi @gconnell. Moved to a new package. Sorry 'bout the merge commit there. Not quite sure how we can get rid of it...

TacoVox:
Ignore the mess that was in between here. Now it looks nice :sunglasses:

gconnell:
Great, thanks for the fast turn-arounds!

PR #452, opened 2018-05-10
Feature/lldp

baba147:
Hi,

In addition to my previous comments, here is how lldp is used. I am using WireShark to verify generated packets:

```
	lldp := layers.LinkLayerDiscovery{
		ChassisID: layers.LLDPChassisID{
			Subtype: layers.LLDPChassisIDSubtypeIfaceAlias,
			ID:      net.HardwareAddr{0x50, 0x3E, 0xAA, 0x41, 0xFB, 0x46},
		},
	}
	err = gopacket.SerializeLayers(buf, opts, &layers.Ethernet{
		SrcMAC:       net.HardwareAddr{0x50, 0x3E, 0xAA, 0x41, 0xFB, 0x46},
		DstMAC:       net.HardwareAddr{0x01, 0x80, 0xC2, 0x00, 0x00, 0x00},
		EthernetType: layers.EthernetTypeLinkLayerDiscovery,
	}, &lldp)
```

### DIFF
--- a/layers/lldp.go
+++ b/layers/lldp.go
@@ -36,6 +36,11 @@ type LinkLayerDiscoveryValue struct {
 	Length uint16
 	Value  []byte
 }
+type LLDPSerializableTLV interface {
+	TLVID() byte
+	Serialize() []byte
+	Len() int
+}
 
 // LLDPChassisIDSubType specifies the value type for a single LLDPChassisID.ID
 type LLDPChassisIDSubType byte
@@ -55,6 +60,21 @@ const (
 type LLDPChassisID struct {
 	Subtype LLDPChassisIDSubType
 	ID      []byte
+}
+
+func (c *LLDPChassisID) TLVID() byte {
+	return byte(LLDPTLVChassisID)
+}
+func (c *LLDPChassisID) Serialize() []byte {
+	var buf = make([]byte, c.Len())
+	idLen := uint16(c.TLVID())<<9 | uint16(len(c.ID)+1) //id should take 7 bits, length should take 9 bits, +1 for subtype
+	binary.BigEndian.PutUint16(buf[0:2], idLen)
+	buf[2] = byte(c.Subtype)
+	copy(buf[3:], c.ID)
+	return buf
+}
+func (c *LLDPChassisID) Len() int {
+	return len(c.ID) + 3 // +2 for id and length, +1 for subtype
 }
 
 // LLDPPortIDSubType specifies the value type for a single LLDPPortID.ID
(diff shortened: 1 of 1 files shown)


layers/lldp.go
```diff
@@ -36,6 +36,11 @@ type LinkLayerDiscoveryValue struct {
 	Length uint16
 	Value  []byte
 }
+type LLDPSerializableTLV interface {
```
gconnell: Where/how is this useful?  I'm not seeing it used anywhere, and I'm hesitant to add something specifically for serializing a single layer type.

layers/lldp.go
```diff
@@ -57,6 +62,21 @@ type LLDPChassisID struct {
 	ID      []byte
 }
 
+func (c *LLDPChassisID) TLVID() byte {
```
gconnell: Please add godoc comments to all exported functions.

gconnell: Also, should probably return LLDPTLVType

layers/lldp.go
```diff
@@ -57,6 +62,21 @@ type LLDPChassisID struct {
...
+	return byte(LLDPTLVChassisID)
+}
+func (c *LLDPChassisID) Serialize() []byte {
```
gconnell: do these functions need to be exported?

layers/lldp.go
```diff
@@ -57,6 +62,21 @@ type LLDPChassisID struct {
...
+	var buf = make([]byte, c.Len())
+	idLen := uint16(c.TLVID())<<9 | uint16(len(c.ID)+1) //id should take 7 bits, length should take 9 bits, +1 for subtype
+	binary.BigEndian.PutUint16(buf[0:2], idLen)
+	buf[2] = byte(c.Subtype)
+	copy(buf[3:], c.ID)
+	return buf
+}
+func (c *LLDPChassisID) Len() int {
```
gconnell: `Len` seems overly generalized... maybe `SerializedLen`?

layers/lldp.go
baba147: Actually my plan was to use this for organizationally specific TLV. However, I realized we can have complete functionality without this. I will remove this.

layers/lldp.go
baba147: Noted.

layers/lldp.go
baba147: I suppose we don't need to export these. I will keep in mind.

layers/lldp.go
baba147: Noted.

layers/lldp.go
```diff
@@ -37,6 +37,10 @@ type LinkLayerDiscoveryValue struct {
 	Value  []byte
 }
 
+func (c *LinkLayerDiscoveryValue) len() int {
```
gconnell: going to merge as is, but is this function actually used?  if not, please remove in future pull.

PR #453, opened 2018-05-13
layers: Decode ICMPv6 Pings

florianl:

Add decoder for ICMPv6 pings.


gconnell:
Thanks!

PR #461, opened 2018-05-31
-Adding missing parsing in ospf.go

congdoe:
-Added missing parsing within extractLSAInformation for RouterV2 structs.
-Added field, Number of Metric, for RouterV2 struct


congdoe:
I signed it!

PR #462, opened 2018-06-06
Add support for out-of-band VLAN IDs on afpacket sources.

deadpixi:
The Linux kernel supports VLAN offloading, which will strip off VLAN headers before passing the packet data up to the capture layer. This results in the odd situation where the data presented to the capture layer is not what was actually visible on the wire.

For those who needed visibility into the VLAN layer, the `afpacket.OptAddVLANHeader` option was provided. When set, a synthetic VLAN header is created and copied over. While this works and provides a more accurate view of the data on the wire, it is slow (as noted in the documentation of the option, sometimes by an order of magnitude).

This commit augments the `gopacket.CaptureInfo` structure by adding a `VLAN` field that stores the VID (VLAN ID) of the packet if it was provided out-of-band by the capture mechanism. If it was not provided, this field defaults to -1 (it should not default to zero, as it is possible to have a zero VID, which is used for priority tagging).

This provides a fast mechanism to access VLAN information for those who need it while retaining the `afpacket.OptAddVLANHeader` mechanism for those who need it.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -288,6 +288,7 @@ retry:
 	ci.CaptureLength = len(data)
 	ci.Length = h.current.getLength()
 	ci.InterfaceIndex = h.current.getIfaceIndex()
+	ci.VLAN = h.current.getVLAN()
 	atomic.AddInt64(&h.stats.Packets, 1)
 	h.headerNextNeeded = true
 	h.mu.Unlock()
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -53,6 +53,13 @@ func tpacketAlign(v int) int {
 	return int((uint(v) + tpacketAlignment - 1) & ((^tpacketAlignment) - 1))
 }
 
+// AncillaryVLAN structures are used to pass the captured VLAN
+// as ancillary data via CaptureInfo.
+type AncillaryVLAN struct {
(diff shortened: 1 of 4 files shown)


deadpixi:
I signed it!

deadpixi:
Perhaps it would be better to have a sice of `interface{}`s in the returned `CaptureInfo` that are "ancillary data" or something? It could be populated with, for example, `afpacket.AuxDataVLAN` or something that could be discovered at runtime.That would keep it generic but allow extra information should the capture layer be able to provide it.

This would be analogous to the ancillary data on UNIX domain sockets.

gconnell:
Hey, sorry for the slow response, just coming back from a vacation.

I like the idea of ancillary data.  So I'm guessing it would be something like this:

```go
func (c *CaptureInfo) AddAncillary(a interface{}) { ... }
func (c CaptureInfo) Ancillary() []interface{} { ... }
```

And it would be used like this:

```go
for _, a := range ci.Ancillary() {
  switch v := a.(type) {
    case afpacket.AncillaryVLAN:
      log.Printf("vlan ID %d", int(v))
  }
}
```

deadpixi:
@gconnell I've switched to a more generic "ancillary data" mechanism now in my branch. Currently it's simply a slice of `interface{}` that can be appended to by the packet provider; I can switch to using accessor methods if you think that's better. Please let me know.

Issue #459, opened 2018-05-22
How to access the af_packet statistics?


styleex:
Type SocketStatsV3 is a structure from C, how can I access the values of its fields?

The following code is not compiled:

```
s1, s2, err: = inputAFpacket.SocketStats ()
log.Println (s2.tp_drops)
```

gconnell:
Hey, I don't know specifically, but try this first:

```go
log.Printf("%+v", s2)
```

See what that shows, and it might help out.  This is a general solution, so if it works here it'll work anywhere :D

gconnell:
What does it print?  I'm hoping that if it prints `foo: 1`, then s2.foo will work.

styleex:
Yes.

`Printf("%+v", s2)`: "{tp_packets:273306075 tp_drops:92749 tp_freeze_q_cnt:11}"

`s2.tp_drops` build error: `2.tp_drops undefined (cannot refer to unexported field or method tp_drops)`

styleex:
By the way, why is the summation of the received statistics with the old state? https://github.com/google/gopacket/blob/master/afpacket/afpacket.go#L348

gconnell:
To answer your last question, it's because every getsockopt call resets the counters, and we wanted to provide the counters since the beginning of time.

styleex:
By the way, for exporting metrics, for example, in prometheus it would be more convenient if the difference between two readings was given. Then you can use the metric type Counter, not Guage.

In my opinion, generally, the binding with the code should be as transparent as possible and do not make unnecessary logic. Preaggregation of the metric can be done in a higher-level wrapper.

gconnell:
Use of tools like prometheus is actually exactly why we sum.  Counter assumes the value never resets to zero and is from beginning-of-time.  The problem with resetting to zero is that the sample rate matters a ton, while if we DONT reset to zero, we can accidentally miss a sampling and still be fine.

styleex:
Well, not exactly. Counter is stored on the monitoring system side, we send it diffs of this metric and the monitoring system itself adds this diff with the old value.

I mean that this summation can be performed anywhere, and not in binding:

1. In a high-level code (golang instead of cgo)
2. In the monitoring system

Therefore, it would be cool if the method of obtaining statistics worked exactly as described in the manuals. And another, higher-level code-utility for golang, could preaggregate it and give pre-aggregated values.

gconnell:
[...]

PR #463, opened 2018-06-06
Added accessors to the private socket stats fields.

deadpixi:
The `SocketStats` and `SocketStatsV3` fields are private by default (as noted in #459). This simply adds accessors for them.


deadpixi:
@caleron Please see the latest commit.

PR #466, opened 2018-06-11
ReadPacketDataNoCopy - return only reference to the data

assafmo:
`ReadPacketData` and `readPacketHeader` returns `[]bytes` and `gopacket.CaptureInfo` by value which leads to a lot of memory allocation at the calling end.  
`ReadPacketDataNoCopy` and `readPacketHeaderNoCopy` also does the allocations, but returns `*[]bytes` and `*gopacket.CaptureInfo` which avoids memory allocations at the calling end.

From my limited tests, this increases reading speed by ~20%.

screenshot from 2018-06-11 12-53-00

### DIFF
--- a/pcapgo/read.go
+++ b/pcapgo/read.go
@@ -139,6 +139,35 @@ func (r *Reader) readPacketHeader() (ci gopacket.CaptureInfo, err error) {
 	return
 }
 
+// ReadPacketDataNoCopy reads next packet from file, returns reference to the data.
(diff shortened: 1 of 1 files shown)


assafmo:
I signed it!

assafmo:
Acctually this may increase GC time bcause the pointer escape prevents Go from allocating the buffers on the stack.

I'm not sure how to debug this. @gconnell any thoughts?

assafmo:
This doesn't help at all. Go passes slices by pointer. The above picture is misleading, `newobject` is called is different places.

Issue #439, opened 2018-04-26
OOM/infinite loop on malformed packets


JustinAzoff:
I ran gopacket through gofuzz for a while (forgot it was running)

```
// +build gofuzz
package gopacketfuzz

import (
	"github.com/google/gopacket"
	"github.com/google/gopacket/layers"
)

func Fuzz(data []byte) int {
	packet := gopacket.NewPacket(data, layers.LayerTypeEthernet, gopacket.Default)
	if packet == nil {
		return 0
	}
	layers := packet.Layers()
	if len(layers) == 0 {
		return 0
	}
	return 1
}
```

And found 3 packets that cause gopacket to use all memory before  being killed:

```
package main

import (
	"log"

	"github.com/google/gopacket"
	"github.com/google/gopacket/layers"
)

var packets = [][]byte{
	// workdir/crashers/f951a2ef03a3f465b0b75a4d9510a88d56404afa
	[]byte{
		0x5c, 0xfa, 0x00, 0x00, 0xfa, 0x6c, 0xbd, 0xe4, 0x19, 0x72, 0x1d, 0x1f,
		0x08, 0x00, 0x45, 0x00, 0x00, 0xad, 0x00, 0x00, 0x40, 0x00, 0x40, 0x59,
		0x6e, 0xef, 0x6b, 0x96, 0x61, 0x11, 0xb4, 0x96, 0xc0, 0xa8, 0x02, 0x04,
		0x00, 0xa2, 0x80, 0x00, 0x01, 0x1f, 0x08, 0x00, 0x45, 0x00, 0x6f, 0x76,
		0x65, 0x72, 0x66, 0x6c, 0x7f, 0xff, 0x12, 0x12, 0x00, 0x5a, 0xf6, 0x73,
		0xa2, 0xba, 0x81, 0x90, 0x00, 0x01, 0x40, 0x6f, 0x77, 0x20, 0x6f, 0x6e,
		0x20, 0x63, 0x68, 0x61, 0x72, 0x61, 0x63, 0x74, 0x00, 0x00, 0x00, 0x01,
		0x96, 0x20, 0x76, 0x61, 0x6c, 0x75, 0x65, 0x20, 0x00, 0x6e, 0x00, 0x6f,
		0x77,
	},
	// workdir/crashers/b2fc2d04a6b15bd53f9d76f0a769038ea82d21b3
	[]byte{
[...]

PR #468, opened 2018-06-14
Fix infinite loop when parsing malformed OSPF packet

rstrand:
Fix the infinite loop when parsing OSPF packets discovered in issue #439


rstrand:
I've just rebased this on master so the tests now pass

gconnell:
Thanks for the fix!

PR #469, opened 2018-06-14
Fix test errors in reassembly

rstrand:


rstrand:
It looks like go 1.10 added cgo support for go vet which is why this has only just started to be an issue
https://golang.org/doc/go1.10#vet

gconnell:
Oops, apologies.  I just fixed this in another pull.  Could you please sync from master and see if things are still working as expected for you?  Just getting back from vacation, so I'm working through a backlog :)

rstrand:
No problem, I'll close this pull request as you have fixed it already!

PR #470, opened 2018-06-17
pcapgo: add OpenEthernet and a full capture example

stapelberg:
I just implemented this in a project of mine and figured it might make a good addition to the pcapgo package :)

### DIFF
--- a/pcapgo/capture.go
+++ b/pcapgo/capture.go
@@ -1,0 +1,52 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package pcapgo
+
+import (
+	"net"
+	"syscall"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/mdlayher/raw"
+)
+
+// EthernetHandle wraps a raw.Conn, implementing gopacket.PacketDataSource so
+// that the handle can be used with gopacket.NewPacketSource.
+type EthernetHandle struct {
+	*raw.Conn
+}
+
+// ReadPacketData implements gopacket.PacketDataSource.
+func (h *EthernetHandle) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
+	b := make([]byte, 4096) // TODO(correctness): how much space do we need?
+	n, _, err := h.ReadFrom(b)
+	if err != nil {
+		return nil, gopacket.CaptureInfo{}, err
+	}
+	data := b[:n]
+	return data, gopacket.CaptureInfo{
+		CaptureLength: len(data),
+		Length:        len(data),
+		Timestamp:     time.Now(),
+	}, nil
+}
+
+// NewEthernetHandle implements pcap.OpenLive for ethernet interfaces only.
+func NewEthernetHandle(ifname string) (*EthernetHandle, error) {
+	intf, err := net.InterfaceByName(ifname)
+	if err != nil {
+		return nil, err
+	}
+
+	conn, err := raw.ListenPacket(intf, syscall.ETH_P_ALL, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	return &EthernetHandle{conn}, nil
+}
--- a/pcapgo/capture_test.go
+++ b/pcapgo/capture_test.go
@@ -1,0 +1,40 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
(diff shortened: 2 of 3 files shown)


stapelberg:
Friendly ping? :)

pcapgo/capture.go
```diff
@@ -0,0 +1,48 @@
+package pcapgo
```
gconnell: please copy over copyright notices to top of new files.

pcapgo/capture.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// ReadPacketData implements gopacket.PacketDataSource.
+func (h *EthernetHandle) ReadPacketData() ([]byte, gopacket.CaptureInfo, error) {
+	b := make([]byte, 4096) // TODO(correctness): how much space do we need?
+	n, _, err := h.ReadFrom(b)
+	if err != nil {
+		return nil, gopacket.CaptureInfo{}, err
+	}
+	packet := gopacket.NewPacket(b[:n], layers.LayerTypeEthernet, gopacket.Default)
```
gconnell: Don't create a new packet and return it.  Just return b[:n], and gopacket's https://github.com/google/gopacket/blob/master/packet.go#L794 will wrap it into a packet.
With the current code, a full packet decode happens, then the entire packet object is thrown out.

pcapgo/capture.go
```diff
@@ -0,0 +1,48 @@
...
+	data := packet.Data()
+	return data, gopacket.CaptureInfo{
+		CaptureLength: len(data),
+		Length:        len(data),
+		Timestamp:     time.Now(),
+	}, nil
+}
+
+// OpenEthernet implements pcap.OpenLive for ethernet interfaces only.
+func OpenEthernet(ifname string) (*EthernetHandle, error) {
```
gconnell: to make things perfectly clear, maybe "NewEthernetHandle(...)"

pcapgo/capture_test.go
```diff
@@ -0,0 +1,34 @@
+package pcapgo_test
```
gconnell: copyright notice here plz

pcapgo/capture_test.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"log"
+	"os"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+	"github.com/google/gopacket/pcapgo"
+)
+
+func Example_captureEthernet() {
```
gconnell: nit: CaptureEthernet

pcapgo/capture_test.go
```diff
@@ -0,0 +1,34 @@
...
+	f, err := os.Create("/tmp/lo.pcap")
```
gconnell: I've never seen this before... why does it work?  Why does this file translate to net.InterfaceByName working?  Inquiring minds like mine want to know :)

pcapgo/capture.go
stapelberg: Done

pcapgo/capture.go
stapelberg: Whoops, of course. Done.

pcapgo/capture_test.go
stapelberg: Done.

pcapgo/capture.go
stapelberg: Done.

stapelberg:
I addressed some of your comments, but got to run now :) I’ll try to address the remaining ones later.

Thanks for your review!

pcapgo/capture_test.go
stapelberg: This is a load-bearing lowercase “c”: if I make it Example_CaptureEthernet, godoc won’t show the example.

pcapgo/capture_test.go
stapelberg: Not sure I follow. The filename has nothing to do with the rest of the code? NewEthernetHandle takes an ifname, which is passed to net.InterfaceByName, and we specify ifname="lo".

stapelberg:
All addressed now. PTAL.

pcapgo/capture_test.go
gconnell: lolwut

gconnell:
Looks like Travis is having some trouble: https://travis-ci.org/google/gopacket/builds/425491900?utm_source=github_status&utm_medium=notification

I believe due to the import of mdlayher/raw in pcapgo.  

I think this would be fixed by adding:
  - go get github.com/google/gopacket/pcapgo
to the set of 'go get' calls in the Travis config?

stapelberg:
Woops, didn’t realize the package names were enumerated in the travis configuration. Added now.

stapelberg:
Thanks!

PR #472, opened 2018-06-21
layers: Add name & getters to LCM

gustafj:
Adds getter for fingerprint of a LCM layer, getter for all Supported
LCM fingerprints and the string representation of the LCM name when
registering a new LCM payload layer.

Co-Authored-By: TacoVox <jonas@derkahler.de>

### DIFF
--- a/layers/lcm.go
+++ b/layers/lcm.go
@@ -43,13 +43,18 @@ type LCM struct {
 	ChannelName string
 	// Gopacket helper fields
 	Fragmented  bool
-	fingerprint uint64
+	fingerprint LCMFingerprint
 	contents    []byte
 	payload     []byte
 }
 
+// LCMFingerprint is the type of a LCM fingerprint.
+type LCMFingerprint uint64
+
 var (
-	lcmLayerTypes  = map[uint64]gopacket.LayerType{}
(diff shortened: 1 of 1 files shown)


TacoVox:
I am obviously fine with this P/R.
I give my consent.

TacoVox:
@gconnell small poke :sunglasses:

layers/lcm.go
```diff
@@ -191,3 +206,8 @@ func (lcm LCM) LayerPayload() []byte {
 func (lcm LCM) Payload() []byte {
 	return lcm.LayerPayload()
 }
+
+// Fingerprint returns the LCM fingerprint of the underlying message.
+func (lcm LCM) Fingerprint() LCMFingerprint {
+	return lcm.fingerprint
```
gconnell: just export (capitalize) this value?  that way folks can set it as well as get it, which would be nice if we ever make this serializable.

gustafj: @gconnell thanks for the response! :sunglasses: 
That would be a larger change, and have effect on the `lcmLayerTypes` map. I agree that serializeable is nice, but that is IMHO out of scope for this change. 
I'd rather just have it as is in this PR, and if / when its made serializeable we can reevaluate.

gconnell:
Sounds good

PR #474, opened 2018-06-25
pcapgo: read.go - SetSnaplen to better control read parameters

assafmo:
This helps to better control what packet size to allow to read and what packet size to dump.

E.g. if a pcap has snaplen of 1500 but has packets of sizes bigger than 1500, [wireshark allows that](https://www.wireshark.org/lists/wireshark-dev/201307/msg00061.html), but pcapgo dumps the packet and reads the next 16 bytes (which are part of the faulty packet payload but pcapgo thinks it's the next header, which will probably also be faulty because it's not really a packet header).  
This can lead to a lot of faulty reads.

Tools like tcpslice, mergecap, tcpdump. pcapfix, etc usually ignore the pcap's snaplen and define their own read snaplen of 262144 or 131072 (e.g. tcpdump [defines max snaplen](https://github.com/the-tcpdump-group/tcpdump/blob/6e80fcdbe9c41366df3fa244ffe4ac8cce2ab597/netdissect.h#L290) and [uses it](https://github.com/the-tcpdump-group/tcpdump/blob/66384fa15b04b47ad08c063d4728df3b9c1c0677/print.c#L343-L358) when it reads a packet).

### DIFF
--- a/pcapgo/read.go
+++ b/pcapgo/read.go
@@ -149,6 +149,11 @@ func (r *Reader) Snaplen() uint32 {
 	return r.snaplen
 }
 
+// SetSnaplen sets the snapshot length of the capture file.
+func (r *Reader) SetSnaplen(newSnaplen uint32) {
+	r.snaplen = newSnaplen
+}
+
 // Reader formater
 func (r *Reader) String() string {
 	return fmt.Sprintf("PcapFile  maj: %x min: %x snaplen: %d linktype: %s", r.versionMajor, r.versionMinor, r.snaplen, r.linkType)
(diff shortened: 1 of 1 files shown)


pcapgo/read.go
```diff
@@ -149,6 +149,11 @@ func (r *Reader) Snaplen() uint32 {
 	return r.snaplen
 }
 
+// SetSnaplen sets the snapshot length of the capture file.
```
gconnell: Could you add a larger comment here discussing why you'd want to override the snaplen that's encoded in the file?  I get it from the pull request description, but I think it's worth specifically mentioning here to justify the existence of this function.

assafmo: You are right. I'll expand on this in the source code as well.

assafmo:
@gconnell I added a comment about `SetSnaplen` existence. :smile:

PR #477, opened 2018-06-26
Adding ModbusTCP decoder

vgraveto:
Hi @florianl 

Here is my pull request that contributes with ModbusTCP layer for the repository.
If any review is needed please let me know.

Regards,
VG

### DIFF
--- a/layers/modbustcp.go
+++ b/layers/modbustcp.go
@@ -26,6 +26,13 @@ const mbapRecordSizeInBytes int = 7
 const modbusPDUMinimumRecordSizeInBytes int = 2
 const modbusPDUMaximumRecordSizeInBytes int = 253
 
+// ModbusProtocol type
+type ModbusProtocol uint16
+
+const (
+	ModbusProtocolModbus ModbusProtocol = 0
+)
+
 //******************************************************************************
 
 // ModbusTCP Type
@@ -37,10 +44,10 @@ const modbusPDUMaximumRecordSizeInBytes int = 253
 type ModbusTCP struct {
 	BaseLayer // Stores the packet bytes and payload (Modbus PDU) bytes .
 
-	TransactionIdentifier uint16 // Identification of a MODBUS Request/Response transaction
-	ProtocolIdentifier    uint16 // 0 = MODBUS protocol
-	Length                uint16 // Number of following bytes (includes 1 byte for UnitIdentifier + Modbus data length
-	UnitIdentifier        uint8  // Identification of a remote slave connected on a serial line or on other buses
+	TransactionIdentifier uint16         // Identification of a MODBUS Request/Response transaction
+	ProtocolIdentifier    ModbusProtocol // It is used for intra-system multiplexing
+	Length                uint16         // Number of following bytes (includes 1 byte for UnitIdentifier + Modbus data length
+	UnitIdentifier        uint8          // Identification of a remote slave connected on a serial line or on other buses
 }
 
 //******************************************************************************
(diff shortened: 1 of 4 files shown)


vgraveto:
I signed CLA!

vgraveto:
I added the GitHub email to the CLA!

vgraveto:
I signed CLA with new email used on GitHub!

vgraveto:
New try on updating CLA

vgraveto:
Sorry for the delay. I hope everything is ok now.

vgraveto:
Hi @florianl 

I think everything is now ok, with my PR. Just let me know if any review is needed.

Regards,
VG

vgraveto:
Hi @gconnell 

Could you please take same time to review this PR
Regards,
VG

layers/modbustcp.go
```diff
@@ -0,0 +1,138 @@
+// Developed by Vitor Graveto
```
gconnell: Please change to "Copyright 2018, The GoPacket Authors, All rights reserved."

and add your name as an "author" under the top-level AUTHORS file.

layers/modbustcp.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+//******************************************************************************
+
+// Payload returns Modbus Protocol Data Unit (PDU) composed by Function Code and Data, it is carried within ModbusTCP packets
+func (d *ModbusTCP) Payload() []byte {
+	return d.BaseLayer.Payload
+}
+
+//******************************************************************************
```
gconnell: this last comment seems spurious?

layers/modbustcp.go
```diff
@@ -0,0 +1,138 @@
...
+	d.Length = binary.BigEndian.Uint16(data[4:6])
+
+	// Length should have the size of the payload plus one byte (size of UnitIdentifier)
+	if d.Length != uint16(len(d.BaseLayer.Payload)+1) {
+		df.SetTruncated()
+		return errors.New("ModbusTCP packet with wrong field value (Length)")
+	}
+	d.UnitIdentifier = uint8(data[6])
+
+	// Return no error.
```
gconnell: comment seems spurious

layers/modbustcp.go
```diff
@@ -0,0 +1,138 @@
...
+// --------
+// Type ModbusTCP implements the DecodingLayer interface. Each ModbusTCP object
+// represents in a structured form the MODBUS Application Protocol header (MBAP) record present as the TCP
+// payload in an ModbusTCP TCP packet.
+//
+type ModbusTCP struct {
+	BaseLayer // Stores the packet bytes and payload (Modbus PDU) bytes .
+
+	TransactionIdentifier uint16 // Identification of a MODBUS Request/Response transaction
+	ProtocolIdentifier    uint16 // 0 = MODBUS protocol
```
gconnell: should this be an enum type?

layers/modbustcp.go
vgraveto: It could simply be remove.
The idea was to say something like "All erros checked and no error detected"
Could also change it to this other sentence, what do you think?

gconnell: In general, more comments are better, but personally I try to go for comments that explain the "why", not the "what".  I'd aim for describing the reasoning behind the code over describing the code itself.  In this case "return no error" specifically describes exactly what the next line is going to do, so I think it's safe to remove.  In general, too, I think Go programmers seeing 'return nil' at the end of a function already get the idea of that meaning "no errors, we're successfully done"

layers/modbustcp.go
vgraveto: It could be an enum, but the documentation has no other possible values for it, see chapter 3.3.1 for MBAP Header description at attached file.

mbap

layers/modbustcp.go
vgraveto: Ok I will remove it

layers/modbustcp.go
vgraveto: On this field some undocumented version may exist and break the decoder if we impose an enum. So if you agree I would leave it this way.

gconnell: Sorry, by "enum" I meant:


```go
type ModbusProtocol uint16
const (
  ModbusProtocolModbus ModbusProtocol = 0
)
```

That way, if we eventually find other protocols, we can add them, but you can always accept any uint16 value.

vgraveto: Ok, good idea I will change it.

gconnell:
Fix lint error, and we should be good to go :)

gconnell:
Thanks for the contribution!

PR #484, opened 2018-07-12
Add ERSPAN Type II header support

tcheever:
Also includes an ERSPAN connection type with basic functions.

### DIFF
--- a/layers/erspan2.go
+++ b/layers/erspan2.go
@@ -1,0 +1,72 @@
+// Copyright 2018 Google, Inc. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"encoding/binary"
+
+	"github.com/google/gopacket"
+)
+
+// ERSPANIIHeader contains all of the fields found in an ERSPAN Type II header
+// https://tools.ietf.org/html/draft-foschiano-erspan-03
+type ERSPANIIHeader struct {
+	IsTruncated                         bool
+	Version, CoS, TrunkEncap            uint8
+	VLANIdentifier, SessionID, Reserved uint16
+	Index                               uint32
+}
+
+const (
+	//ERSPANIIVersionObsolete - The obsolete value for the version field
+	ERSPANIIVersionObsolete = 0x0
+	// ERSPANIIVersion - The current value for the version field
+	ERSPANIIVersion = 0x1
+)
+
+// NewERSPANIIHeader is a convenience "constructor" which sets common default values
+func NewERSPANIIHeader() *ERSPANIIHeader {
+	return &ERSPANIIHeader{
+		Version: ERSPANIIVersion,
+	}
+}
+
+// DecodeFromBytes decodes the given bytes into this layer.
+func (h *ERSPANIIHeader) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	h.Version = data[0] & 0xF0 >> 4
+	h.VLANIdentifier = binary.BigEndian.Uint16(data[:2]) & 0x0FFF
+	h.CoS = data[2] & 0xE0 >> 5
+	h.TrunkEncap = data[2] & 0x18 >> 3
+	h.IsTruncated = data[2]&0x4>>2 != 0
+	h.SessionID = binary.BigEndian.Uint16(data[2:4]) & 0x03FF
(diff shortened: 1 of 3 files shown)


tcheever:
I signed it!

lhausermann:
Dear @tcheever would you double check if you signed the CLA with the right email address ? It should be the one linked to you GitHub account (see the CLA robot message)

tcheever:
@lhausermann Hmm, I do have the correct email address and github username in the CLA.  I did not have a publicly visible email address in my github settings, however, so I've exposed that now.

layers/erspan2.go
```diff
@@ -0,0 +1,173 @@
...
+	"encoding/binary"
+	"net"
+	"time"
+
+	"github.com/google/gopacket"
+)
+
+// Header contains all of the fields found in an ERSPAN2 header
+// https://tools.ietf.org/html/draft-foschiano-erspan-03
+type Header struct {
```
lhausermann: This should be called ERPSANIIHeader

layers/erspan2.go
```diff
@@ -0,0 +1,173 @@
...
+	IsTruncated                         bool
+	Version, CoS, TrunkEncap            uint8
+	VLANIdentifier, SessionID, Reserved uint16
+	Index                               uint32
+}
+
+const (
```
lhausermann: Same here, if the const are related only to ERSPANII It must be the prefix of the const names

layers/erspan2.go
```diff
@@ -0,0 +1,173 @@
...
+		twoByteInt |= 0x400
+	}
+	binary.BigEndian.PutUint16(bytes[2:], twoByteInt)
+
+	fourByteInt := uint32(h.Reserved&0x0FFF)<<20 | h.Index&0x000FFFFF
+	binary.BigEndian.PutUint32(bytes[4:], fourByteInt)
+	return nil
+}
+
+// Conn - A connection to a device to send packets within ERSPAN2 encapsulation
```
lhausermann: I don't think this "Conn" struct + methods should be in a Layer. I guess you are using that in your app code to send packets, but a layer should stick to DecodeFromBytes and SerializeTo methods

layers/erspan2.go
```diff
@@ -0,0 +1,173 @@
...
+	opts   ConnOpts
+}
+
+// ConnOpts - ERSPAN2 connection options
+type ConnOpts struct {
+	MaxMTU uint16
+}
+
+// Dial - Connect to a device given an IP address string.
+func Dial(addr string, opts *ConnOpts) (*Conn, error) {
```
lhausermann: Same as Conn() here

layers/erspan2.go
```diff
@@ -0,0 +1,173 @@
...
+func (e *Conn) IPConn() *net.IPConn {
+	return e.ipConn
+}
+
+// Options provides access to a copy of the connection options
+func (e *Conn) Options() ConnOpts {
+	return e.opts
+}
+
+const (
```
lhausermann: If needed, please use the const already defined in Gopacket other layers (ie IP / Ethernet)

layers/erspan2_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestNewERSPAN2(t *testing.T) {
+	erspan := NewHeader()
+	if erspan.Version != Version {
+		t.Fatalf("Got %v, expected %v\n", erspan.Version, Version)
+	}
+}
+
+// Need root permissions or net_raw_sock linux capabilities for Dial()
+// This test is skipped but here for documentation
+func TestDialWriteClose(t *testing.T) {
```
lhausermann: As for Conn()

layers/erspan2.go
tcheever: Ah yes, good point.  I had the different layers separated into their own packages so callers would have `erspan2.Header` or `erspan2.NewHeader()`.  There are a few things, as you've pointed out, than need to be cleaned up a bit to fit into the `layers` package.

layers/erspan2.go
tcheever: Yeah, had a feeling that would be the feedback but worth a shot!  I'll remove the connection portions.

tcheever:
Ok, I fixed golint on my system.  For some reason the binary wasn't executable anymore even though it had adequate permissions and was of the right format.  Updating golint fixed the issue.

gconnell:
Laurent, quick ping on this:  are you okay with the changes made?

Also, it looks like this isn't currently linked into LayerType stuff.  Could you add a Layer type to layers/layertypes.go?

PR #485, opened 2018-07-13
Add NewBPF func to create BPF from a dead (unassociated) handle

rschmied:
NewBPF allows to create a BPF without requiring an existing handle. This allows to match packets obtained from a-non GoPacket capture source  to be matched.

```golan
bpfi := NewBPF(layers.LinkTypeEthernet, MaxFrameSize, "icmp")
buf := make([]byte, MaxFrameSize)
ci := gopacket.CaptureInfo{}
n, _ := someIO.Read(buf)
ci.CaptureLength = n
ci.Length = n
if bpfi.Matches(ci, buf) {
    doSomething()
}
```

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -695,6 +695,28 @@ func (p *Handle) NewBPF(expr string) (*BPF, error) {
 	return bpf, nil
 }
 
+// NewBPF allows to create a BPF without requiring an existing handle.
+// This allows to match packets obtained from a-non GoPacket capture source
+// to be matched.
+//
+// buf := make([]byte, MaxFrameSize)
+// bpfi, _ := pcap.NewBPF(layers.LinkTypeEthernet, MaxFrameSize, "icmp")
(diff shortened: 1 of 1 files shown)


lhausermann:
Dear @rschmied 

Thanks for your contribution. It's true that BPF filtering is useful nevermind the source of the packet
I would like to see some code cleaning before merging it

1- I don't think you need a 'merge' commit in a PR . Please remove it.
2- I see you made other changes related to PCAP warnings.  I would suggest to move that part into a second PR.

Best

rschmied:
thanks @lhausermann did as you proposed. Regarding 2), I'm not sure whether there was / is actually something worth persisting. The only valid relevant change from my end was the pcap.go / NewBPF() func.

rschmied:
Hmm, you are not referring to PR #384 are you?

lhausermann:
I did not figure out that it was already a separate PR such as PR #384

lhausermann:
Thanks !

Issue #487, opened 2018-07-18
reassembly: Flush might close half connection and cause data loss.


noibar:
If assembler.FlushWithOptions is called, and one of the halfConnection is seen only before the specified time, half of the connection is closed. If another packet that belongs to this halfConnection is received, the stream pool returns the main connection, with half connection closed, and the data is discarded without and error or log message. 

```go:
 func (a *Assembler) AssembleWithContext(netFlow gopacket.Flow, t *layers.TCP, ac AssemblerContext) {
    // ...

    if half.closed {
        return
    }

    // ...
}
```

lhausermann:
Great @noibar , second bug fixed into @sentryo contribution (defrag and reassembly)
Hey, we are hiring by the way ;-)

PR #488, opened 2018-07-18
Fix issue #487, close connection when two halfs should be closed.

noibar:
fix issue: 
https://github.com/google/gopacket/issues/487


gconnell:
Nice change, thanks!

PR #489, opened 2018-07-21
Fix for zeroing memory returned from PrependBytes()

greigbannister:
PrependBytes() expects the user to overwrite all bytes.  This fix
ensures that all bits in all bytes are correctly overwritten


gconnell:
Thanks for the fix!  Happy to merge once CLA is signed.

greigbannister:
No worries.  Just got notification that the CLA has been completed, so we should be good to go.

Issue #483, opened 2018-07-09
reassembly 


smithjacobj:
Currently there appears to be an issue on CPU saturation where a call to `reassemblyObject.CaptureInfo` results in an invalid access in `page.CaptureInfo`. Details are slim at the moment since I just started investigating this, but I have a couple backtraces (with the project name redacted)

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x8ee246]

goroutine 245 [running]:
github.com/?/main/vendor/github.com/google/gopacket/reassembly.(*page).captureInfo(0xc4222f2f40, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/?/main/vendor/github.com/google/gopacket/reassembly/tcpassembly.go:242 +0x56
github.com/?/main/vendor/github.com/google/gopacket/reassembly.(*reassemblyObject).CaptureInfo(0xc42017d0a0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/?/main/vendor/github.com/google/gopacket/reassembly/tcpassembly.go:162 +0xcc
github.com/?/main/agent/evengine/pcapture.(*tcpStream).ReassembledSG(0xc426edca00, 0x14979e0, 0xc42017d0a0, 0x148a6a0, 0xc422e14270)
        /go/src/github.com/?/main/agent/evengine/pcapture/pcapture.go:160 +0x110
github.com/?/main/vendor/github.com/google/gopacket/reassembly.(*Assembler).sendToConnection(0xc42017d000, 0xc4271b3508, 0xc4271b3578, 0x148a6a0, 0xc422e14270, 0x8857eefe)
        /go/src/github.com/?/main/vendor/github.com/google/gopacket/reassembly/tcpassembly.go:1080 +0xab
github.com/?/main/vendor/github.com/google/gopacket/reassembly.(*Assembler).skipFlush(0xc42017d000, 0xc4271b3508, 0xc4271b3578)
        /go/src/github.com/?/main/vendor/github.com/google/gopacket/reassembly/tcpassembly.go:1166 +0xbd
[...]

PR #492, opened 2018-07-23
Jds reassembly nil

stinkyfingers:
Addresses #483 -  Currently, 
- `NewAssembler` initializes `Assembler.ret as an array `[16]byteContainer (interface)`. 
- In `buildSG()`, the `Assembler.ret` array is assigned to `Assembler.cacheSG.all`. 
- `Assembler.cacheSG` is passed to` half.stream.ReassembledSG()`int `sendToConnection()`. 
- We're occasionally seeing nil pointer panics on `Assembler.cacheSG.CaptureInfo()`, if we call it from our implementation of `ReassembledSG()`. 

It seems like, if the Assembler's cacheSG array doesn't get filled, there can be nils in it. This PR inits the Assembler with a slice of `byteContainer`.


gconnell:
Hey, thanks for tracking this down!  A couple of quick requests:

1) CLA please :)
2) could we change from `ret = []byteContainer{}` to `ret = make([]byteContainer, 0, initialSize)`
3) looks like this PR has a couple spurious commits in it (code/revert/etc).  Would you mind squashing things to a single commit?

Issue #493, opened 2018-07-25
reassembly: accept data that was acked but not flushed yet.


noibar:
the reassembly.Assembler calls stream.Accept() with last accepted ack number. Later on, the ack number might be in use in the TCPOptionCheck.Accept() in order to identify re-emitted data. In the current implementation, if ack packet is received before the data, the data will be ignored even though it wasn't assembled yet. 
An ack packet might be received before the data because incoming port don't always keep the packets order.

PR #494, opened 2018-07-25
Reassembly data after ack

noibar:
Solves #493


gconnell:
Awesome, thanks for the fix!

PR #502, opened 2018-08-23
Adding ReaderStream.Seen()

martin-osborne:
Hi,

I'm trying to get timing information from PCAP files for an application layer protocol that runs over TCP.  I have not been able to work out how to do this without the modifications in this pull request.  Apologies if there is an easier approach that I've missed.

The change simply adds a method to access the `Seen` field of the current `Reassembly`.  Tests are also provided.

Thank you for taking the time to look at this.

Martin.

### DIFF
--- a/tcpassembly/tcpreader/reader.go
+++ b/tcpassembly/tcpreader/reader.go
@@ -46,8 +46,10 @@ package tcpreader
 
 import (
 	"errors"
-	"github.com/google/gopacket/tcpassembly"
 	"io"
+	"time"
+
+	"github.com/google/gopacket/tcpassembly"
 )
 
 var discardBuffer = make([]byte, 4096)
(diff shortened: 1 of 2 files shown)


gconnell:
Hey, Martin,

Thanks for the code!  Once the CLA's submitted, I'm happy to merge this in.

That said, a few things you might want to know:
1) gopacket/tcpassembly is mostly deprecated, with gopacket/reassembly taking over and more actively maintained
2) gopacket/tcpassembly/tcpreader is generally more for things that care just about getting an io.Reader (IE: they don't need timing), so I'd expect that those caring about timing would just use tcpassembly directly
3) note that there's a possibility for .Seen times to be wildly off.  For example, given a request that comes from 3 packets, we could get packets 1 and 3 at time T, but not get packet 2 until time T+1.  If you asked "what's the timestamp of the last thing we read" after reading 1/2/3, you'd get time T, not T+1, if I remember correctly.

martin-osborne:
Hi,

Thanks for responding.  `gopacket/reassembly` seems to give me everything I need.  Many thanks for the guidance.

Thanks,

Martin.

PR #505, opened 2018-08-30
Convert LLC and SNAP to DecodingLayer interface

notti:
This moves code from the decodeX functions to DecodeFromBytes for both LLC and SNAP and changes decodeX to use DecodeFromBytes. The only behavioral change I introduced is error handling in case the passed byte-slice is too short.

Since I touched the file, I also added a second commit that removes the underscores in LLC.SerializeTo from the variable names to make the linter happy - I can make a separate pull request for this if you like.


notti:
Oh from the travis log I just saw, that llc.go is on the linter blacklist - after this it should be possible to take it off this blacklist.

PR #506, opened 2018-08-30
chg: implement CanDecode() for ICMPv6 messages

rschmied:
this adds the `CanDecode()` interface to the implemented ICMPv6 messages to `layers/icmp6msg.go` which will then allow to build a packet data parser with those specific ICMPv6 messages.


gconnell:
Thanks!

Issue #517, opened 2018-09-14
Wrong Serialization of an IPv6HopByHop packet


cimnine:
I'm experiencing an error when serializing IPv6HopByHop packets.

The problem, to be exact, is that the HopByHop header is written twice to the buffer.

It's very easy to replicate this problem:

```go
// in icmp6hopbyhop_test.go
func TestPacketICMPv6WithHopByHop(t *testing.T) {
  // add this after the original code
  checkSerialization(p, t)
}

// I've added the checkSerialization() function to `base_test.go`

// Checks that when a serialized version of p is decoded, p and the serialized version of p are the same.
// Does not work for packets where the order of options can change, like icmpv6, dhcpv6, etc.
func checkSerialization(p gopacket.Packet, t *testing.T) {
	buf := gopacket.NewSerializeBuffer()
	opts := gopacket.SerializeOptions{
		ComputeChecksums: false,
		FixLengths: false,
	}
	if err := gopacket.SerializePacket(buf, opts, p); err != nil {
		t.Error("Failed to encode packet:", err)
	}

	p2 := gopacket.NewPacket(buf.Bytes(), LinkTypeEthernet, gopacket.Default)
	if p2.ErrorLayer() != nil {
		t.Error("Failed to decode the re-encoded packet:", p2.ErrorLayer().Error())
	}

	if p2.Dump() != p.Dump() {
		t.Errorf("The decoded and the re-encoded packet are different!\nDecoded:\n%s\n Re-Encoded:\n%s", p.Dump(), p2.Dump())
	}
}
```

The resulting packet looks like this:

        -- FULL PACKET DATA (198 bytes) ------------------------------------
        00000000  33 33 00 00 00 16 1e c3  e3 b7 c4 d5 86 dd 60 00  |33............`.|
        00000010  00 00 00 88 00 01 00 00  00 00 00 00 00 00 00 00  |................|
        00000020  00 00 00 00 00 00 ff 02  00 00 00 00 00 00 00 00  |................|
        00000030  00 00 00 00 00 16 3a 00  05 02 00 00 01 00 3a 00  |......:.......:.|
        00000040  05 02 00 00 01 00 8f 00  9e ed 00 00 00 06 03 00  |................|
        00000050  00 00 ff 02 00 00 00 00  00 00 00 00 00 01 ff b7  |................|
        00000060  c4 d5 03 00 00 00 ff 02  00 00 00 00 00 00 00 00  |................|
[...]

PR #522, opened 2018-09-20
Adds MLDv1 and MLDv2 (Multicast Listener Discovery)

cimnine:
This adds code to support decoding and encoding of MLDv1 and MLDv2 packets.

The implementation follows [rfc2710](https://tools.ietf.org/html/rfc2710) for MLDv1 and [rfc3810](https://tools.ietf.org/html/rfc3810) for MLDv2.

There are some helper functions where they make sense for implementing the protocol, e.g. when there is a reference in the RFC.

I've commented out some test code regarding serialization that would currently not pass due to #517, which #530 would fix.


cimnine:
Closed in favour of #533

PR #524, opened 2018-09-22
fix afpacket revents still set to 0

safchain:
For example if an interface being captured is deleted
poll should return an error but if not catched
pollForFirstPacket will remain in a infinite loop
as poll won't return an error.


safchain:
Still investigating

safchain:
With this fix the revents field is set properly when a captured interface is deleted

PR #530, opened 2018-09-28
Serialized Layer Introspection [Fixes #517]

cimnine:
This PR adds a way for the current layer to see all the layers that were successfully serialized before the current layer, during the serialization of the current layer.

I've come up with this solution in an attempt to fix #517 in a clean way, i.e. without inspecting the actual buffer. I believe the method I've chosen can be useful for other serialization implementations, like such that have the same challenges as described in #517. (tl;dr: Information is present in the "main layer" but also in a separate layer, but must not be serialized twice).

It's important to understand that the solution, as implemented, always prefers the information present in the extra layer over the information present in the "main layer". E.g. in #517, the HopByHop information of the IPv6 layer is ignored when an IPv6HopByHop layer was serialized before.

Bonus: I've cleaned up the `golint` errors in `ip6.go`. As a reviewer you might want to review this PR commit by commit.

PR #533, opened 2018-10-01
Adds MLDv1 and MLDv2 (Multicast Listener Discovery) [rebased onto #530]

### LINKED ISSUE #517: Wrong Serialization of an IPv6HopByHop packet

cimnine:
This adds code to support decoding and encoding of MLDv1 and MLDv2 packets.

The implementation follows [RFC 2710](https://tools.ietf.org/html/rfc2710) for MLDv1 and [RFC 3810](https://tools.ietf.org/html/rfc3810) for MLDv2.

There are some helper functions where they make sense for implementing the protocol, e.g. when there is a reference in the RFC.

This is #522 rebased onto the solution proposed in #530 (which fixes the bug described in #517).

Fixes #517 
Closes #522 
Closes #530


layers/mldv1.go
```diff
@@ -0,0 +1,174 @@
+// Copyright 2018 Nine Internet Solutions AG. All rights reserved.
```
gconnell: Would you be okay to modify this to say "Copyright 2018 GoPacket Authors" and put your (and/or your company's) name in the top-level AUTHORS file?

layers/mldv1.go
```diff
@@ -0,0 +1,174 @@
...
+
+// DecodeFromBytes decodes the given bytes into this layer.
+func (m *MLDv1Message) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	if len(data) < 20 {
+		df.SetTruncated()
+		return errors.New("ICMP layer less than 20 bytes for Multicast Listener Query Message V1")
+	}
+
+	m.MaximumResponseDelay = time.Duration(binary.BigEndian.Uint16(data[0:2])) * time.Millisecond
+	m.MulticastAddress = data[4:20]
```
gconnell: looks like data[2] and data[3] are ignored?

layers/mldv1.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	if m.MaximumResponseDelay < 0 {
+		return errors.New("maximum response delay must not be negative")
+	}
+	dms := m.MaximumResponseDelay / time.Millisecond
+	if dms > math.MaxUint16 {
+		return fmt.Errorf("maximum response delay %dms is more than the allowed 65535ms", dms)
+	}
+	binary.BigEndian.PutUint16(buf[0:4], uint16(dms))
```
gconnell: note that PutUint16 will never touch buf[2] or buf[3].  If you expect them to be zero, you may want to specifically set them as zero.

layers/mldv1.go
```diff
@@ -0,0 +1,174 @@
...
+
+	copy(buf[4:], m.MulticastAddress)
```
gconnell: maybe specifically say buf[4:20], in case of the (erroneous, but possible) error where MulticastAddress is longer than 16 bytes?

gconnell: And/or check that `len(MulticastAddress) == 16` and return an error if it doesn't

layers/mldv1.go
```diff
@@ -0,0 +1,174 @@
...
+	return nil
+}
+
+// Sums this layer up nicely formatted
+func (m *MLDv1Message) String() string {
+	return fmt.Sprintf(
```
gconnell: have you looked at calling `gopacket.LayerString(m)` ?  I think it should work well here.  Note that that's what Packet.String() does by default.

layers/mldv1_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright 2018 Nine Internet Solutions AG. All rights reserved.
```
gconnell: ditto here

layers/mldv2_test.go
```diff
@@ -0,0 +1,134 @@
+// Copyright 2018 Nine Internet Solutions AG. All rights reserved.
```
gconnell: ditto here

layers/mldv2.go
```diff
@@ -0,0 +1,600 @@
+// Copyright 2018 Nine Internet Solutions AG. All rights reserved.
```
gconnell: ditto here

layers/mldv2.go
```diff
@@ -0,0 +1,600 @@
...
+	"math"
+	"net"
+	"time"
+
+	"github.com/google/gopacket"
+)
+
+const (
+	// S Flag bit is 1
+	sTrue uint8 = 0x8
```
gconnell: mind prefixing these with their type?  layers/ is a very large, single global namespace (mostly because I'm bad at planning ahead...)

```go
mldv2STrue = 0x8
mldv2SMask = 0x8
mldv2QRVMask = 0x7
```

layers/mldv2.go
```diff
@@ -0,0 +1,600 @@
...
+
+		copy(buf[0:], m.SourceAddresses[i].To16())
+	}
+
+	return nil
+}
+
+// Sums this layer up nicely formatted
+func (m *MLDv2MulticastListenerQueryMessage) String() string {
+	return fmt.Sprintf(
```
gconnell: ditto on gopacket.LayerString (or gopacket.LayerDump)

layers/mldv2.go
```diff
@@ -0,0 +1,600 @@
...
+func (*MLDv2MulticastListenerQueryMessage) LayerType() gopacket.LayerType {
+	return LayerTypeMLDv2MulticastListenerQuery
+}
+
+// CanDecode returns the set of layer types that this DecodingLayer can decode.
+func (*MLDv2MulticastListenerQueryMessage) CanDecode() gopacket.LayerClass {
+	return LayerTypeMLDv2MulticastListenerQuery
+}
+
+// QQI calculates the QQI based on the QQIC
```
gconnell: ah, that clears things up :P

layers/mldv2.go
```diff
@@ -0,0 +1,600 @@
...
+	if err != nil {
+		return err
+	}
+
+	copy(buf[0:2], []byte{0x0, 0x0})
+	binary.BigEndian.PutUint16(buf[2:4], m.NumberOfMulticastAddressRecords)
+	return nil
+}
+
+// Sums this layer up nicely formatted
```
gconnell: function docstring should start with function name

```go
String sums ...
```

cimnine:
Thanks for the feedback. I've rebased the original branch before I've seen your feedback. I'll see, that I can implement your feedback.

layers/mldv1.go
cimnine: in the `AUTHOR` file, what are 'Authors' and what are 'Contributors'?

gconnell: Somewhat subjective, but I'd say "author" is more than 25 lines of code.  So you're an author ;)

layers/mldv2.go
cimnine: As I'm calling functions (for a nicer representation of some values for humans) the convenience function, to me, does not seem to be the right fit here.

layers/mldv1.go
cimnine: I would like to leave the implementation as-is here, because the representation does the conversion of the `MaximumResponseDelay` into milliseconds, the unit the RFC refers to for this field.

layers/mldv1.go
cimnine: I've made this explicit:

https://github.com/google/gopacket/pull/533/files#diff-0736ecd395163e25dc6d8d220cd145e1R39

cimnine:
This PR is ready for another review. Your recommendations where very valuable and I've addressed to most of them. I left a comment where I disagree. The branch is rebased onto the latest master (8b9dff4 as of this writing) and should therefore be ready to merge.

(The build also went through, but the webhook to notify Github apparently didn't.)

gconnell:
Thanks for the code!

PR #534, opened 2018-10-04
Partial TLS Support

jselvi:
I implemented a piece of TLS protocol since I needed it for a tool that I'm coding at this moment.
I have been waiting to have enough spare time to implement the whole protocol, but that could take many months, so I decided to contribute what I have at this moment. Maybe other people could do the same and implement what they need, and we will finally implement the entire protocol all together.

I have tried to mimic your coding style, so it should fit your code with minor changes.


layers/layertypes.go
```diff
@@ -11,6 +11,8 @@ import (
 )
 
 var (
+	// Pending to set the proper number
+	LayerTypeTLS                          = gopacket.RegisterLayerType(1000, gopacket.LayerTypeMetadata{Name: "TLS", Decoder: gopacket.DecodeFunc(decodeTLS)})
```
gconnell: Just throw this at the bottom of the list, incrementing the top-most current number by one :)

layers/tcp.go
```diff
@@ -308,11 +308,7 @@ func decodeTCP(data []byte, p gopacket.PacketBuilder) error {
 	if err != nil {
 		return err
 	}
-	if p.DecodeOptions().DecodeStreamsAsDatagrams {
```
gconnell: Please re-add this if statement.  It allows users to decide whether they want their packets decoded beyond the TCP/UDP layer or not.  This is important in cases where, for example, someone is using port 994 for something other than ircs.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
+// Copyright 2012 Google, Inc. All rights reserved.
+// Copyright 2009-2011 Andreas Krennmair. All rights reserved.
+// TLS decoder is a contribution from Jose Selvi
```
gconnell: Rather than these 3 lines, just add yourself as an author in the toplevel AUTHORS file, and say

// Copyright 2018 The GoPacket Authors.  All rights reserved.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+
+	"github.com/google/gopacket"
+)
+
+// TLSType defines the type of data after the TLS Record
+type TLSType uint8
+
+// TLSType known values.
+const (
+	TLSchange_cipher_spec TLSType = 20
```
gconnell: Use CamelCase everywhere, and capitalize the first letter after the acronym.

TLSChangeCipherSpec
TLSAlert
TLSHandshake
...

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+	TLSalert              TLSType = 21
+	TLShandshake          TLSType = 22
+	TLSapplication_data   TLSType = 23
+	TLSunknown            TLSType = 255
+)
+
+func (tt TLSType) String() string {
```
gconnell: Please add a (brief) comment on all exported functions and structs, in typical godoc format (first word is name of the thing).  It can be as simple as:

// String implements fmt.Stringer.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+	case TLSalert:
+		return "Alert"
+	case TLShandshake:
+		return "Handshake"
+	case TLSapplication_data:
+		return "Application Data"
+	}
+}
+
+type TLSVersion struct {
```
gconnell: How about:
```go
type TLSVersion uint16

func (tv TLSVersion) Major() uint8 {
  return uint8(tv >> 8)
}
func (tv TLSVersion) Minor() uint8 {
  return uint8(tv)
}
```
This way, you can pull it from the packet with:

`Version = TLSVersion(binary.BigEndian.Uint16(data[x:y]))`

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+	major uint8
+	minor uint8
+}
+
+func (tv TLSVersion) String() string {
+	s := tv.major*10 + tv.minor
+	switch s {
```
gconnell: again, with TLSVersion as a uint16, you could say:

```
case 0x0200: SSL2.0
case 0x0300: SSL3.0
case 0x0301: TLS1.0
case 0x0302: TLS1.1
...
```

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+	BaseLayer
+
+	// TLS Records
+	ChangeCipherSpec []TLSchangeCipherSpecRecord
+	Handshake        []TLShandshakeRecord
+	Appdata          []TLSappdataRecord
+	Alert            []TLSalertRecord
+}
+
+type TLSrecordHeader struct {
```
gconnell: TLSRecordHeader

Capitalize first letter after 'TLS'

gconnell: ditto for TLSHandshakeRecord, TLSAppDataRecord, TLSAlertRecord, etc...

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// Payload returns nil.
+func (t *TLS) Payload() []byte {
+	return []byte{}
+}
+
+// SerializeTo writes the serialized form of this layer into the
+// SerializationBuffer, implementing gopacket.SerializableLayer.
+func (t *TLS) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
```
gconnell: If serialization isn't supported, just don't add this function, so it doesn't (falsely) implement the serializablelayer interface.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+	return LayerTypeTLS
+}
+
+// NextLayerType implements gopacket.DecodingLayer.
+func (t *TLS) NextLayerType() gopacket.LayerType {
+	return gopacket.LayerTypePayload
+}
+
+// Payload returns nil.
+func (t *TLS) Payload() []byte {
```
gconnell: You shouldn't need to actually implement this, since BaseLayer already does.  Just don't set BaseLayer.Payload to anything, and it'll return nil.

gconnell: Also, []byte{} does a malloc, I think, so returning nil here would be better all around.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// CanDecode implements gopacket.DecodingLayer.
+func (t *TLS) CanDecode() gopacket.LayerClass {
+	return LayerTypeTLS
+}
+
+// NextLayerType implements gopacket.DecodingLayer.
+func (t *TLS) NextLayerType() gopacket.LayerType {
+	return gopacket.LayerTypePayload
```
gconnell: Probably use LayerTypeZero.

layers/tls.go
```diff
@@ -0,0 +1,205 @@
...
+		if e != nil {
+			return e
+		}
+		t.Appdata = append(t.Appdata, r)
+	}
+
+	if len(data) == tl {
+		return nil
+	}
+	return t.DecodeFromBytes(data[tl:len(data)], df)
```
gconnell: Rather than automatically call DecodeFromBytes on the next layer, instead just return nil, and return the next layer type with the NextLayerType() function.

Something like:

```go
  t.BaseLayer.Payload = data[tl:]
  return nil
}

func (t *TLS) NextLayerType() LayerType {
  if len(t.Payload) == 0 { return LayerTypeZero }
  return LayerTypeTLS
}
```

layers/tls_alert.go
```diff
@@ -0,0 +1,160 @@
+// Copyright 2012 Google, Inc. All rights reserved.
```
gconnell: see previous comment on simple copyright

layers/tls_alert.go
```diff
@@ -0,0 +1,160 @@
...
+type TLSalertLevel uint8
+type TLSalertDescr uint8
+
+const (
+	TLSalert_warning       TLSalertLevel = 1
+	TLSalert_fatal         TLSalertLevel = 2
+	TLSalert_unknown_level TLSalertLevel = 255
+
+	TLSalert_close_notify                TLSalertDescr = 0
+	TLSalert_unexpected_message          TLSalertDescr = 10
```
gconnell: TLSFooBarBaz instead of TLSfoo_bar_baz everywhere.

layers/tls_alert.go
```diff
@@ -0,0 +1,160 @@
...
+		t.EncryptedMsg = data
+	}
+
+	return nil
+}
+
+func (al TLSalertLevel) String() string {
+	switch al {
+	default:
+		return "Unknown"
```
gconnell: when returning unknown values, maybe add the numeric value for debugging purposes?

```
  return fmt.Sprintf("Unknown(%d)", al)

File: layers/tls_alert.go [hunk 16]
```diff
@@ -0,0 +1,160 @@
...
+		return "Fatal"
+	}
+}
+
+func (ad TLSalertDescr) String() string {
+	switch ad {
+	default:
+		return "Unknown"
+	case TLSalert_close_notify:
+		return "close_notify"
```
gconnell: note on naming:  feel free to keep the returned strings as "lower_with_under", just variable names need to be UpperCamel

layers/tls_appdata.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/google/gopacket"
+)
+
+type TLSappdataRecord struct {
+	TLSrecordHeader
+	Payload []byte
+}
+
+// DecodeFromBytes decodes the slice into the TLS struct.
+func (t *TLSappdataRecord) DecodeFromBytes(h TLSrecordHeader, data []byte, df gopacket.DecodeFeedback) error {
```
gconnell: since TLSappdataRecord isn't a gopacket layer, probably don't export this function (call it decodeFromBytes instead of DecodeFromBytes)

Ditto for other things that aren't layer types themselves.

layers/tls.go
gconnell: Actually, thinking about this more, I think there might be a bit of a redesign needed here.  I'd have to go look at the spec, but from your code I'm guessing that it's possible/likely for TLS to put multiple values together, IE: <Eth/IP/TCP/TLS1/TLS2/TLS3>, where TLSX are all LayerTypeTLS.

This doesn't play nicely with DecodingLayerParser, since it populates the structs it's given.  If it's given a TLS struct, it'll populate it with TLS1, then overwrite it with TLS2 then TLS3.

As an alternative, I'm wondering if it'd be better to have the TLS layer contain a repeated set of your current TLS values.  Something like this:

```go
type TLS struct {  // LayerTypeTLS
  Records []TLSRecord
}

type TLSRecord struct {  // move all your current 'TLS struct' stuff here
}
```

This way, if you pass a &TLS to DecodingLayerParser, it'll happily add as many TLSRecords as it sees in a single packet here.  To avoid having to malloc a bunch of TLSRecords, here's a trick we've used a few times:

```go
type TLS struct {
  Records []TLSRecord
  buf [2]TLSRecord
}

func (t *TLS) DecodeFromBytes(...) {
  if t.Records == nil {
    t.Records = t.buf[:0]
  }
  // now, populate with:
  for each record {
    var record TLSRecord
    ... record.foo = bar ...
    t.Records = append(t.Records, record)
  }
}
```

layers/tls_test.go
```diff
@@ -0,0 +1,243 @@
...
+		t.Error("Failed to decode packet:", p.ErrorLayer().Error())
+	}
+	checkLayers(p, []gopacket.LayerType{LayerTypeTLS}, t)
+
+	r := p.Layer(LayerTypeTLS).(*TLS).Appdata
+	if len(r) != 2 {
+		t.Errorf("Wrong number of application data records, expected 2, got %d", len(r))
+	}
+
+	d1 := r[0].Payload
```
gconnell: consider doing a full DeepEqual here (search for DeepEqual in some other tests for an example)

layers/tls.go
jselvi: @gconnell Do you mean that if somebody encapsulates TCP packets inside TCP Payload (2 TCP headers), the first one would be hidden for gopacket?

You are right, TLS allows (and It's easy to see) several TLS records in the same packet. That's why I was processing a TLS layer and then continue with the remaining bytes. I think that was working in my tests but, to be honest, I'm not sure I'm getting date from the first layer once I have processed the following layers. I'll double check it.

jselvi: Actually, as far as I remember from TLS specification, TLS layer is actually a set of TLS Records.
My first approach was as you said, creating a slice of TLSRecords.

The point is that there are several kinds of records and I didn't find a way to use an interface (or similar) here, so the only solution that I found was to create a TLSChangeCipherSpecRecord, TLSHandshakeRecord, etc inside each TLSRecord.

Finally, I decided to use four slices (one per record type). I think it's the most efficient option, since I have an empty slice for those records that weren't found in the packet. My only concern here was that I'm losing which record was first, second, etc, but I think it doesn't really matter.

Do you really think a redesign is necessary here? I feel pretty comfortable with this approach.

jselvi:
@gconnell I think I fixed everything except your redesign proposal, since I don't really understand why we need such redesign. From my side I think it's ready to merge. If you think that redesign is important I'm open to do it but, to be honest, I can't see any advantage on doing it. Could we talk about it?

layers/tls.go
gconnell: Your use of 4 slices seems like a fine choice to me.  I think after one minor refactor to the current implementation we should be ready to merge.

The important thing about implementing a `DecodeFromBytes()` is that, when called on a new set of bytes, it totally resets the current layer to that set of bytes.  The fact that the current one just appends instead of increments is an erroneous implementation, but an easily fixable one.  Let's try renaming the current recursive `DecodeFromBytes` to `decodeSingleRecord`, then make a new `DecodeFromBytes` implementation that does the following:

```go
func (t *TLS) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
  t.BaseLayer.Contents = data  // Contents should contain all bytes in this layer
  t.BaseLayer.Payload = nil
  // Reset all of the slices by discarding their values, but keep the actual slices around so the next calls to append don't need to malloc
  t.ChangeCipherSpec = t.ChangeCipherSpec[:0]
  t.Handshake = t.Handshake[:0]
  t.Appdata = t.Appdata[:0]
  t.Alert = t.Alert[:0]
  // Now recursively call decodeSingleRecord (the old DecodeFromBytes) and return its error
  return t.decodeSingleRecord(data)
}
```

GoPacket has two main ways of decoding layers.  The first, via creating a Packet, mallocs a new layer at every call, via the registered decoder (that's your already-quite-nice `decodeTLS` call, which you'll notice does a malloc on line 107).  Since it's creating a new TLS{} struct, all of the slices will already be empty, which you'd like.

The second method is to create a single struct once, then call `DecodeFromBytes` on it once for each packet.  This is done within the DecodingLayerParser, but consider this case, as it's functionally equivalent:

```go
var tls TLS
var bytes1 = []byte{...}  // Contains 3 Appdata records
tls.DecodeFromBytes(bytes1, ...)
// len(tls.Appdata) == 3
var bytes2 = []byte{...} // Contains 2 Appdata records
tls.DecodeFromBytes(bytes2, ...)
// len(tls.Appdata) == 2, the original 3 have been cleared
```

With the current impl, after decoding `bytes2`, `len(tls.Appdata)` would erroneously be 5.  The new impl just clears things up slightly so that `DecodeFromBytes` starts fresh each run.

You'll note the use of `[:0]` to clear slices above.  That's because one of the things a `DecodeFromBytes` should try to do as little as possible is malloc new values.  In effect, consider that a single `TLS` is having `DecodeFromBytes` called on it many times in a tight loop.  On aggregate, we'd like it to malloc new values zero times per call.  It's okay if a few mallocs happen initially (and they will here, as the first set of slices is grown).  But since that set of slices is staying populated (its capacity isn't decreasing, just its length), future calls can use the prior ones' mallocs.

Thanks again for your hard work on this!  I think with just that one small change, we should be ready to go here.

layers/tls.go
```diff
@@ -0,0 +1,196 @@
...
+//  +--+--+--+--+--+--+--+--+
+
+// TLS is actually a slide of TLSrecord structures
+type TLS struct {
+	BaseLayer
+
+	// TLS Records
+	ChangeCipherSpec []TLSChangeCipherSpecRecord
+	Handshake        []TLSHandshakeRecord
+	Appdata          []TLSAppDataRecord
```
gconnell: minor nit, but could this be AppData to match the TLSAppDataRecord?

layers/tls_test.go
```diff
@@ -0,0 +1,318 @@
...
+	0x51, 0x52, 0x53, 0x54, 0x55, 0x56, 0x57, 0x58, 0x59, 0x5a, 0x5b, 0x5c, 0x5d, 0x5e, 0x5f,
+	0xd1, 0xd2, 0xd3, 0xd4, 0xd5, 0xd6, 0xd7, 0xd8, 0xd9, 0xda, 0xdb, 0xdc, 0xdd, 0xde, 0xdf,
+}
+
+var testTLSDecodeOptions = gopacket.DecodeOptions{
+	SkipDecodeRecovery:       true,
+	DecodeStreamsAsDatagrams: true,
+}
+
+func TestParseTLSClientHello(t *testing.T) {
```
gconnell: You can add a new small test here if you want to make sure the right behavior for the previously described `DecodeFromBytes` is followed:

```go
func TestTLSClientHelloDecodeFromBytes(t *testing.T) {
  var tls TLS
  if err := tls.DecodeFromBytes(tlsClientHello, ...); err != nil {
    t.Fatalf("failed first decode: %v", err)
  }
  if !reflect.DeepEquals(tls, clientHelloDecoded) {
    t.Fatalf("first decode doesn't match")
  }
  if err := tls.DecodeFromBytes(tlsClientHello, ...); err != nil {
    t.Fatalf("failed second decode: %v", err)
  }
  if !reflect.DeepEquals(tls, clientHelloDecoded) {
    t.Fatalf("second decode doesn't match")
  }
}
```

jselvi:
@gconnell Thanks a lot! Now I understand much better why the change is required.
It should be fine now.

gconnell:
Thanks a ton for the help on GoPacket!

jselvi:
Happy to contribute! 👍

layers/tls.go
```diff
@@ -0,0 +1,208 @@
...
+
+func (t *TLS) decodeTLSRecords(data []byte, df gopacket.DecodeFeedback) error {
+	if len(data) < 5 {
+		df.SetTruncated()
+		return errors.New("TLS record too short")
+	}
+
+	// since there are no further layers, the baselayer's content is
+	// pointing to this layer
+	t.BaseLayer = BaseLayer{Contents: data[:len(data)]}
```
jensenhwa, line 138: Hi all, sorry to resurrect this old PR thread, but I've been attempting to use this code here and am wondering why we include this line. Shouldn't the BaseLayer's contents encompass the entirety of the TLS layer, not just the last decoded record as it does here? I would think this line could be completely removed, given L119, but not sure if @jselvi had something else in mind.

Would really appreciate hearing your reasoning behind this line, otherwise I'm tempted to suggest its removal. Thanks in advance.

jselvi: I coded this years ago, so not sure about what I had in mind :)

But yeah, it seems like you said. I remember I started coding a dumb decoder, based on others that were already coded, and I started modifying it to add the features that I needed. I guess I didn't realize that this line didn't make sense at the time I made this function recursive.

I think tests had a pretty good coverage for this decoder, so I would just try to remove the line and run them. If they are all green, I think we can keep this line removed.

PR #537, opened 2018-10-04
Fix poll in afpacket. This fixes #529

### LINKED ISSUE #529: afpacket unix.Poll doesn't report revents which lead in infinite loop
Since this commit: https://github.com/google/gopacket/commit/5527bfcadcaabe78fa8db887e997d3a71171d425

afpacket makes use of unix.Poll instead of C.poll. It seems that this change introduced a bug. 
The Revents field of pollSet struct is never set, meaning errors are not anymore reported.

This can be tested like this : 

1. creating a veth pair
```
ip link add veth0 type veth peer name veth1
ip l set veth1 up
ip l set veth0 up
```

2. starting capture
```
package main

import (
        "fmt"

        "github.com/google/gopacket/afpacket"
)

func main() {

        tpacket, _ := afpacket.NewTPacket(
                afpacket.OptInterface("veth0"),
        )

        fmt.Println("Start capture")
        for {
                if _, _, err := tpacket.ReadPacketData(); err != nil {
                        panic(err)
                }
        }
        fmt.Println("Capture stopped")
}
```

3. deleting the veth pair while capturing

I didn't find exactly why this happens but we could reuse C.poll instead of unix.Poll. Applying the following patch makes the errors reported again.

```
diff --git a/afpacket/afpacket.go b/afpacket/afpacket.go
index 3acf814..0c1d202 100644
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -117,7 +117,7 @@ type TPacket struct {
        // current is the current header.
        current header
        // pollset is used by TPacket for its poll() call.
-       pollset unix.PollFd
+       pollset C.struct_pollfd
        // shouldReleasePacket is set to true whenever we return packet data, to make sure we remember to release that data back to the kernel.
        shouldReleasePacket bool
        // headerNextNeeded is set to true when header need to move to the next packet. No need to move it case of poll error.
@@ -457,16 +457,16 @@ func (h *TPacket) getTPacketHeader() header {
 func (h *TPacket) pollForFirstPacket(hdr header) error {
        tm := int(h.opts.pollTimeout / time.Millisecond)
        for hdr.getStatus()&C.TP_STATUS_USER == 0 {
-               h.pollset.Fd = int32(h.fd)
-               h.pollset.Events = unix.POLLIN
-               h.pollset.Revents = 0
-               n, err := unix.Poll([]unix.PollFd{h.pollset}, tm)
+               h.pollset.fd = C.int(h.fd)
+               h.pollset.events = C.POLLIN
+               h.pollset.revents = 0
+               n, err := C.poll(&h.pollset, 1, C.int(tm))
                if n == 0 {
                        return ErrTimeout
                }
 
                atomic.AddInt64(&h.stats.Polls, 1)
-               if h.pollset.Revents&unix.POLLERR > 0 {
+               if h.pollset.revents&C.POLLERR > 0 {
                        return ErrPoll
                }
                if err != nil {
```

notti:
As mentioned in #529 Revents is empty. Well the original code copied
h.pollset into a slice as argument to unix.Poll and then discards this
slice.... -> unix.Poll modified the slice, which was immediately
discarded. No wonder h.pollset.Revents is empty.

This removes pollset from the TPacket struct (according to the compiler
pollset doesn't escape to the heap anyway), and properly uses a slice
for everything.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -455,13 +455,13 @@ func (h *TPacket) getTPacketHeader() header {
 func (h *TPacket) pollForFirstPacket(hdr header) error {
 	tm := int(h.opts.pollTimeout / time.Millisecond)
 	for hdr.getStatus()&C.TP_STATUS_USER == 0 {
-		pollset := []unix.PollFd{
-			unix.PollFd{
+		pollset := [1]unix.PollFd{
+			{
 				Fd:     int32(h.fd),
 				Events: unix.POLLIN,
 			},
 		}
-		n, err := unix.Poll(pollset, tm)
+		n, err := unix.Poll(pollset[:], tm)
 		if n == 0 {
 			return ErrTimeout
 		}
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -116,8 +116,6 @@ type TPacket struct {
 	offset int
 	// current is the current header.
 	current header
-	// pollset is used by TPacket for its poll() call.
-	pollset unix.PollFd
 	// shouldReleasePacket is set to true whenever we return packet data, to make sure we remember to release that data back to the kernel.
 	shouldReleasePacket bool
 	// headerNextNeeded is set to true when header need to move to the next packet. No need to move it case of poll error.
@@ -457,16 +455,19 @@ func (h *TPacket) getTPacketHeader() header {
 func (h *TPacket) pollForFirstPacket(hdr header) error {
 	tm := int(h.opts.pollTimeout / time.Millisecond)
 	for hdr.getStatus()&C.TP_STATUS_USER == 0 {
-		h.pollset.Fd = int32(h.fd)
-		h.pollset.Events = unix.POLLIN
-		h.pollset.Revents = 0
-		n, err := unix.Poll([]unix.PollFd{h.pollset}, tm)
+		pollset := []unix.PollFd{
+			unix.PollFd{
+				Fd:     int32(h.fd),
+				Events: unix.POLLIN,
+			},
+		}
+		n, err := unix.Poll(pollset, tm)
 		if n == 0 {
 			return ErrTimeout
 		}
 
 		atomic.AddInt64(&h.stats.Polls, 1)
-		if h.pollset.Revents&unix.POLLERR > 0 {
+		if pollset[0].Revents&unix.POLLERR > 0 {
 			return ErrPoll
 		}
 		if err != nil {


afpacket/afpacket.go
```diff
@@ -457,16 +455,19 @@ func (h *TPacket) getTPacketHeader() header {
 func (h *TPacket) pollForFirstPacket(hdr header) error {
 	tm := int(h.opts.pollTimeout / time.Millisecond)
 	for hdr.getStatus()&C.TP_STATUS_USER == 0 {
-		h.pollset.Fd = int32(h.fd)
-		h.pollset.Events = unix.POLLIN
-		h.pollset.Revents = 0
-		n, err := unix.Poll([]unix.PollFd{h.pollset}, tm)
+		pollset := []unix.PollFd{
```
gconnell: I'm somewhat worried that this does an alloc for each poll (not the worst, but we can probably improve).  How about this:

pollset := [1]unix.PollFd{ ... }
unix.Poll(pollset[:], tm)

notti: Well as I have written in the pullrequest the go compiler says that pollset does not escape, which means it should live purely on the heap. But since checking is better than believing I added a memory profile to the afpacket example and tried it out with iperf on localhost (this uncovered another but with afpacket I'll send a pull request for) and according to the memory profile no allocations happen there.

A look at the assembly shows:
```
		pollset := []unix.PollFd{
  0x4d4d7d		0fb7058a331000		MOVZX github.com/google/gopacket/afpacket.statictmp_1+6(SB), AX	
  0x4d4d84		8b0d7e331000		MOVL github.com/google/gopacket/afpacket.statictmp_1(SB), CX	
  0x4d4d8a		0fb7157b331000		MOVZX github.com/google/gopacket/afpacket.statictmp_1+4(SB), DX	
  0x4d4d91		894c2440		MOVL CX, 0x40(SP)						
  0x4d4d95		6689542444		MOVW DX, 0x44(SP)						
  0x4d4d9a		6689442446		MOVW AX, 0x46(SP)						
				Fd:     int32(h.fd),
  0x4d4d9f		488b442458		MOVQ 0x58(SP), AX	
  0x4d4da4		488b08			MOVQ 0(AX), CX		
  0x4d4da7		894c2440		MOVL CX, 0x40(SP)	
		n, err := unix.Poll(pollset, tm)
  0x4d4dab		488d4c2440		LEAQ 0x40(SP), CX			
  0x4d4db0		48890c24		MOVQ CX, 0(SP)				
  0x4d4db4		48c744240801000000	MOVQ $0x1, 0x8(SP)			
  0x4d4dbd		48c744241001000000	MOVQ $0x1, 0x10(SP)			
  0x4d4dc6		488b4c2438		MOVQ 0x38(SP), CX			
  0x4d4dcb		48894c2418		MOVQ CX, 0x18(SP)			
  0x4d4dd0		e85bdcfeff		CALL golang.org/x/sys/unix.Poll(SB)	
```
This means the golang created a single `PollFd` statically with everything zero except for Events (filled with `POLLIN`), that gets copied to registers (`0x4d4d7d` - `8a`) and then to the stack (`0x4d4d91` - `9a`) - so no allocation. `0x4d4db0` - `0x4d4dbd` create the slice on the stack (`CX` holds the pointer to the previous `PollFd` at `0x40(SP)`, and the two `$0x1` are `Len` and `Cap`).
So basically golang actually created a single static `PollFD` in the binary, which gets copied to the stack, followed by setting `Fd`, and finally copying the slice header onto the stack.

Since this looks a bit different then I suspected (and I also just learned about `statictmp` - didn't know that one) I also tried your suggestion:
```
                pollset := [1]unix.PollFd{
  0x4d4d7d              48c744243800000000      MOVQ $0x0, 0x38(SP)     
                                Fd:     int32(h.fd),
  0x4d4d86              488b442458              MOVQ 0x58(SP), AX       
  0x4d4d8b              488b08                  MOVQ 0(AX), CX          
  0x4d4d8e              894c2438                MOVL CX, 0x38(SP)       
  0x4d4d92              66c744243c0100          MOVW $0x1, 0x3c(SP)     
                n, err := unix.Poll(pollset[:], tm)
  0x4d4d99              488d4c2438              LEAQ 0x38(SP), CX                       
  0x4d4d9e              48890c24                MOVQ CX, 0(SP)                          
  0x4d4da2              48c744240801000000      MOVQ $0x1, 0x8(SP)                      
  0x4d4dab              48c744241001000000      MOVQ $0x1, 0x10(SP)                     
  0x4d4db4              488b4c2440              MOVQ 0x40(SP), CX                       
  0x4d4db9              48894c2418              MOVQ CX, 0x18(SP)                       
  0x4d4dbe              e86ddcfeff              CALL golang.org/x/sys/unix.Poll(SB)   
```
Hmm ok that one does the same - but instead of the static `PollFd` it just uses a zero literal (`0x4d4d7d`) and an additional `MOVW` for the `POLLIN` (`0x4d4d92`). The rest is the same (slice header).

So although my version also doesn't allocate anything, I think I like yours more (less MOVs). I'll send a modified version.

gconnell:
Thanks for the detailed looking-into-things!

PR #540, opened 2018-10-08
geneve fix infinite loop with options len miss-matching the total opt…

safchain:
…ion len


gconnell:
Thanks for the fast response!

PR #541, opened 2018-10-10
Update parser.go

ATpiu:
the same issue in Example usage should also be corrected and the url is https://godoc.org/github.com/google/gopacket


ATpiu:
I signed it!

gconnell:
Thanks!

PR #542, opened 2018-10-11
Adding 'payload' support for all management frames

Galadros:
A number of management frames don't have `DecodeFromBytes` implemented and only inherit it.  Normally this would be relatively innocuous- if those frames don't have specific fields that we're extracting, you wouldn't think we'd need this function.  However, the structs, such as `Dot11MgmtProbeReq`, all have a "Payload" field which can be extracted via the `LayerPayload()` function.  If you try and decode a packet step by step, by starting with a first layer such as `layers.LayerTypeRadioTap` and then successively calling `NextLayerType()` and `LayerPayload()` after you decode the current layer, you're unable to retrieve information elements from these frame types.  The default `DecodeFromBytes` method implemented for the `Dot11Mgmt` type never sets the `Payload` field for that layer.  As such, you can never reach the information elements by successfully going through the layers in this way.

I updated the file by adding `NextLayerType()` and `DecodeFromBytes()` to almost every management frame subtype that didn't have them.  The `DecodeFromBytes()` function is mostly a dummy function; all it does is set `m.Payload` equal to the current `data` following the methodology that other `DecodeFromBytes()` methods use to set the payload.  I also added `NextLayerType()` to return `LayerTypeDot11InformationElement` for the management frames that can contain information elements but didn't have `NextLayerType()` implemented.

Example of loop follows; this loop will fail to retrieve info elements from a probe request packet, for instance:

```
for len(data) > 0 {
        decoder, ok := mydecoders[typ]
        if !ok {
            if l.IgnoreUnsupported {
                return nil
            }
            return gopacket.UnsupportedLayerType(typ)
        } else {
            if typ != 82 {
                if err = decoder.DecodeFromBytes(data, df); err != nil {
                    return err
                }
            } else {
                if err = (*infoElements)[index].DecodeFromBytes(data, df); err != nil {
                    return err
                }
                decoder = (*infoElements)[index]
                index = index + 1
            }
        }
        *decoded = append(*decoded, typ)
        typ = decoder.NextLayerType()
        data = decoder.LayerPayload()

    }

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -129,7 +129,7 @@ func BenchmarkUnbufferedChannel(b *testing.B) {
 	defer close(ca)
 	go func() {
 		defer close(cb)
-		for _ = range ca {
+		for range ca {
 			cb <- true
 		}
 	}()
@@ -144,7 +144,7 @@ func BenchmarkSmallBufferedChannel(b *testing.B) {
 	defer close(ca)
 	go func() {
 		defer close(cb)
-		for _ = range ca {
+		for range ca {
 			cb <- true
 		}
 	}()
@@ -159,7 +159,7 @@ func BenchmarkLargeBufferedChannel(b *testing.B) {
 	defer close(ca)
 	go func() {
 		defer close(cb)
-		for _ = range ca {
+		for range ca {
 			cb <- true
 		}
 	}()

### CODE REVIEW

File: layers/dot11.go [hunk 1]
```diff
@@ -2061,6 +2086,11 @@ func decodeDot11MgmtAction(data []byte, p gopacket.PacketBuilder) error {
 
 func (m *Dot11MgmtAction) LayerType() gopacket.LayerType  { return LayerTypeDot11MgmtAction }
 func (m *Dot11MgmtAction) CanDecode() gopacket.LayerClass { return LayerTypeDot11MgmtAction }
+func (m *Dot11MgmtAction) NextLayerType() gopacket.LayerType {return LayerTypeDot11InformationElement}
+func (m *Dot11MgmtAction) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	m.Payload = data
```
notti: This is is why `TestPacketDot11MgmtAction` fails.
The frame in this test has 2 fixed parameters (category code and action) and after that one `LayerTypeDot11InformationElement` followed by 4 byte `FCS`.
In your version the two fixed parameters get decoded as `LayerTypeDot11InformationElement`, which is wrong and thus the incorrect length.
Sadly I can't find a good source for how this is done correctly (https://raw.githubusercontent.com/wireshark/wireshark/a5faef2e4fe9c8bb4b55dcb3e60c32a2a504ac9f/epan/dissectors/packet-ieee80211.c is way too much to read for reviewing this)

layers/dot11.go
```diff
@@ -1796,6 +1796,10 @@ func (m *Dot11MgmtReassociationResp) CanDecode() gopacket.LayerClass {
 func (m *Dot11MgmtReassociationResp) NextLayerType() gopacket.LayerType {
 	return LayerTypeDot11InformationElement
 }
+func (m *Dot11MgmtReassociationResp) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	m.Payload = data
```
notti: This type has fixed parameters

layers/dot11.go
```diff
@@ -1874,6 +1882,14 @@ func (m *Dot11MgmtMeasurementPilot) LayerType() gopacket.LayerType {
 func (m *Dot11MgmtMeasurementPilot) CanDecode() gopacket.LayerClass {
 	return LayerTypeDot11MgmtMeasurementPilot
 }
+func (m *Dot11MgmtMeasurementPilot) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	m.Payload = data
```
notti: This type has fixed parameters

layers/dot11.go
```diff
@@ -1928,6 +1944,11 @@ func decodeDot11MgmtATIM(data []byte, p gopacket.PacketBuilder) error {
 
 func (m *Dot11MgmtATIM) LayerType() gopacket.LayerType  { return LayerTypeDot11MgmtATIM }
 func (m *Dot11MgmtATIM) CanDecode() gopacket.LayerClass { return LayerTypeDot11MgmtATIM }
+func (m *Dot11MgmtATIM) NextLayerType() gopacket.LayerType {return LayerTypeDot11InformationElement}
```
notti: The body of ATIM is null according to the spec - so there should be no next layer

layers/dot11.go
```diff
@@ -2073,6 +2103,11 @@ func decodeDot11MgmtActionNoAck(data []byte, p gopacket.PacketBuilder) error {
 
 func (m *Dot11MgmtActionNoAck) LayerType() gopacket.LayerType  { return LayerTypeDot11MgmtActionNoAck }
 func (m *Dot11MgmtActionNoAck) CanDecode() gopacket.LayerClass { return LayerTypeDot11MgmtActionNoAck }
+func (m *Dot11MgmtActionNoAck) NextLayerType() gopacket.LayerType {return LayerTypeDot11InformationElement}
+func (m *Dot11MgmtActionNoAck) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	m.Payload = data
```
notti: This type has fixed parameters

Galadros:
Hey- sorry I didn't get back to you earlier, I kinda checked out with family for the holidays.  Just to be clear- you're saying that my changes overlook certain fixed elements in some frame types, and as such they're essentially incomplete.  Is that a fair assessment?  I'm correct that as-is I can't get the payload through standard layer decoding- so in order to use that info I _do_ have to make some changes- but that in adding the ability to ask for the payload of some management frames that didn't previously provide payload info, I didn't account for some fixed elements in certain frame types.

layers/dot11.go
Galadros: Good point, tweaked that.

layers/dot11.go
Galadros: Fixed- I simply mirrored over the parsing code for association response frames, since they have identical- and identically formatted- fixed paramaters.

notti:
> you're saying that my changes overlook certain fixed elements in some frame types, and as such they're essentially incomplete. Is that a fair assessment?

Yes. I suspect that's why those were missing `DecodeFromBytes`.

> I'm correct that as-is I can't get the payload through standard layer decoding- so in order to use that info I _do_ have to make some changes- but that in adding the ability to ask for the payload of some management frames that didn't previously provide payload info, I didn't account for some fixed elements in certain frame types.

I'm not completely sure what you mean here.
Adding the fixed fields in Dot11<frametype> struct, decode them in `DecodeFromBytes` (and remove the consumed bytes from the payload) and encode them in `SerializeTo` should suffice (like in `Dot11MgmtProbeResp` for example). Well except for frame types that explicitly have no payloads (e.g. ATIM I think).

For the FCS issue, it should be ok to include a temporary workaround in the one test that will then fail, add a comment that that needs to be resolved in the future, also state it in the documentation and keep the issue around (#544 - rename it or make a new one that better states the issue with the FCS), because resolving that sounds like a huge effort and after all this has been around for a long time and no one complained...

PR #545, opened 2018-10-17
Add go implementation for handling pcapng files

notti:
This is a native pcapng implementation for reading and writing files.

Code was optimized with the help of benchmarking and analysing compiler
output by
 * trying to eliminate bounds checks
 * limit itab use (interfaces)
 * limit number of go function calls
 * trying to inline as much as possible without duplicating too much
   code (manually and automatically)
 * named return values

Things that lead to surprising results:
 * use of unsafe for writing (pcapng states one should use native
   endianess) - encoding.binary is way too optimized
 * binary.*.PutUint16(..., 0) is faster than assigning 0 to both slice
   elements
 * Inline budget is already used up by just calling io.ReadFull(r.r,
   ...)

### DIFF
--- a/pcapgo/ngread.go
+++ b/pcapgo/ngread.go
@@ -232,13 +232,14 @@ RESTART:
 
 	var section NgSectionInfo
 
+OPTIONS:
 	for {
 		if err := r.readOption(); err != nil {
 			return err
 		}
 		switch r.currentOption.code {
 		case ngOptionCodeEndOfOptions:
-			goto DONE
+			break OPTIONS
 		case ngOptionCodeComment:
 			section.Comment = string(r.currentOption.value)
 		case ngOptionCodeHardware:
@@ -249,7 +250,7 @@ RESTART:
 			section.Application = string(r.currentOption.value)
 		}
 	}
-DONE:
+
 	if _, err := r.r.Discard(int(r.currentBlock.length)); err != nil {
 		return err
 	}
@@ -330,13 +331,14 @@ func (r *NgReader) readInterfaceDescriptor() error {
 	intf.LinkType = layers.LinkType(r.getUint16(r.buf[:2]))
 	intf.SnapLength = r.getUint32(r.buf[4:8])
 
+OPTIONS:
 	for {
 		if err := r.readOption(); err != nil {
 			return err
 		}
 		switch r.currentOption.code {
 		case ngOptionCodeEndOfOptions:
-			goto DONE
+			break OPTIONS
 		case ngOptionCodeInterfaceName:
 			intf.Name = string(r.currentOption.value)
 		case ngOptionCodeComment:
@@ -354,7 +356,6 @@ func (r *NgReader) readInterfaceDescriptor() error {
 			intf.TimestampResolution = NgResolution(r.currentOption.value[0])
 		}
 	}
-DONE:
 	if _, err := r.r.Discard(int(r.currentBlock.length)); err != nil {
 		return err
 	}
@@ -405,13 +406,14 @@ func (r *NgReader) readInterfaceStatistics() error {
 	*stats = ngEmptyStatistics
 	stats.LastUpdate = time.Unix(r.convertTime(ifaceID, ts)).UTC()
 
+OPTIONS:
 	for {
 		if err := r.readOption(); err != nil {
 			return err
 		}
 		switch r.currentOption.code {
 		case ngOptionCodeEndOfOptions:
-			goto DONE
+			break OPTIONS
 		case ngOptionCodeComment:
 			stats.Comment = string(r.currentOption.value)
 		case ngOptionCodeInterfaceStatisticsStartTime:
@@ -426,7 +428,6 @@ func (r *NgReader) readInterfaceStatistics() error {
 			stats.PacketsDropped = r.getUint64(r.currentOption.value[:8])
 		}
 	}
-DONE:
 	if _, err := r.r.Discard(int(r.currentBlock.length)); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


notti:
Im not completely sure about some points:

 * pcapng can store packets with varying link types. I created this pull request to match libpcap in this regard (ignore packets where the link type doesn't match), but it is possible to provide options with which all the packets can be read. Currently, the gopacket interface does not support this - so I put the LinkType in the ci.AncillaryData. I'm not completely sure if this is the best option.
 * I used test files generated with https://github.com/hadrielk/pcapng-test-generator and two additional tests. I'm not completely sure if this is ok with the license and how to add those files to the repository. Is the way I did it ok?

pcapnggo/write.go
```diff
@@ -0,0 +1,392 @@
...
+	if _, err = w.w.Write(w.buf[:28]); err != nil {
+		return
+	}
+
+	if _, err = w.w.Write(data); err != nil {
+		return
+	}
+
+	binary.LittleEndian.PutUint32(w.buf[:4], 0)
+	_, err = w.w.Write(w.buf[4-padding : 8]) // padding + length
```
notti: No matter what I did, I couldn't convince go to remove this bounds check - but it's the only one in the hot path, so I think it should be ok.

notti:
I just noticed, that golint is somehow broken on travis:
package github.com/golang/lint/golint: code in directory /home/travis/gopath/src/github.com/golang/lint/golint expects import "golang.org/x/lint/golint"

gconnell:
Would it be possible to put this into the existing pcapgo/ directory, maybe as NgReader() and NgWriter() or something like that?

pcapgo/ is meant as somewhat of a dumping ground for native-Go pcap stuff.

notti:
Ah I misread the package documentation... Did the move. Hopefully I didn't miss anything in the documentation...

pcapgo/ngread.go
```diff
@@ -0,0 +1,599 @@
...
+	if r.bigEndian {
+		return binary.BigEndian.Uint64(buffer)
+	}
+	return binary.LittleEndian.Uint64(buffer)
+}
+
+// Now the pcapng implementation
+
+// readBlock reads a the blocktype and length from the file. If the type is a section header, endianess is also read.
+func (r *NgReader) readBlock() (err error) {
```
gconnell: I'm wary of having the returned error be a named return variable named "err".  Lots of Go folks have a ton of muscle memory doing `err := ...` in their code, but if someone does that in a function that returns `(err error)` it has the potential to shadow the return var and cause problems.  I'd lean towards one of the following:

1. just return `error`, and say `err := ...; return err` where you need to
2. use a more descriptive return var name `(returnedErr error)`, to avoid hitting folks' muscle memory

I used to return `(err error)` a ton myself, but I managed to shoot myself in the foot a few times with this, and I've had other folks modifying my code make the same mistakes.

Finally, it's nice to explicitly see a `return nil` at the place in the function where success is returned... otherwise it's non-obvious which is the successful path.  Even seeing a `return` at the end of the function isn't a guarantee... did we reach the end of the function because everything went well, or did we break out of a loop because something went wrong?

pcapgo/ngread.go
```diff
@@ -0,0 +1,599 @@
...
+
+	var section NgSectionInfo
+
+	for {
+		if err = r.readOption(); err != nil {
+			return
+		}
+		switch r.currentOption.code {
+		case ngOptionCodeEndOfOptions:
+			goto DONE
```
gconnell: instead of `goto DONE` here, how about naming the for loop and doing a `break loop` instead?

pcapgo/ngread.go
```diff
@@ -0,0 +1,599 @@
...
+		}
+	}
+}
+
+// SkipSection skips the contents of the rest of the current section and reads the next section header.
+func (r *NgReader) SkipSection() (err error) {
+	if err = r.skipSection(); err != nil {
+		return
+	}
+	err = r.readSectionHeader()
```
gconnell: just `return r.readSectionHeader()` ?

pcapgo/ngread.go
```diff
@@ -0,0 +1,599 @@
...
+	intf.LinkType = layers.LinkType(r.getUint16(r.buf[:2]))
+	intf.SnapLength = r.getUint32(r.buf[4:8])
+
+	for {
+		if err = r.readOption(); err != nil {
+			return
+		}
+		switch r.currentOption.code {
+		case ngOptionCodeEndOfOptions:
+			goto DONE
```
gconnell: name/break loop

```
loop:
  for {
    ...
    break loop
    ...
  }
```

pcapgo/ngread.go
notti: Normally I don't do that for errors, but I read on the internet that named return values give better code (and there is even a bug open with golang), but I must admit I didn't measure it in this case - and seems unnamed is a tiny bit faster in this case (but with a high p value - so YMMV).
For better clarity I change it.

pcap/pcapnggo_test.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2018 The GoPacket Authors. All rights reserved.
```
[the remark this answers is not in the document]
notti: Ah didn't know - I just copied it from another file (but don't remember it from which one)

pcapgo/ngread.go
notti: Fixed by the named -> unnamed return value patch

pcapgo/ngread.go
notti: Well I found it this way clearer since the label is where the code is going to instead of (in the source code - not the generated code) jumping up to the loop header and then jumping to the end of the loop. But since it's all the same to me I'll change it.

pcapgo/ngread.go
notti: See above

PR #548, opened 2018-10-26
add additional 802.11-2016 options

christophert:
802.11-2016/802.11mc includes additional options used for device location.


lhausermann:
Great thanks

PR #549, opened 2018-10-26
Recognize More DHCPv6 Options

cimnine:
This PR adds names to all of the [IANA-listed DHCPv6 options][1] as of 2018-10-26.

In the process I've also extracted anything _DHCPv6 options_-related into it's own file.

[1]: https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#dhcpv6-parameters-2


layers/dhcpv6_options.go
```diff
@@ -0,0 +1,610 @@
+package layers
```
gconnell: Please add full copyright notice here:

```
// Copyright 2012 The GoPacket Authors. All rights reserved.
//
// tree.
```

cimnine:
Is there anything else?

PR #551, opened 2018-10-28
pcapgo/read.go: ReadPacketData also check that CaptureLength <= Length

assafmo:
According to the [pcap packet header specs](https://wiki.wireshark.org/Development/LibpcapFileFormat#Record_.28Packet.29_Header), `CaptureLength <= Length`.


gconnell:
Thanks!

PR #552, opened 2018-10-29
macs: parse flags at startup

sentryo:
Parse command flags at startup to make the URL customizable as was
originally intended.


gconnell:
While this is obviously correct code-wise, I wonder if it'd be better to just update the URL in-code, so future runs are guaranteed to get the latest correct location?  In short, I wonder if the URL _should_ be a flag.  But I'm happy to support this, so rebase/merge-ing.

Issue #419, opened 2018-02-07
PF_RING error when compiling


hamidrezabstn:
when compile my project this error shows

`github.com/google/gopacket/pfring
../../pfring/pfring.go:155: r.pkthdr.extended_hdr.if_index undefined (type C.struct_pfring_extended_pkthdr has no field or method if_index)`

florianl:
Can you provide a minimal working code example,that produces the error?

florianl:
In the used library I still can find the field id_index at 
https://github.com/ntop/PF_RING/blob/5660e0c69648e41a0beb4c58969956372753d4f1/kernel/linux/pf_ring.h#L317

hamidrezabstn:
hi
i've installed PF_RING using their official tutorial : [Tutorial](https://www.ntop.org/pf_ring/installation-guide-for-pf_ring/)
and inside httpassembly  example of gopacket i just include pfring and replace it with pcap! 
and when i try to compile it i see this error! 
i also check  my pf_ring.h file and it has that property just like you have said but still i see that error

florianl:
hi @hamidrezabstn 
it would be great, if you could provide a minimal working code example, that causes this error. In addition, this error sounds for me like a dependency problem.

hamidrezabstn:
hi @florianl 
as i've said  "inside httpassembly example of gopacket i just include pfring and replace it with pcap!"
[example file](https://github.com/google/gopacket/tree/master/examples/httpassembly)

florianl:
hi @hamidrezabstn 
maybe I missunderstood you. If I would use pfring inside the httpassembly example, it would look like [1] (this is not tested and may contain errors, as I don't have PF_RING installed)

Can you run the examples from the PR_FING library [2]?

[1] https://gist.github.com/florianl/d291f08c3a76b0a393c870ff34de7f33
[2] https://github.com/ntop/PF_RING/tree/dev/userland/examples

ggmartins:
+1
```
$go build gistfile1.go
# github.com/google/gopacket/pfring
src/github.com/google/gopacket/pfring/pfring.go:155: r.pkthdr.extended_hdr.if_index undefined (type _Ctype_struct_pfring_extended_pkthdr has no field or method if_index)
```

```
$go get github.com/google/gopacket/pfring
# github.com/google/gopacket/pfring
[...]

Issue #185, opened 2016-04-18
very slow performance with pfring...


derktam:
I have just done a test, regardless of the number of slots, Using the pfring in 1gbps occured a lot of drop(Tot pkt Lost)

gconnell:
Could you provide a little more detail on the code you're running?  Do you have a minimum viable program that exhibits this?

If you're using PacketSource or ReadPacketData, maybe switch to ReadPacketDataTo to avoid an extra alloc.  Doesn't look like we've implemented ZeroCopyReadPacketData for pfring, but we probably should.

Are you doing other stuff in your code, tcp assembly or the like?

Could you generate a CPU profile with your run (https://golang.org/pkg/runtime/pprof/) and generate a SVG of its output to add here?  That might point to a smoking gun.

rahulpowar:
I have a similar issue with PF_RING. Specifically the latencies seem very high.

I have created a minimal `ping` using gopacket and pfring here: https://github.com/redsift/pingdemo

When testing this on Ubuntu 16.04.1 LTS with PF_RING 6.6.0 and the pfring e1000e driver, the latency suggests something is wrong.

The linked project fires a ping packet and times the response.

Hitting 8.8.8.8 takes ~500ms

```
PF_RING ping timing demo: ping 8.8.8.8 from 10.60.0.175
ping time=500.586337ms
ping time=500.571822ms
ping time=500.571827ms
ping time=500.575797ms
ping time=500.577616ms
```

The same test with the zero copy interface is mostly the same though I guess this is not relevant for my single packet test case.

```
PF_RING ping timing demo: ping 8.8.8.8 from 10.60.0.175
#########################################################################
# ERROR: You do not seem to have a valid PF_RING ZC license 6.6.0.170424 for eno1 [Intel 1 Gbit e1000e family]
# ERROR: Please get one at http://shop.ntop.org/.
#########################################################################
# We're now working in demo mode with packet capture and
# transmission limited to 5 minutes
#########################################################################
ping time=500.57761ms
ping time=500.579852ms
ping time=500.565223ms
ping time=500.56318ms
ping time=500.569313ms
```

As a control, command line ping reports tiny latencies.

```
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
64 bytes from 8.8.8.8: icmp_seq=1 ttl=59 time=4.50 ms
64 bytes from 8.8.8.8: icmp_seq=2 ttl=59 time=4.12 ms
[...]

PR #553, opened 2018-11-01
Fix pf_ring 7.2 and maybe performance issues. This fixes #467, #419

### LINKED ISSUE #467: Build fails when using 'pfring 7.2.0-stable'
**vendor/github.com/google/gopacket/pfring/pfring.go:155: r.pkthdr.extended_hdr.if_index undefined (type C.struct_pfring_extended_pkthdr has no field or method if_index)**

   

- PF_RING/kernel/linux/pf_ring.h  (branch 7.2.0-stable)
```
struct pfring_extended_pkthdr {
  u_int64_t timestamp_ns;  /* Packet timestamp at ns precision. Note that if your NIC supports
                  hardware timestamp, this is the place to read timestamp from */
#define PKT_FLAGS_CHECKSUM_OFFLOAD    1 << 0 /* IP/TCP checksum offload enabled */
#define PKT_FLAGS_CHECKSUM_OK         1 << 1 /* Valid checksum (with IP/TCP checksum offload enabled) */
#define PKT_FLAGS_IP_MORE_FRAG        1 << 2 /* IP More fragments flag set */
#define PKT_FLAGS_IP_FRAG_OFFSET      1 << 3 /* IP fragment offset set (not 0) */
#define PKT_FLAGS_VLAN_HWACCEL        1 << 4 /* VLAN stripped by hw */
#define PKT_FLAGS_FLOW_OFFLOAD_UPDATE 1 << 6 /* Flow update metadata, see generic_flow_update struct (keep flag compatible with ZC) */
#define PKT_FLAGS_FLOW_OFFLOAD_PACKET 1 << 7 /* Flow raw packet, pkt_hash contains the flow_id (keep flag compatible with ZC) */
#define PKT_FLAGS_FLOW_OFFLOAD_MARKER 1 << 8 /* Flow raw packet belongs to a flow that has been marked (keep flag compatible with ZC) */
  u_int32_t flags;

  u_int8_t rx_direction;   /* 1=RX: packet received by the NIC, 0=TX: packet transmitted by the NIC */
  int32_t  if_index;       /* index of the interface on which the packet has been received.
                  It can be also used to report other information */
  u_int32_t pkt_hash;      /* Hash based on the packet header */

  /* --- short header ends here --- */

  struct {
    int32_t bounce_interface; /* Interface Id where this packet will bounce after processing
                 if its values is other than UNKNOWN_INTERFACE */
    struct sk_buff *reserved; /* Kernel only pointer */
  } tx;

  /* NOTE: leave it as last field of the memset on parse_pkt() will fail */
  struct pkt_parsing_info parsed_pkt; /* packet parsing info */
} __attribute__((packed));

/* NOTE: Keep 'struct pfring_pkthdr' in sync with 'struct pcap_pkthdr' */

struct pfring_pkthdr {
  /* pcap header */
  struct timeval ts;    /* time stamp */
  u_int32_t caplen;     /* length of portion present */
  u_int32_t len;        /* length of whole packet (off wire) */
  struct pfring_extended_pkthdr extended_hdr; /* PF_RING extended header */
} __attribute__((packed));

```

The structure seems to have not changed from the previous version, but a build error occurs.

### LINKED ISSUE #543: pf_ring read packet slow with cgo
I use the go interface of af_packet and pf_ring to capture packets. I find the the pfring.ReadPacketDataTo() function takes three times as much time as afpacket.ReadPacketDataTo().
Both of these two function call the C interface with Cgo. What the reason of this difference ?
Is that the reason that pfring.ReadPacketDataTo() call the Cgo more frequently and afpacket.ReadPacketDataTo() avoid to call Cgo each packet? And the consumption of Cgo is very large.

In pfring.ReadPacketDataTo(), each packet reading operation will call Cgo by **C.pfring_readpacketdatato_wrapper**:
```go
r.bufPtr = (*C.u_char)(unsafe.Pointer(&data[0]))
	result := NextResult(C.pfring_readpacketdatato_wrapper(r.cptr, r.bufPtr, C.u_int(len(data)),  &r.pkthdr))  //cgo
}
```
However,in afpacket.ReadPacketDataTo(), the function avoid to call Cgo each packet reading.(It read a block of packets at each Cgo(**h.pollForFirstPacket**) operation )
```go
retry:
	if h.current == nil || !h.headerNextNeeded || !h.current.next() {
		if h.shouldReleasePacket {
			h.releaseCurrentPacket()
		}
		h.current = h.getTPacketHeader()
		if err = h.pollForFirstPacket(h.current); err != nil {
			h.headerNextNeeded = false
			//h.mu.Unlock()
			return
		}
		// We received an empty block
		if h.current.getLength() == 0 {
			goto retry
		}
	}
```

notti:
Rework of the core parts of pfring. Due to the use of packed structures in
7.2 cgo can't access some fields of `struct pfring_pkthdr`.
This copies the needed fields to a non-packed struct first.

To improve speed, `pfring_recv` is now called with a length of zero,
enabling zero copy. The old version caused the userlab pf_ring library
to copy the packet contents. This maybe fixes #543, #185.
Additionally, pfring now uses `timestamp_ns` instead of `ts`, which increases
the timestamp resolution from microsecond to nanosecond. Furthermore,
cgo usage was improved by using the same techniques as in pcap
(allocations on the heap).

Added `ZeroCopyReadPacketData`. This one uses not zero copy from kernel
space directly to the buffer. The old `ReadPacketDataTo` caused pf_ring
userspace library to copy the data (buffer_len != 0). The new
`ReadPacketDataTo` still does the copy (but now in gocode) since zero
copy is not possible with the current interface. This method is there to
keep legacy code working, but newer versions should switch over to
`ZeroCopyReadPacketData`.

The following benchmarks were performed in a virtual machine by using
pktgen kernel module on the host and virtio-net in the guest. Since
there is no `ZeroCopyReadPacketData`, the `PfringReadZero` lines compare
`ReadPacketDataTo` in the old implementation with
`ZeroCopyReadPacketData` in the new version.

65536 byte size packets:
```
name              old time/op    new time/op    delta
PfringRead-2        47.5µs ±10%    42.8µs ± 7%    -9.85%  (p=0.012 n=8+10)
PfringReadZero-2    42.5µs ±25%    32.9µs ± 7%   -22.65%  (p=0.000 n=10+10)
PfringReadTo-2      42.5µs ±25%    36.8µs ± 9%      ~     (p=0.143 n=10+10)

name              old speed      new speed      delta
PfringRead-2      1.30GB/s ±26%  1.53GB/s ± 7%   +17.57%  (p=0.003 n=10+10)
PfringReadZero-2  1.58GB/s ±22%  1.99GB/s ± 7%   +25.90%  (p=0.000 n=10+10)
PfringReadTo-2    1.58GB/s ±22%  1.79GB/s ± 9%      ~     (p=0.143 n=10+10)

name              old alloc/op   new alloc/op   delta
PfringRead-2        65.6kB ± 0%    65.5kB ± 0%    -0.02%  (p=0.000 n=10+10)
PfringReadZero-2     16.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)
PfringReadTo-2       16.0B ± 0%      0.0B       -100.00%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op  delta
PfringRead-2          2.00 ± 0%      1.00 ± 0%   -50.00%  (p=0.000 n=10+10)
PfringReadZero-2      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
PfringReadTo-2        1.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)
```

58 byte size packets:
```
name              old time/op    new time/op     delta
PfringRead-2        46.8µs ±23%      3.2µs ±85%    -93.15%  (p=0.000 n=10+10)
PfringReadZero-2    2.48µs ±11%     2.33µs ± 2%     -6.37%  (p=0.022 n=10+8)
PfringReadTo-2      2.48µs ±11%     2.63µs ± 6%       ~     (p=0.053 n=10+9)

name              old speed      new speed       delta
PfringRead-2      1.27MB/s ±24%  21.17MB/s ±60%  +1570.88%  (p=0.000 n=10+10)
PfringReadZero-2  23.5MB/s ±10%   24.9MB/s ± 2%     +6.33%  (p=0.017 n=10+8)
PfringReadTo-2    23.5MB/s ±10%   22.0MB/s ± 6%       ~     (p=0.053 n=10+9)

name              old alloc/op   new alloc/op    delta
PfringRead-2        65.6kB ± 0%      0.1kB ± 0%    -99.90%  (p=0.000 n=10+10)
PfringReadZero-2     16.0B ± 0%       0.0B        -100.00%  (p=0.000 n=10+10)
PfringReadTo-2       16.0B ± 0%       0.0B        -100.00%  (p=0.000 n=10+10)

name              old allocs/op  new allocs/op   delta
PfringRead-2          2.00 ± 0%       1.00 ± 0%    -50.00%  (p=0.000 n=10+10)
PfringReadZero-2      1.00 ± 0%       0.00        -100.00%  (p=0.000 n=10+10)
PfringReadTo-2        1.00 ± 0%       0.00        -100.00%  (p=0.000 n=10+10)
```

### DIFF
--- a/pfring/pfring_test.go
+++ b/pfring/pfring_test.go
@@ -1,0 +1,64 @@
+package pfring
+
+import (
+	"flag"
+	"log"
+	"testing"
+)
+
+var iface = flag.String("i", "eth0", "Interface to read packets from")
+
+func BenchmarkPfringRead(b *testing.B) {
+	var ring *Ring
+	var err error
+	if ring, err = NewRing(*iface, 65536, FlagPromisc); err != nil {
+		log.Fatalln("pfring ring creation error:", err)
+	}
+	if err = ring.SetSocketMode(ReadOnly); err != nil {
+		log.Fatalln("pfring SetSocketMode error:", err)
+	} else if err = ring.Enable(); err != nil {
+		log.Fatalln("pfring Enable error:", err)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, ci, _ := ring.ReadPacketData()
+                b.SetBytes(int64(ci.CaptureLength))
+	}
+}
+
+func BenchmarkPfringReadZero(b *testing.B) {
+	var ring *Ring
+	var err error
+	if ring, err = NewRing(*iface, 65536, FlagPromisc); err != nil {
+		log.Fatalln("pfring ring creation error:", err)
+	}
+	if err = ring.SetSocketMode(ReadOnly); err != nil {
+		log.Fatalln("pfring SetSocketMode error:", err)
+	} else if err = ring.Enable(); err != nil {
+		log.Fatalln("pfring Enable error:", err)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, ci, _ := ring.ZeroCopyReadPacketData()
+                b.SetBytes(int64(ci.CaptureLength))
+	}
+}
+
+func BenchmarkPfringReadTo(b *testing.B) {
+	var ring *Ring
+	var err error
+	if ring, err = NewRing(*iface, 65536, FlagPromisc); err != nil {
+		log.Fatalln("pfring ring creation error:", err)
+	}
+	if err = ring.SetSocketMode(ReadOnly); err != nil {
+		log.Fatalln("pfring SetSocketMode error:", err)
+	} else if err = ring.Enable(); err != nil {
+		log.Fatalln("pfring Enable error:", err)
+	}
+	buffer := make([]byte, 65536*2)
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		ci, _ := ring.ReadPacketDataTo(buffer)
+                b.SetBytes(int64(ci.CaptureLength))
+	}
+}
--- a/pfring/pfring_test.go
+++ b/pfring/pfring_test.go
@@ -22,7 +22,7 @@ func BenchmarkPfringRead(b *testing.B) {
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		_, ci, _ := ring.ReadPacketData()
-                b.SetBytes(int64(ci.CaptureLength))
+		b.SetBytes(int64(ci.CaptureLength))
 	}
 }
 
@@ -40,7 +40,7 @@ func BenchmarkPfringReadZero(b *testing.B) {
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		_, ci, _ := ring.ZeroCopyReadPacketData()
-                b.SetBytes(int64(ci.CaptureLength))
+		b.SetBytes(int64(ci.CaptureLength))
 	}
 }
 
@@ -59,6 +59,6 @@ func BenchmarkPfringReadTo(b *testing.B) {
 	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		ci, _ := ring.ReadPacketDataTo(buffer)
-                b.SetBytes(int64(ci.CaptureLength))
+		b.SetBytes(int64(ci.CaptureLength))
 	}
 }
--- a/pfring/pfring_test.go
+++ b/pfring/pfring_test.go
@@ -1,3 +1,9 @@
+// Copyright 2019 The GoPacket Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
 package pfring
 
 import (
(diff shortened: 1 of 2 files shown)


notti:
@gconnell any idea how govet fails here? The pfring directory isn't checked by govet, I didn't change anything outside it, and if I execute go vet on my computer there are no errors.

notti:
ahh nevermind I looked at the wrong line

pfring/pfring_test.go
```diff
@@ -0,0 +1,64 @@
+package pfring
```
gconnell: copyright.

notti: I knew I forgot something...

PR #555, opened 2018-11-03
Add nanosecond timestamp resolution to pcap, and Resolution API

### LINKED ISSUE #298: Support Nanosecond timestamp in pcap files
There's currently no support for getting nanosecond timestamps out of read pcap files using the libpcap library. There is currently support for this in the pcapgo module, but using the C pcap library will allow me to use the BPF filter. 

I'm happy to do the work to make the change if you are happy to accept it, but wanted feedback on the approach before I started. 

I was going to use the pcap_fopen_offline_with_tstamp_precision method to open the pcap files, if it existed. If it did not, then I will create a stub implementation, similar to how pcap_set_immediate_mode is stubbed out. If the sub method is used, then read time stamps will be multiplied by 1000. 

For live captures, I wanted to add a call to "pcap_set_tstamp_precision" to set the precision to nanoseconds, and again, if it fails, do the multiply. 

I don't see this having any external effects on the library because all time stamps are exposed as a Timestamp which is a nanosecond format. 

Thanks for your comments.

notti:
Resolution API provides a standardized way to report timestamp resolution of PacketSources.

In case libpcap (version < 1.5), the file, hardware, or os does not
support nanosecond resolution, microsecond is used. Actual resolution
can be queried with Resolution().
This fixes #298.


pcap/pcap.go
```diff
@@ -33,11 +33,42 @@ package pcap
...
 
+#define PCAP_TSTAMP_PRECISION_MICRO	0
+#define PCAP_TSTAMP_PRECISION_NANO	1
+
+#ifdef WIN32
+pcap_t *pcap_hopen_offline_with_tstamp_precision(intptr_t osfd, u_int precision,
+  char *errbuf) {
+  return pcap_hopen_offline(osdf, errbuf);
+}
+#else
```
gconnell: move these into pcap_unix/pcap_windows?

pcap/pcap.go
```diff
@@ -33,11 +33,42 @@ package pcap
...
 
 int pcap_set_immediate_mode(pcap_t *p, int mode) {
   return PCAP_ERROR;
 }
 
+#define PCAP_TSTAMP_PRECISION_MICRO	0
+#define PCAP_TSTAMP_PRECISION_NANO	1
+
+#ifdef WIN32
+pcap_t *pcap_hopen_offline_with_tstamp_precision(intptr_t osfd, u_int precision,
```
gconnell: it looks like right now precision is being ignored?  maybe add a comment in-code saying why?

(same for fopen)

time.go
```diff
@@ -0,0 +1,43 @@
...
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package gopacket
+
+import "fmt"
+
+// TimestampResolution represents the resolution of timestamps in Base^Exponent.
+type TimestampResolution struct {
```
gconnell: should this struct have a ToTime(int64) function attached to it?

pcapgo/pcapng.go
```diff
@@ -157,6 +158,17 @@ type NgInterface struct {
 	scaleDown  uint64
 }
 
+// Resolution returns the timestamp resolution of acquired timestamps before scaling to NanosecondTimestampResolution.
+func (i NgInterface) Resolution() (ret gopacket.TimestampResolution) {
+	if i.TimestampResolution.Binary() {
```
gconnell: maybe make TimestampResolution have an Inverse() function that returns a TimestampResolution?

time.go
notti: Hmm I'm not quiet sure if I understand correctly what this function should do.
Did you mean something, that returns `argument*TimestampResolution` as time.Time?
Hmm if I think more about this, this could actually only return a time.Duration, since the TimestampResolution does not specify an epoch (e.g., ntp timestamps use a different epoch than unix)

pcap/pcap.go
notti: Ah yeah, thats for libpcap < 1.5 (see ifndef above) since before that version libpcap didn't support changing the resolution (it's fixed to micro). I'll add a comment.

notti:
I addressed everything except the ToTime, which I think we should finish discussing before merging.

notti:
Maybe it would also be better to limit base to 2 and 10 for now? (Not sure if there is any other base anywhere)

pcap/doc.go
```diff
@@ -12,6 +12,12 @@ This package is meant to be used with its parent,
 http://github.com/google/gopacket, although it can also be used independently
 if you just want to get packet data from the wire.
 
+Depending on libpcap version, os support, or file timestamp resolution,
+nanosecond resolution is used for the internal timestamps. Returned timestamps
+are always scaled to nanosecond resolution due to the usage of time.Time.
+libpcap must be at least version 1.5 to support nanosecond timestamps. OpenLive
+suports only microsecond resolution.
```
gconnell: supports

notti: Omg I stared for minutes at this and had to ask my wife who immediately pointed out that there is a p missing o_O

gconnell: Just think, if I hadn't been here to review the code, an insignificant typo may have made it into a comment :P 

This is how I know I'm really useful as a reviewer :D

notti: fixed

time.go
notti: I added a ToDuration function, tests and discovered a bug (milli vs micro was wrong...)

PR #556, opened 2018-11-04
Ip4defrag - set IPv4 Id field to match the value of the packets instead of 0

FaranIdo:
According to the RFC791:
- "the identification field is used to distinguish the fragments of one datagram from those of another"
- "To assemble the fragments of an internet datagram, an internet protocol module (for example at a destination host) combines internet datagrams that all have the same value for the four fields: identification, source, destination, and protocol".

It is fine to use the Id of the last packet for the output packet because the Id field is used to identify seen flows (all packets in the same flow have the same Id value).

Added fixed code + test that validate it.


notti:
Thanks!

PR #559, opened 2018-11-06
reassemblydump example: Fixed data races on errorsMap and tcpStream

dreadl0ck:
The reassemblydump example contains two data races:

1) The errorsMap is potentially accessed simultaneously by multiple goroutines.
A mutex was added to make it thread safe.

2) When using stream.run for both client and server for parsing http requests and responses,
both use a pointer to the parent stream to access the stream.urls array without any synchronization.
A mutex was also added to the tcpStream structure to prevent this.

### DIFF
--- a/examples/reassemblydump/main.go
+++ b/examples/reassemblydump/main.go
@@ -117,13 +117,16 @@ func (h *httpReader) Read(p []byte) (int, error) {
 
 var outputLevel int
 var errorsMap map[string]uint
+var errorsMapMutex sync.Mutex
 var errors uint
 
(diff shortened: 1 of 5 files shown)


dreadl0ck:
try again

lhausermann:
Hi @dreadlock would you double check that the CLA is signed with your email linked to your GitHub account? Thanks

dreadl0ck:
@lhausermann: The initial commit had an invalid email (I switched to a new computer), I tried to change it with git rebase but that didn't seem to work. I will delete my fork and try again.

PR #560, opened 2018-11-08
reassemblydump example: Fixed data races on errorsMap and tcpStream

dreadl0ck:
The reassemblydump example contains two data races:

The errorsMap is potentially accessed simultaneously by multiple goroutines.
A mutex was added to make it thread safe.

When using stream.run for both client and server for parsing http requests and responses,
both use a pointer to the parent stream to access the stream.urls array without any synchronization.
A mutex was also added to the tcpStream structure to prevent this.


gconnell:
Thanks!  Excellent bug finding/fixing!

PR #566, opened 2018-11-27
Add support for GRE Ack as used in PPTP

x-way:
PPTP adds an additional GRE header field which needs to be accounted for when computing the offset for the next-layer payload.


gconnell:
thanks!

PR #568, opened 2018-11-27
Move non-windows stuff to pcap_unix; Convert windows to implicit linking

notti:
This add support for npcap on windows.
Windows now supports both npcap and winpcap at the same time. If both
winpcap and npcap are installed at the same time, npcap is preferred.
Building on windows now doesn't need neither cgo nor npcap/winpcap sdks.

The way this works is that all the non-windows stuff is now in `pcap_unix.go`, while all the windows stuff is in `pcap_windows.go`. Those two files provide a go-only internal api to `pcap.go` which to the outside looks the same as before (except for the small change mentioned below).
The new windows code now doesn't use cgo (this is a pain to use on windows) and uses implicit linking (I couldn't get delay loading to work with cgo). => neither cgo nor npcap/winpcap sdk is needed. syscall.Syscall is used for all the calls and the `wpcap.dll` is loaded in `init()`. Since both npcap and winpcap provide a `wpcap.dll` with libpcap api - no distinctions between the two are necessary (except for the not supported api functions in older libpcap versions as before). The only place where npcap is used, is an additional ddlload path that is added before loading the actual `wpcap.dll`.

Gains: Now the pcap tests work on windows (only with npcap - with winpcap some of the tests fail due to outdatet libpcap (e.g. no pcapng support, ...))

API of pcap.go is the same as before except for one small change:
https://github.com/google/gopacket/blob/a3c4a85b4a4cecfbf39ebd0c6a892bd22f485978/pcap/pcap.go#L795

The function signature of this one seems to be wrong and slipped through the cracks (all related functions (e.g., `DatalinkValToName`) accept `int` as value and not `C.int`
Additionally, I think `C.*`-types should never be exposed and this seems like a mistake.
If this is bad since it might break something, I can change it back.

Ad performance: the new version has an additional function call in `getNextBufPtrLocked` since `pcapNextPacketEx` can't be inlined. But the new `ReadPacketData` is faster since it doesn't use `C.GoBytes` any longer (which is a call into C code! - I think this function is supposed to be used from within C and not go...)
Testing with the gopacket_benchmark was inconclusive.

Ad windows: Right now there is no support to select winpcap if both npcap and winpcap are installed... I don't know if there would be a good way to do that.


pcap/defs_windows_386.go
```diff
@@ -0,0 +1,67 @@
+// Code generated by cmd/cgo -godefs; DO NOT EDIT.
```
gconnell: is it possible to add copyright notices to the godef files?  if not, no worries.

pcap/gen_defs_windows.go
```diff
@@ -0,0 +1,40 @@
+// +build ignore
```
gconnell: copyright here

pcap/generate_windows.bat
```diff
@@ -0,0 +1,5 @@
+SET GOARCH=386
```
gconnell: copyright here

pcap/gen_defs_windows.go
```diff
@@ -0,0 +1,40 @@
...
+
```
gconnell: add a comment up here detailing what godefs does and that the .bat file should be run to take advantage of this file?

pcap/generate_windows.bat
```diff
@@ -0,0 +1,5 @@
+SET GOARCH=386
+go tool cgo -godefs -- -I C:\npcap-sdk-1.01\Include gen_defs_windows.go | gofmt > defs_windows_386.go
+SET GOARCH=amd64
+go tool cgo -godefs -- -I C:\npcap-sdk-1.01\Include gen_defs_windows.go | gofmt > defs_windows_amd64.go
+SET GOARCH=
```
gconnell: add newline to end of file for general cleanliness?

pcap/pcap.go
```diff
@@ -349,16 +225,16 @@ func (p *Handle) ReadPacketData() (data []byte, ci gopacket.CaptureInfo, err err
...
 
 const (
-	aeNoError      = 0
-	aeActivated    = C.PCAP_ERROR_ACTIVATED
-	aePromisc      = C.PCAP_WARNING_PROMISC_NOTSUP
-	aeNoSuchDevice = C.PCAP_ERROR_NO_SUCH_DEVICE
-	aeDenied       = C.PCAP_ERROR_PERM_DENIED
-	aeNotUp        = C.PCAP_ERROR_IFACE_NOT_UP
-	aeWarning      = C.PCAP_WARNING
+	aeNoError      = activateError(0)
```
gconnell: are these necessary anymore, or should we just use pcapBlah?

gconnell: can we make the pcapBlah things be an error type?

pcap/pcap_unix.go
```diff
@@ -41,12 +130,482 @@ int pcap_wait(pcap_t *p, int usec) {
...
+		return p.pcapGeterr()
+	}
+	return nil
+}
+
+func (p *Handle) pcapListDatalinks() (datalinks []Datalink, err error) {
+	var dltbuf *C.int
+
+	n := int(C.pcap_list_datalinks(p.cptr, &dltbuf))
+	if -1 == n {
```
gconnell: anywhere you currently check for `blah == -1`, can you change it to `blah < 0`?  It looks like sometimes there are other errors returned than PCAP_ERROR, and `<0` seems safer.  For example, `pcap_list_datalinks` can return `PCAP_ERROR_NOT_ACTIVATED` per https://nmap.org/npcap/guide/wpcap/pcap_list_datalinks.html, and that's -3.

pcap/pcap_unix.go
```diff
@@ -41,12 +130,482 @@ int pcap_wait(pcap_t *p, int usec) {
...
+	}
+}
+
+func pcapCreate(device string) (*InactiveHandle, error) {
+	buf := (*C.char)(C.calloc(errorBufferSize, 1))
+	defer C.free(unsafe.Pointer(buf))
+	dev := C.CString(device)
+	defer C.free(unsafe.Pointer(dev))
+
+	// This copies a bunch of the pcap_open_live implementation from pcap.c:
```
gconnell: is this comment still true?  it seems like we only call pcap_create here, not any other pcap_open_live stuff.

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+	"time"
 	"unsafe"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+const npcapPath = "\\Npcap"
+
+func initDllPath() {
```
gconnell: can this function store the path it ends up using and supply that in the errors returned by mustLoad/mightLoad to be more specific about exactly which (full) paths it's having trouble loading things from?

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+	for i := range bval {
+		if bval[i] == 0 {
+			return string(bval[:i])
+		}
+	}
+	return string(bval[:])
+}
+
+// bytePtrToString returns a string copied from pointer to a null terminated byte array
+// WARNING: ONLY SAFE WITH IF r POINTS TO C MEMORY!
```
gconnell: I see you like to live dangerously ;)

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+	return bpfInstruction
+}
+
+func pcapBpfProgramFromInstructions(bpfInstructions []BPFInstruction) pcapBpfProgram {
+	var bpf pcapBpfProgram
+	bpf.Len = uint32(len(bpfInstructions))
+	cbpfInsns, _, _ := syscall.Syscall(callocPtr, 2, uintptr(len(bpfInstructions)), uintptr(unsafe.Sizeof(bpfInstructions[0])), 0)
+
+	//SMELL: this looks strange:
+	copy((*[bpfInstructionBufferSize]BPFInstruction)(unsafe.Pointer(cbpfInsns))[0:len(bpfInstructions)], bpfInstructions)
```
gconnell: maybe break this out into a for loop so there's a lower ratio of magic per line?

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+}
+
+func pcapLookupnet(device string) (netp, maskp uint32, err error) {
+	buf := make([]byte, errorBufferSize)
+	dev, err := syscall.BytePtrFromString(device)
+	if err != nil {
+		return 0, 0, err
+	}
+	e, _, _ := syscall.Syscall6(pcapLookupnetPtr, 4, uintptr(unsafe.Pointer(dev)), uintptr(unsafe.Pointer(&netp)), uintptr(unsafe.Pointer(&maskp)), uintptr(unsafe.Pointer(&buf[0])), 0, 0)
+	if pcapCint(e) == -1 {
```
gconnell: as commented elsewhere, change these to `< 0`?

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+	n := int(pcapCint(ret))
+
+	if n == -1 {
+		return nil, p.pcapGeterr()
+	}
+	defer syscall.Syscall(pcapFreeDatalinksPtr, 1, uintptr(unsafe.Pointer(dltbuf)), 0, 0)
+
+	datalinks = make([]Datalink, n)
+
+	dltArray := (*[100]pcapCint)(unsafe.Pointer(dltbuf))
```
gconnell: maybe pop a log.Error here if `n > 100`?

pcap/pcap_windows.go
```diff
@@ -7,18 +7,712 @@
...
+	}
+	return nil
+}
+
+func (p *Handle) pcapSnapshot() int {
+	ret, _, _ := syscall.Syscall(pcapSnapshotPtr, 1, uintptr(p.cptr), 0, 0)
+	return int(pcapCint(ret))
+}
+
+func (t TimestampSource) pcapTstampTypeValToName() string {
```
gconnell: make this return (string, error)?

pcap/defs_windows_386.go
notti: not directly (cgo removes all comments :( )
But I reworked code generation to not use a bat file (now the whole process is more extensible) - now the generated ones have copyright notices

pcap/gen_defs_windows.go
notti: done (now in generate_defs.go)

pcap/generate_windows.bat
notti: done (now in generate_defs.go)

pcap/gen_defs_windows.go
notti: done (now in generate_defs.go)

pcap/pcap_windows.go
notti: Sadly there is no other way for doing this :( (and all the other implementations do the same thing)

pcap/pcap_windows.go
notti: Not directly, but possible via kernel32.GetModuleFileName; Added this code (with fallback to old behaviour if this doesn't work)

pcap/pcap_unix.go
notti: Hmm I think not -> removed it.

pcap/pcap_unix.go
notti: fixed

pcap/pcap_windows.go
notti: all occurrences fixed

pcap/pcap_windows.go
notti: changed it to `1 << 30` as used elsewhere - maximum number guarded by n anyway (this is just to treat the c array as a go array so we can index it)

pcap/pcap_windows.go
notti: not an option since the public interface also just returns string - this would change the gopacket api :(

pcap/pcap_windows.go
notti: done

gconnell:
Looks great!  Merging in now...

PR #569, opened 2018-11-29
Add support for HDLC-like PPP as used in PPTP

x-way:
PPP frames when used in PPTP start with the Address/Control byte sequence 0xff 0x03 which must be recognized and properly handled when decoding the frames.

With this change gopacket is able to properly decode PPTP encapsulated traffic: https://www.cloudshark.org/captures/1073471675b4


gconnell:
Would you be able to add a test case for this?

gconnell:
Awesome, thanks!

PR #570, opened 2018-12-04
Fix length in DHCPv6Option when FixLenghts is true

cimnine:
Previously, `gopacket.SerializeOptions{FixLengths: true}` would not have any effect on the lengths in `layers.DHCPv6Option.encode()`. This has been fixed.


gconnell:
Thanks!

PR #571, opened 2018-12-11
 Add ZeroCopyReadPacketData to pcapgo.Reader 

WGH-:
Like ZeroCopyReadPacketData in pcapco.NgReader, it's not true zero copy, as data is still copied from the io.Reader. Still, memory allocation for every packet is avoided.


WGH-:
I signed it!

gconnell:
Neat, thanks!

gconnell:
And thanks for your first GoPacket code contribution!  Really appreciated :)

PR #574, opened 2018-12-17
added support for sflow ethernet frame flows

jpbede:
Added support for sFlow ethernet frame flows


jpbede:
I've added now a test with data from a live network. 
I've also added a field that i've forgotten.

notti:
Thanks for the contribution.

PR #578, opened 2018-12-20
reassembly: small optimizations to overlapExisting

jandos:
Reassembly's overlapExisting function had 2 unnecessary operations

- Adding `diff` to `start` when we already know that result will be same as `half.nextSeq`
- Subtracting `diff` from a variable `e` which is never used later


notti:
Looks good. Thanks

PR #579, opened 2018-12-21
pcap: use C.struct, not _Ctype_struct

ianlancetaylor:
The future Go 1.12 version of the cgo tool will issue an error for Go
code that uses names like _Ctype_struct_S, as that is in the namespace
that cgo uses for its own generated code. Change this code to use the
supported form C.struct_S, so that it builds with the future 1.12.

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -208,7 +208,7 @@ type InterfaceAddress struct {
 // BPF is a compiled filter program, useful for offline packet matching.
 type BPF struct {
 	orig string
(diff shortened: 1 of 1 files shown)


notti:
@ianlancetaylor Since #568 was merged this code moved to a different location. Would you be interested in updating your pull-request?

ianlancetaylor:
I have no idea how to correctly update the pull request, so I will recreate it.

ianlancetaylor:
See https://github.com/google/gopacket/pull/589 .

PR #580, opened 2018-12-23
sFlow decode when enterprise id is 0

jpbede:
Detected some unexpected behavior with e.g. pmacct

pmacct is using enterpriseID 43874 and flowRecordType 2 for something different which causes problems. 

Try to decode sflow records, only when enterprise id is 0, which signals that it using the default structs according to the specification.


notti:
Thanks for the contribution

PR #584, opened 2018-12-30
Add FTP layer

x-way:


notti:
See #587 for tcp discussion

PR #589, opened 2019-01-04
pcap: use C.struct, not _Ctype_struct

ianlancetaylor:
The future Go 1.12 version of the cgo tool will issue an error for Go
code that uses names like _Ctype_struct_S, as that is in the namespace
that cgo uses for its own generated code. Change this code to use the
supported form C.struct_S, so that it builds with the future 1.12.

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -368,7 +368,7 @@ func (p *Handle) ListDataLinks() (datalinks []Datalink, err error) {
 	return p.pcapListDatalinks()
 }
 
-// compileBPFFilter always returns an allocated _Ctype_struct_bpf_program
+// compileBPFFilter always returns an allocated C.struct_bpf_program
 // It is the callers responsibility to free the memory again, e.g.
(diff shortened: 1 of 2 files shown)


notti:
Looks good. Thanks

gconnell:
Thanks, Ian :)

mkumatag:
@notti when can I expect this in the release? is there any plan for a new release?

gconnell:
I'm super bad at releases, so they tend to be on-request.

@mkumatag happy to mark HEAD as a release, if you'd like.
@notti always feel free to create a new release, if you'd like.

notti:
@gconnel Was just about to ask that :)
Well the only thing I wanted to wait for was for feedback for the unix/windows split in pcap - but looks like the fallout is over (had two bug reports, which I resolved) - so I'll make a new release a bit later in the evening.

notti:
@mkumatag Just tagged a new release

PR #590, opened 2019-01-06
Expose SLL AddrType (link-layer address type)

x-way:
This corresponds to the sll_hatype field in the sll_header struct of libpcap: https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/sll.h#L85
It's the last field of the sll_header struct which was not exposed in LinuxSLL


gconnell:
Thanks!

PR #602, opened 2019-01-14
Avoid a cpu spin when NextPacket() returns EBADF

hdm:
The NextPacket() function can return EBADF when a gopcap EthernetHandle has been closed (triggered by the raw SetTimeout()). The packetsToChannel() function should probably break on any error, but I figured there was a reason for its current behavior.


hdm:
I signed it!

gconnell:
Thanks!

PR #606, opened 2019-01-28
handle error in documentation code

florianl:

This PR updates the documentation, so that the returned error is also handled.


gconnell:
Thanks!

PR #608, opened 2019-02-01
Pcap bpf boost

yerden:
`hdr` struct instance in `pcapOfflineFilter()` escapes through Cgo call. Preclude pointer escaping by placing `pcapPkthdr` inside `BPF` struct, thus allocating it on the heap and by performing pointer conversions.

Also added a benchmark to test this.

```
benchmark                             old ns/op     new ns/op     delta
BenchmarkPcapNonmatchingBPFFilter     999           756           -24.32%
BenchmarkPcapMatchingBPFFilter        1113          884           -20.58%

benchmark                             old allocs     new allocs     delta
BenchmarkPcapNonmatchingBPFFilter     1              0              -100.00%
BenchmarkPcapMatchingBPFFilter        1              0              -100.00%

benchmark                             old bytes     new bytes     delta
BenchmarkPcapNonmatchingBPFFilter     32            0             -100.00%
BenchmarkPcapMatchingBPFFilter        32            0             -100.00%
```

### DIFF
--- a/pcap/bpf_test.go
+++ b/pcap/bpf_test.go
@@ -1,0 +1,70 @@
+// Copyright 2018 The GoPacket Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+package pcap
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/gopacket"
+	"github.com/google/gopacket/layers"
+)
+
+var (
+	snaplen = 65535
+	packet  = [...]byte{
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, // dst mac
+		0x0, 0x11, 0x22, 0x33, 0x44, 0x55, // src mac
+		0x08, 0x0, // ether type
+		0x45, 0x0, 0x0, 0x3c, 0xa6, 0xc3, 0x40, 0x0, 0x40, 0x06, 0x3d, 0xd8, // ip header
+		0xc0, 0xa8, 0x50, 0x2f, // src ip
+		0xc0, 0xa8, 0x50, 0x2c, // dst ip
+		0xaf, 0x14, // src port
+		0x0, 0x50, // dst port
+	}
+	matchingBPFFilter    = "ip and tcp and port 80"
+	nonmatchingBPFFilter = "udp and port 80"
+)
+
+func BenchmarkPcapNonmatchingBPFFilter(b *testing.B) {
+	bpf, err := NewBPF(layers.LinkTypeEthernet, snaplen, nonmatchingBPFFilter)
+	if err != nil {
+		b.Fatal("incorrect filter")
+	}
+
+	ci := gopacket.CaptureInfo{
+		InterfaceIndex: 0,
+		CaptureLength:  len(packet),
+		Length:         len(packet),
+		Timestamp:      time.Now(),
+	}
+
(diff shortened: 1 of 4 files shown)


pcap/bpf_test.go
```diff
@@ -0,0 +1,65 @@
+package pcap
```
gconnell: please add copyright notice.  See the one on https://github.com/google/gopacket/blob/master/layers/tls_appdata.go for an easily copy-paste-able version.  Feel free to add your name to the AUTHORS file at the top-level as well if you'd like.

pcap/bpf_test.go
```diff
@@ -0,0 +1,65 @@
...
+		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, // dst mac
+		0x0, 0x11, 0x22, 0x33, 0x44, 0x55, // src mac
+		0x08, 0x0, // ether type
+		0x45, 0x0, 0x0, 0x3c, 0xa6, 0xc3, 0x40, 0x0, 0x40, 0x06, 0x3d, 0xd8, // ip header
+		0xc0, 0xa8, 0x50, 0x2f, // src ip
+		0xc0, 0xa8, 0x50, 0x2c, // dst ip
+		0xaf, 0x14, // src port
+		0x0, 0x50, // dst port
+	}
+	goodBPFFilter = "ip and tcp and port 80"
```
gconnell: possibly rename "matching" and "nonmatching", since the "bad" filter itself is actually a perfectly good filter that just doesn't match the above packet :)

pcap/pcap_unix.go
```diff
@@ -343,12 +347,15 @@ func pcapLookupnet(device string) (netp, maskp uint32, err error) {
 
 func (b *BPF) pcapOfflineFilter(ci gopacket.CaptureInfo, data []byte) bool {
 	var hdr C.struct_pcap_pkthdr
+	p := uintptr(unsafe.Pointer(&hdr))
```
notti: This can cause some trouble:
According to https://golang.org/pkg/unsafe/ this is illegal to do, because now the compiler is free to think, that you're actually not using `hdr`, after line 354, which can lead to surprising results.

If one reads the documentation closely, one would discover that this is only valid for syscalls (actually functions defined in assembly) then there is a second issue:
Between converting this pointer to `uintptr` and actually accessing it, go might change the stack (stack split etc). But with this construct go doesn't know about the pointer any longer, so it can't adjust it - leading `pcap_offline_filter_escaping` to point to the wrong memory. Actually cgo tries to lock down the stack, but this takes a bit (the cgo call below actually generates new function that contains a pointer check and a call to another autogenerated function that itself calls `cgocall`, which in turn then locks down the stack after some housekeeping) - if we hit this in the small window surprising stuff will happen..
So this can fail in surprising ways.

Solution:
The same thing is also in `pcap_next_ex_escaping` - but there we reference stuff from `*Handle`, which is guaranteed to life on the heap due to the way it is created with e.g. `pcapOpenLive` (there `Handle` escapes to the heap). So still not safe according to the documentation, but at least safe from the things that can happen.

So correct way would be to put the hdr into *BPF, where it is guaranteed to live on the heap.
To keep compatibility with windows this must be done as `pcapPkthdr` which is a typedef to `C.struct_pcap_pkthdr`.

yerden: Thanks for the explanation and suggestion. I'll handle it.

notti:
Thanks for the patch.

PR #612, opened 2019-02-04
Soft fail on missing WinPCAP during init()

hdm:
This PR changes the behavior of pcap.init() on Windows to no longer panic if WinPCAP is not found. This allows the caller to install WinPCAP [or npcap] on the fly. A caller that wishes to maintain the previous panic() behavior can add the following code to their package:
```
func init() {
	err := pcap.LoadWinPCAP()
	if err != nil {
		panic(err)
	}
}
```
The downside of this change is that callers that don't check LoadWinPCAP() may crash if they call a pcap function and WinPCAP is not available. This seems mostly harmless, as the program would have crashed either way, and the new behavior offers more control over this process.

### DIFF
--- a/pcap/pcap_windows.go
+++ b/pcap/pcap_windows.go
@@ -21,6 +21,8 @@ import (
 	"github.com/google/gopacket/layers"
 )
 
+var pcapLoaded = false
+
 const npcapPath = "\\Npcap"
 
 func initDllPath(kernel32 syscall.Handle) {
@@ -152,9 +154,18 @@ var (
 )
 
 func init() {
+	LoadWinPCAP()
+}
+
+// LoadWinPCAP attempts to dynamically load the wpcap DLL and resolve necessary functions
+func LoadWinPCAP() error {
+	if pcapLoaded {
+		return nil
+	}
+
 	kernel32, err := syscall.LoadLibrary("kernel32.dll")
 	if err != nil {
-		panic("couldn't load kernel32.dll")
+		return fmt.Errorf("couldn't load kernel32.dll")
 	}
 	defer syscall.FreeLibrary(kernel32)
 
@@ -162,16 +173,16 @@ func init() {
 
 	wpcapHandle, err = syscall.LoadLibrary("wpcap.dll")
 	if err != nil {
-		panic("Couldn't load wpcap.dll")
+		return fmt.Errorf("couldn't load wpcap.dll")
 	}
 	initLoadedDllPath(kernel32)
 	msvcrtHandle, err = syscall.LoadLibrary("msvcrt.dll")
 	if err != nil {
-		panic("Couldn't load msvcrt.dll")
+		return fmt.Errorf("couldn't load msvcrt.dll")
 	}
 	callocPtr, err = syscall.GetProcAddress(msvcrtHandle, "calloc")
 	if err != nil {
-		panic("Couldn't get calloc function")
+		return fmt.Errorf("couldn't get calloc function")
 	}
 
 	pcapStrerrorPtr = mustLoad("pcap_strerror")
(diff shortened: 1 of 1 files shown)


notti:
> This PR changes the behavior of pcap.init() on Windows to no longer panic if WinPCAP is not found. This allows the caller to install WinPCAP [or npcap] on the fly. A caller that wishes to maintain the previous panic() behavior can add the following code to their package:

This would change behavior for existing programs being the reason why I wrote it that way.

> The downside of this change is that callers that don't check LoadWinPCAP() may crash if they call a pcap function and WinPCAP is not available. This seems mostly harmless, as the program would have crashed either way, and the new behavior offers more control over this process.

But now the panic will be some kind of Segmentation Fault (don't have windows lying around right now to check what exactly happens if you `CALL` address 0) instead of an informative message what's missing.

hdm:
Thanks for the comment. Would adding specific checks to each calling function resolve this for you? Something like: 

```
if pcapDatalinkPtr == 0 {
    panic("WinPCAP not loaded");
}
```

hdm:
An alternative to checking each ptr would be to be add a LoadWinPCAP() check to every function. It would be easier to maintain consistency at least. An example would be:
```
func SomePcapFunction(...) {
    verifyWinPCAP()
...
}

func verifyWinPCAP() {
    err := LoadWinPCAP()
    if err != nil { panic(err) }
    return
}
```

notti:
> An alternative to checking each ptr would be to be add a LoadWinPCAP() check to every function. It would be easier to maintain consistency at least. An example would be:
> 
> ```
> func SomePcapFunction(...) {
>     verifyWinPCAP()
> ...
> }
> 
> func verifyWinPCAP() {
>     err := LoadWinPCAP()
>     if err != nil { panic(err) }
>     return
> }
> ```

Yes, that sounds better.
But I don't think we need the check in every function, since not everything can be called without calling something else first, e.g., `handle.ReadPacketData` can't be called without having the `handle` first - so `ReadPacketData` shouldn't need one, but `OpenLive`, `OpenOffline`, and `OpenOfflineFile` do. This also helps with performance, since the hotpath `ReadPacketData` stays the same while the `Open*` functions are called only a low number of times anyway.

hdm:
Got it. I'll update this PR shortly. Thanks!

hdm:
This update tweaks `pcapOpenLive`, `openOffline`, `openOfflineFile`, and `pcapOpenDead` to return a nil handle and error if WinPCAP has not been loaded. The `pcapLookupnet` and `pcapFindAllDevs` functions have been updated to also return an error if WinPCAP has not been loaded. Finally, the `pcapLibVersion` function has been updated to panic if WinPCAP has not been loaded. This should cover the main entry points into this package.

hdm:
The test cases are failing appropriately when winpcap/npcap is not available:
```
gopacket\pcap>go test
--- FAIL: TestPcapFileRead (0.00s)
    pcap_test.go:86:



        Processing file test_loopback.pcap



    pcap_test.go:95: couldn't load wpcap.dll
--- FAIL: TestBPF (0.00s)
    pcap_test.go:123: couldn't load wpcap.dll
--- FAIL: TestBPFInstruction (0.00s)
    pcap_test.go:156: couldn't load wpcap.dll
--- FAIL: TestPCAPGoWrite (0.00s)
    pcapgo_test.go:43: couldn't load wpcap.dll
--- FAIL: TestPCAPGoNgWrite (0.00s)
    pcapnggo_test.go:47: couldn't load wpcap.dll
2019/02/17 10:49:51 couldn't load wpcap.dll
exit status 1
FAIL    github.com/hdm/gopacket/pcap    0.032s
```

hdm:
Thanks! These changes have been made.

notti:
Thanks for the contribution

PR #622, opened 2019-03-06
afpacket: use single implementation of tpacket alignment function

tklauser:
There are two tpacket alignment functions, tpacketAlign and tpAlign. The
former has a bug (i.e. it doesn't match TPACKET_ALIGN from
linux/if_packet.h), so only use the latter. Also make tpacketAlignment a
const.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -42,17 +42,12 @@ import (
 import "C"
 
 var pageSize = unix.Getpagesize()
-var tpacketAlignment = uint(C.TPACKET_ALIGNMENT)
 
 // ErrPoll returned by poll
 var ErrPoll = errors.New("packet poll failed")
 
 // ErrTimeout returned on poll timeout
 var ErrTimeout = errors.New("packet poll timeout expired")
-
-func tpacketAlign(v int) int {
-	return int((uint(v) + tpacketAlignment - 1) & ((^tpacketAlignment) - 1))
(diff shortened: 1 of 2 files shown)


tklauser:
The Travis CI failure is due to mdlayher/raw@f434146a17ecdf185e28b9ae8457ef45ee82077a which requires Go 1.12 to work.

notti:
Yeah unfortunately mdlayher/raw does only track the latest version. But pcapgo doesn't really need that library (it doesn't use any features from there exept opening a socket and then reading).
I'm nearly finished (tomorrow) getting rid of the dependency and improving the implementation a lot (e.g. added getting timestamps, captured length, ...) - then this problem is solved

stapelberg:
Is there an easy fix for my use-case?

notti:
@stapelberg I added the missing functionality in b2eab7df3513b927f00c9eba914f62f1f47a8708 and router7 should work again.

Please open an Issue the next time - this makes tracking these things easier.

stapelberg:
> @stapelberg I added the missing functionality in [b2eab7d](https://github.com/google/gopacket/commit/b2eab7df3513b927f00c9eba914f62f1f47a8708) and router7 should work again.

Thanks, I can confirm that it builds again, will do a manual test soon.

> Please open an Issue the next time - this makes tracking these things easier.

Yep, will do.

Issue #163, opened 2016-01-20
packetsToChannel() should propogate error codes


barclayo:
packetsToChannel() explicitly ignores errors from NextPacket(), on Windows this manifests with the following behavior:  if the device is reading via a live interfaces, and the device sleeps then wakes, any reads on the previously open handle will result in a ReadError.  Since this is distinct from io.EOF, the binary will spin within Packets() and eat a full core.

Workaround:  write your own Packets() convenience routine and handle read errors as appropriate.

notti:
I'm not completely sure what the reason was for ignoring all errors except io.EOF. Maybe it would be best to special handle the exact error that would happen.

notti:
Maybe fixed by #625 
If there is an error we overlooked, it simply can be added to the list now.

PR #625, opened 2019-03-16
Handle ErrFileClosing in packetsToChannel()

hdm:
In addition to EBADF, some distributions (Ubuntu 16.04 LTS at least) return the error `internal/poll.ErrFileClosing` in this function, causing an infinite loop if not handled. Since this is an internal error, the accepted way to check for it appears to be a string match. This is similar to the fix in PR #602. 

It would be great if the ignored errors were a whitelist versus a blacklist; what errors are supposed to be ignored here?


hdm:
The last push takes a different approach to this and now bails from the loop if more than 100 errors are encountered in a row. If you prefer a different way to handle this, please let me know!

notti:
This is related to #163
I could not really figure out what the reason was for this mechanism (stuff like EINTR reaching this point?). I think it would be better to recover only in known situations where an immediate retry solves the problem.

hdm:
Would you like this updated to only retry once?

notti:
Well more like fail directly in situations that are clearly unrecoverable (EOF and EBADF sounds like that).
Retry should be ok for situations that might recover.
What I meant was that it would be better to have a list of temporary failures - retry for those - and fail for all others. However, I think it's not possible to come up with such a list.
So:
- EOF, BADF (or is there a case where we get BADF in a temporary situation?) fail now
- others try and fail later
Maybe wait a bit before retrying?

hdm:
Thanks! I'll take another swing at this soon.

hdm:
@notti Could you take a peek at the latest when you get a chance?

notti:
Thanks for implementing the suggested changes.

PR #626, opened 2019-03-26
Support for EDNS (RFC6891)

elchavar:


layers/dns.go
```diff
@@ -904,6 +959,16 @@ type DNSMX struct {
 	Name       []byte
 }
 
+// DNSOPT is a DNS Option, see RFC6891, section 6.1.2
+type DNSOPT struct {
+	Code uint16
```
gconnell: should this be an enum, maybe "DNSOptionCode"?

elchavar: I can make it an enum and include the values whose status is "Standard" or "Optional" in https://www.iana.org/assignments/dns-parameters/dns-parameters.xhtml#dns-parameters-11

Sounds reasonable?

gconnell:
Sounds good to me :)

layers/dns.go
```diff
@@ -816,6 +842,29 @@ func decodeCharacterStrings(data []byte) ([][]byte, error) {
...
+
+	if offset+4 > end {
+		return allOPT, fmt.Errorf("DNSOPT record is of length %d, it should be at least length 4", end-offset)
+	}
+
+	for i := offset; i < end; {
+		opt := DNSOPT{}
+		opt.Code = DNSOptionCode(binary.BigEndian.Uint16(data[offset : offset+2]))
+		l := binary.BigEndian.Uint16(data[offset+2 : offset+4])
+		opt.Data = data[offset+4 : offset+4+int(l)]
```
notti: This is missing a length check

elchavar: Added the length check and a corresponding test in dns_test.go

PR #627, opened 2019-03-28
fix: unix.Timespec/Timeval member type various

boypt:
`unix.Timespec`/`unix.Timeval` members are `int32` in 32-bit arch
 - https://github.com/golang/sys/blob/master/unix/ztypes_linux_arm.go#L24
 - https://github.com/golang/sys/blob/master/unix/ztypes_linux_386.go#L24

But `time.Unix` always accept int64 arguments. Type casting is needed.


notti:
@boypt Any chance on signing the CLA?

boypt:
@googlebot I signed it!

boypt:
@notti CLA signed, tks

notti:
Thanks for the fix.

PR #633, opened 2019-04-07
pcap: correct NewBPF godoc to display code as codeblock

arnottcr:
Fixes golang/gddo#596


notti:
Indeed, it does. Thanks for the fix.

vixie:
ty

PR #634, opened 2019-04-10
Switch router interfaces to map

### LINKED ISSUE #222: routing api should support out of order interfaces
i don't think this should be returning an error here
https://github.com/google/gopacket/blob/master/routing/routing.go#L214-L216

kulinacs:
Switch router interfaces to a map to support discontinuous
interface ids. Fixes #222

Will admit, Line 107 is messy. Wasn't sure how to easily return a pointer from a map. Added some tests to verify routing functionality works as expected as well.


kulinacs:
I signed it!

notti:
> Will admit, Line 107 is messy. Wasn't sure how to easily return a pointer from a map. Added some tests to verify routing functionality works as expected as well.

How about storing pointers to interfaces instead of interfaces? This way there wouldn't be memory allocations upon querying interfaces.

kulinacs:
> How about storing pointers to interfaces instead of interfaces? This way there wouldn't be memory allocations upon querying interfaces.

I had tried that but was having issues. I'll try again over the weekend and report back with the results.

notti:
@kulinacs any news?

kaminek:
up ?

kaminek:
@notti couldn't copy the ifaces pointer because it's a map and you cannot take the address of `r.ifaces[ifaceIdex]`. Any idea on how to work it around?

notti:
Well first idea that comes to my mind is to store pointers in the map instead of the objects themselves.

kulinacs:
@notti updated. Looks like the windows build is erroring out installing npcap.

notti:
This code still does an allocation per query - is this really necessary?

kulinacs:
Allocation removed :grin:

Capstan:
Is this something that can be merged and submitted? Folks are finding #830 instead of #222 and still need this fixed.

gconnell:
thanks!

PR #635, opened 2019-04-10
Added CanDecode() method for EAPOLKey

Galadros:
Added CanDecode() method for EAPOLKey, which previously lacked that method


notti:
Thanks for the fix.

Issue #638, opened 2019-04-19
Panic encoding name in dns.go


bmeadors:
A panic happens while trying to serialize a DNS packet having authority records where the Name field is empty (<Root>).

The issue seems to be that the computeSize function used to determine the needed size of the buffer handles these empty Name fields slightly different than the encode function used to write data to said allocated buffer.  This causes the write offset to incorrectly grow by 1 byte for each authority record written and ends up causing the index out of range panic below.

panic: runtime error: index out of range

goroutine 15 [running]:
github.com/google/gopacket/layers.encodeName(0xc0001a45a0, 0x12, 0x60, 0xc0001aa000, 0x2a7, 0x2a7, 0x29f, 0x295)
	/root/go/src/github.com/google/gopacket/layers/dns.go:724 +0xe8
github.com/google/gopacket/layers.(*DNSResourceRecord).encode(0xc0000cd940, 0xc0001aa000, 0x2a7, 0x2a7, 0x294, 0x1, 0x20, 0x0, 0x0)
	/root/go/src/github.com/google/gopacket/layers/dns.go:754 +0x8dc
github.com/google/gopacket/layers.(*DNS).SerializeTo(0xc000142780, 0x9cc500, 0xc0000871d0, 0xc000030001, 0x40da68, 0x50)
	/root/go/src/github.com/google/gopacket/layers/dns.go:506 +0x5d5
github.com/google/gopacket.SerializeLayers(0x9cc500, 0xc0000871d0, 0x1, 0xc000049f90, 0x1, 0x1, 0x0, 0x0)
	/root/go/src/github.com/google/gopacket/writer.go:210 +0xcd
main.(*DnsContext).Run.func1.2(0xc0000224e0, 0xc000010090)

notti:
Fixed by #638

PR #639, opened 2019-04-19
#638: Fix for serializing DNS packet panic with empty authority name

bmeadors:
Addresses submitted issue #638.

A recent commit (https://github.com/google/gopacket/commit/65567310658fb392f4408da7d375b10e7343eb83#diff-4d9bce33bc5ccd203264657a890010dc) computeSize in dns.go adds an 11 or 12 byte offset to account for packet fields based on if the resource record Name field is empty or not.  This function is used to calculate the required destination buffer size for the serialized data.  However when encoding the data during serialization, the encode function in dns.go would always report it had consumed 12 bytes (1 + 11).  This was causing the caller of the encode function to incorrectly advance the offset by 1 extra byte each time an authority resource record was encoded having an empty Name field and would eventually cause a panic for an "index out of range" condition.

A simple adjustment was made to the code to ensure the number of consumed bytes being returned is correct in all cases to prevent the panic.

### DIFF
--- a/layers/dns.go
+++ b/layers/dns.go
@@ -740,6 +740,7 @@ func encodeName(name []byte, data []byte, offset int) int {
 func (rr *DNSResourceRecord) encode(data []byte, offset int, opts gopacket.SerializeOptions) (int, error) {
 
 	noff := encodeName(rr.Name, data, offset)
+	nSz := noff - offset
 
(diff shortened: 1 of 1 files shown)


notti:
Thanks for the fix.

PR #642, opened 2019-04-23
Fix sflow portname record decoding

safchain:


layers/sflow.go
```diff
@@ -696,8 +696,8 @@ const (
 	SFlowTypeLACPCounters               SFlowCounterRecordType = 7
 	SFlowTypeProcessorCounters          SFlowCounterRecordType = 1001
 	SFlowTypeOpenflowPortCounters       SFlowCounterRecordType = 1004
-	SFlowTypePORTNAMECounters           SFlowCounterRecordType = 1005
```
notti: I agree, that these names don't follow the conventions correctly, we can't change them now, since this might break existing implementations.

layers/sflow.go
```diff
@@ -2390,8 +2390,8 @@ func decodeOpenflowportCounters(data *[]byte) (SFlowOpenflowPortCounters, error)
 	return ofp, nil
 }
 
-//SFlowAppresourcesCounters  :  OVS_Sflow App Resources Counter ( 48 Bytes )
```
notti: See above

layers/sflow.go
```diff
@@ -2450,22 +2450,31 @@ func decodeOVSDPCounters(data *[]byte) (SFlowOVSDPCounters, error) {
 	return dp, nil
 }
 
-//SFlowPORTNAME  :  OVS-Sflow PORTNAME Counter Sampletype ( 20 Bytes )
```
notti: See above.

layers/sflow.go
safchain: Agreed, I suspected this could be an issue :) no problem to revert it

notti:
Thanks for the fix.

PR #643, opened 2019-04-25
Implement DecodingLayer for SIP

### LINKED ISSUE #637: SIP with NewDecodingLayerParser()
Why layers.SIP does not implement DecodingLayer interface?

I try this code
```
var sip layers.SIP
parser := gopacket.NewDecodingLayerParser(layers.LayerTypeEthernet, &eth, &ip4, &udp, &sip)
```
and get an error
```
cannot use &sip (type *layers.SIP) as type gopacket.DecodingLayer in argument to gopacket.NewDecodingLayerParser:
	*layers.SIP does not implement gopacket.DecodingLayer (missing CanDecode method)
```

Maybe author just forgot to write CanDecode method?

x-way:
Fixes #637

### DIFF
--- a/layers/sip.go
+++ b/layers/sip.go
@@ -233,6 +233,16 @@ func (s *SIP) Payload() []byte {
 	return s.BaseLayer.Payload
 }
 
+// CanDecode returns the set of layer types that this DecodingLayer can decode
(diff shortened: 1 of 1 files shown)


notti:
Thanks.

PR #645, opened 2019-05-03
afpacket: use constants from x/sys/unix instead of C headers

tklauser:
The golang.org/x/sys/unix package provides all these constants so there
is no need to use cgo to get them.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -450,7 +450,7 @@ func (h *TPacket) getTPacketHeader() header {
 
 func (h *TPacket) pollForFirstPacket(hdr header) error {
 	tm := int(h.opts.pollTimeout / time.Millisecond)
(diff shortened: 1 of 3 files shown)


notti:
Nice, Thanks.
With the introduction of golang/sys@a43fa875 every required afpacket header definition should also be there making it possible to get rid of complete cgo in afpacket (well except for htons - but this one is trivial to implement).

tklauser:
> With the introduction of golang/sys@a43fa87 every required afpacket header definition should also be there making it possible to get rid of complete cgo in afpacket (well except for htons - but this one is trivial to implement).

Thanks. That commit actually reminded me that I still had this PR lying around :slightly_smiling_face:

notti:
Yeah getting rid of cgo has been on my TODO list for a while... An unrelated project kicked of the golang/sys commit (and I neither didn't expect it to go in that fast nor thought it would be so simple to integrate)

PR #646, opened 2019-05-07
afpacket: use PACKET_FANOUT_* consts from x/sys/unix

tklauser:
Use the unix.PACKET_FANOUT_* constants to define FanoutType values and
add missing values.


notti:
Thanks.

PR #653, opened 2019-05-12
Implement RMCP layer type

gebn:
Adds support for the RMCP header, specified in section 3.2.2.2 of [ASF v2.0](https://www.dmtf.org/sites/default/files/standards/documents/DSP0136.pdf). I plan to implement the ASF (inner) layer very soon in another MR.

Passes `./gc`; any feedback much appreciated!

### DIFF
--- a/layers/rmcp.go
+++ b/layers/rmcp.go
@@ -137,7 +137,7 @@ func (r *RMCP) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOp
 		return err
 	}
 	bytes[0] = r.Version
-	// 1 byte reserved, 0x00
+	bytes[1] = 0x00
 	bytes[2] = r.Sequence
 	bytes[3] = bool2uint8(r.Ack)<<7 | uint8(r.Class) // thanks, BFD layer
 	return nil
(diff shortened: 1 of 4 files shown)


layers/rmcp.go
```diff
@@ -0,0 +1,157 @@
...
+
+// SerializeTo writes the serialized fom of this layer into the SerializeBuffer,
+// partially satisfying SerializableLayer.
+func (r *RMCP) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
+	bytes, err := b.PrependBytes(4)
+	if err != nil {
+		return err
+	}
+	bytes[0] = r.Version
+	// 1 byte reserved, 0x00
```
notti: According to https://godoc.org/github.com/google/gopacket#SerializeBuffer buffers returned from PrependBytes and AppendBytes are not zeroed out - the 0x00 should be written here explicitly.

gebn:
I'm not sure why the Windows build failed... could you hit restart?

notti:
I'm also not sure what happened here. Travis is somehow not as reliable as I wish it would be.

notti:
Thanks for the implementation.

PR #654, opened 2019-05-19
Implement ASF and ASF Presence Pong layer types

gebn:
This builds on top of #653 to add support for the ASF RMCP message type. Presence Ping and Presence Pong are supported; the latter has a dedicated layer type for its payload.

Please merge #653 first; I'll then rebase this as necessary.

### DIFF
--- a/layers/asf_presencepong.go
+++ b/layers/asf_presencepong.go
@@ -146,13 +146,19 @@ func (a *ASFPresencePong) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.
 	if err != nil {
 		return err
 	}
+
 	binary.BigEndian.PutUint32(bytes[:4], a.Enterprise)
 	copy(bytes[4:8], a.OEM[:])
 	bytes[8] = bool2uint8(a.IPMI)&uint8(ASFPresencePongEntityIPMI) |
 		bool2uint8(a.ASFv1)&uint8(ASFPresencePongEntityASFv1)
 	bytes[9] = bool2uint8(a.SecurityExtensions)&uint8(ASFPresencePongInteractionSecurityExtensions) |
 		bool2uint8(a.DASH)&uint8(ASFPresencePongInteractionDASH)
-	// remaining 6 bytes all 0s
+
+	// zero-out remaining 6 bytes
+	for i := 10; i < len(bytes); i++ {
+		bytes[i] = 0x00
+	}
+
 	return nil
 }
 
--- a/layers/asf_presencepong.go
+++ b/layers/asf_presencepong.go
@@ -1,0 +1,161 @@
+// Copyright 2019 The GoPacket Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file in the root of the source tree.
+
+package layers
+
+// This file implements the RMCP ASF Presence Pong message, specified in section
+// 3.2.4.3 of
+// https://www.dmtf.org/sites/default/files/standards/documents/DSP0136.pdf. It
+// also contains non-competing elements from IPMI v2.0, specified in section
+// 13.2.4 of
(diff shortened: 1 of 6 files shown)


gebn:
This is now ready for review!

layers/asf_presencepong_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+
+	"github.com/google/gopacket"
+)
+
+func ASFPresencePongTestDecodeFromBytes(t *testing.T) {
```
notti: Test names *must* start with Test e.g. TestASFPresencePongDecodeFromBytes

layers/asf_presencepong_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	if pp.SecurityExtensions {
+		t.Errorf("do not want security extensions, got true")
+	}
+	if pp.DASH {
+		t.Errorf("do not want DASH, got true")
+	}
+}
+
+func ASFPresencePongTestSupportsDCMI(t *testing.T) {
```
notti: Wrong test name (see above)

layers/asf_presencepong_test.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+}
+
+func serializeASFPresencePong(pp *ASFPresencePong) ([]byte, error) {
+	sb := gopacket.NewSerializeBuffer()
+	err := pp.SerializeTo(sb, gopacket.SerializeOptions{})
+	return sb.Bytes(), err
+}
+
+func ASFPresencePongTestSerializeTo(t *testing.T) {
```
notti: Wrong test name (see above)

layers/asf_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+
+	"github.com/google/gopacket"
+)
+
+func ASFTestDecodeFromBytes(t *testing.T) {
```
notti: Wrong test name (see above)

layers/asf_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+func serializeASF(asf *ASF) ([]byte, error) {
+	sb := gopacket.NewSerializeBuffer()
+	err := asf.SerializeTo(sb, gopacket.SerializeOptions{
+		FixLengths: true,
+	})
+	return sb.Bytes(), err
+}
+
+func ASFTestSerializeTo(t *testing.T) {
```
notti: Wrong test name (see above)

layers/asf_presencepong.go
```diff
@@ -0,0 +1,181 @@
...
+// partially satisfying SerializableLayer.
+func (a *ASFPresencePong) SerializeTo(b gopacket.SerializeBuffer, _ gopacket.SerializeOptions) error {
+	bytes, err := b.PrependBytes(16)
+	if err != nil {
+		return err
+	}
+
+	binary.BigEndian.PutUint32(bytes[:4], a.Enterprise)
+	copy(bytes[4:8], a.OEM[:])
+	bytes[8] = bool2uint8(a.IPMI)&uint8(ASFPresencePongEntityIPMI) |
```
notti: This looks wrong to me (tests will reveal that as soon as they're executed due to the naming issue) -> bool2uint8 will return 1 if it is true, ASFPresencePongEntityIPMI = 1<<7 -> 1 & (1<<7) = 0

gebn: Indeed, for some reason I was expecting `bool2uint8()` to expand `true` to `0xFF`. I've now rewritten this in a longer but simpler way.

gebn:
Should’ve addressed everything. The naming blunder was caused by me getting carried away trying to avoid conflicts with other layers...

PR #657, opened 2019-05-28
parser: accelerate DecodeLayers

yerden:
Most of frequently used (and natively supported) LayerType values are not very big, so we may attempt to use array to search `DecodingLayer`, fallback to map only if `LayerType` is too big. Similar approach as in `RegisterLayerType`. I chose 256 as a decent threshold.

Results on Xeon(R) CPU E5-2620:
```
benchmark                                      old ns/op     new ns/op     delta
BenchmarkDecodingLayerParserIgnorePanic-32     227           191           -15.86%
BenchmarkDecodingLayerParserHandlePanic-32     291           227           -21.99%
```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -39,15 +39,18 @@ type DecodingLayer interface {
 	LayerPayload() []byte
 }
 
+const maxDecodingLayerType = 256
+
 // DecodingLayerParser parses a given set of layer types.  See DecodeLayers for
 // more information on how DecodingLayerParser should be used.
 type DecodingLayerParser struct {
 	// DecodingLayerParserOptions is the set of options available to the
 	// user to define the parser's behavior.
 	DecodingLayerParserOptions
-	first    LayerType
-	decoders map[LayerType]DecodingLayer
-	df       DecodeFeedback
+	first       LayerType
+	decoders    []DecodingLayer
+	decodersMap map[LayerType]DecodingLayer
+	df          DecodeFeedback
 	// Truncated is set when a decode layer detects that the packet has been
 	// truncated.
 	Truncated bool
@@ -58,6 +61,16 @@ type DecodingLayerParser struct {
 // encountered, they'll be parsed.
 func (l *DecodingLayerParser) AddDecodingLayer(d DecodingLayer) {
 	for _, typ := range d.CanDecode().LayerTypes() {
+		if typ >= maxDecodingLayerType {
+			// fallback to map
+			l.decodersMap[typ] = d
+			continue
+		}
+
+		if extra := int64(typ) - int64(len(l.decoders)) + 1; extra > 0 {
+			l.decoders = append(l.decoders, make([]DecodingLayer, extra)...)
+		}
+
 		l.decoders[typ] = d
 	}
 }
@@ -67,6 +80,20 @@ func (l *DecodingLayerParser) AddDecodingLayer(d DecodingLayer) {
 // DecodeLayers.
 func (l *DecodingLayerParser) SetTruncated() {
 	l.Truncated = true
+}
+
+func (l *DecodingLayerParser) lookupDecodingLayer(typ LayerType) (DecodingLayer, bool) {
+	if typ >= maxDecodingLayerType {
+		// fallback to map
+		d, ok := l.decodersMap[typ]
+		return d, ok
+	}
+
+	if int64(typ) >= int64(len(l.decoders)) {
+		return nil, false
+	}
+
+	return l.decoders[typ], l.decoders[typ] != nil
 }
 
 // NewDecodingLayerParser creates a new DecodingLayerParser and adds in all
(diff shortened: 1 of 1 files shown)


yerden:
And one more thing. I don't want to overgeneralize (and please correct me if i'm wrong) but I suspect that most common usage pattern here suggests not so many DecodingLayer-s in a single DecodingLayerParser. Most of the time, up to ten layers is enough.

If that's true, naive linear search of DecodingLayer beats `map`, my laptop (Core i5-3320M) tests:
```
BenchmarkSearchMap3-4       	30000000	        50.4 ns/op
BenchmarkSearchMap5-4       	30000000	        56.2 ns/op
BenchmarkSearchMap10-4      	20000000	        73.2 ns/op

BenchmarkSearchArray3-4     	30000000	        40.2 ns/op
BenchmarkSearchArray5-4     	30000000	        39.3 ns/op
BenchmarkSearchArray10-4    	30000000	        39.5 ns/op

BenchmarkSearchLinear3-4    	30000000	        45.2 ns/op
BenchmarkSearchLinear5-4    	30000000	        50.0 ns/op
BenchmarkSearchLinear10-4   	30000000	        52.8 ns/op
```

That's a definite win if some custom LayerType values are in use, so you don't want to use array indexing (too big values).

And, in case of a simple Eth->IPv4->TCP packet, about 60-70% of DecodeLayers() call amounts to searching for an appropriate DecodingLayer rather than layer decoding itself.

yerden:
I will investigate other ways to boost performance and will return later.

gconnell:
Hey, yerden,

This looked super promising; is there a reason the PR was closed out?

yerden:
Well, other ideas came up to me, like:

1. `DecodingLayerParser` doesn't consider level of a layer. For example, LayerTypeEthernet may encounter at the beginning only, rather than in middle of a packet, so when looking for the first level decoder, `layers.Ethernet` should be the primary suspect to test against.

2. As I said earlier, in case we have only few decoders, linear search is preferable to `map`.

3. Proposed approach is good when using gopacket-owned LayerType values but it doesn't work when working with custom LayerType values. Maybe we should implement different array values >1000.

4. Maybe we could provide end-user with a tool to supply his own map implementation for LayerType->DecodingLayer lookup. This would require to break API.

So... I though I should take a break, test these ideas and revisit later. Or, alternatively, I can simply reopen this PR, allow it to merge and work on it for next PR. Any ideas?

yerden:
Hello, everyone,

I've investigated various ways to gain some extra performance for DecodingLayerParser. In the course of it, I was pursuing these goals:
* avoid breaking API;
* avoid performance degradation in existing API;
* provide maximum flexibility to suit end-user's needs.

So, I ended up adding DecodingLayerContainer which stores DecodingLayer values and allows to address them. Various implementations of DecodingLayerContainer are also presented: DecodingLayerMap, DecodingLayerSparse, DecodingLayerArray.

DecodingLayerParser is refactored to use DecodingLayerContainer. DecodingLayerMap is used by default in current implementation as the most versatile one. Additionally, new decoding logic omits redundant lookup of the first LayerType value. From another side, additional level of indirection may impose some overhead.

Of course, the user may write his/her own DecodingLayerContainer implementation if necessary. Those in need may refer to LayersDecoder function implementation in the package.

Here are benchmark results comparing with master branch, somewhat average value taken:
```
benchmark                                     old ns/op     new ns/op     delta
BenchmarkDecodingLayerParserIgnorePanic-4     133           130           -2.26%
BenchmarkDecodingLayerParserHandlePanic-4     188           181           -3.72%
```
As you can see, the performance is nearly the same (or even slightly better) for default DecodingLayerMap container.

Some additional benchmarks added to test new containers:

```
BenchmarkDecodingLayerParserIgnorePanic-4               10000000               130 ns/op
BenchmarkDecodingLayerParserHandlePanic-4               10000000               179 ns/op

BenchmarkDecodingLayerParserSparseIgnorePanic-4         20000000               113 ns/op (-13.1%)
BenchmarkDecodingLayerParserSparseHandlePanic-4         10000000               163 ns/op (-8.9%)

BenchmarkDecodingLayerParserArrayIgnorePanic-4          10000000               122 ns/op (-6.2%)
BenchmarkDecodingLayerParserArrayHandlePanic-4          10000000               170 ns/op (-5.0%)

BenchmarkDecodingLayerParserMapIgnorePanic-4            10000000               130 ns/op (~)
BenchmarkDecodingLayerParserMapHandlePanic-4            10000000               178 ns/op (~)
```

As you can see, DecodingLayerSparse container is the fastest with standard LayerType values.

These three benchmarks show how DecodingLayerContainer may be used as a standalone layers decoding tool. As you can see, DecodingLayerParser with identical container imposes some additional overhead though adds some abstraction.
```
BenchmarkDecodingLayerArray-4                           20000000               118 ns/op (-3.3%)
BenchmarkDecodingLayerMap-4                             10000000               124 ns/op (-4.6%)
BenchmarkDecodingLayerSparse-4                          20000000               106 ns/op (-6.2%)
```

Tests were run on a laptop (Core i5-3320M), go 1.12.5. Please cross-check the results since I don't have access to another not heavily loaded physical machine.

Some caveat. LayersDecoder method in all presented DecodingLayerContainer implementations is a carbon copy of the package's LayersDecoder function so it leads to some code redundancy. It may be avoided simply by doing indirect call to LayersDecoder() but in that case the performance gains are not convincing that much. I believe it is so due to inlining done by compiler.

Any thoughts?

PR #658, opened 2019-05-28
Fix double release bug in reassembly

jandos:
While testing `reassembly` package on live traffic, I have encountered a problem with `checkOverlap` function hanging in an infinite loop. After some investigation, I found out that assembler's `pageCache` sometimes returns a `page` that is already in use, i.e. which was already taken from pageCache before, and not yet released. That made me think, that at some point in past that `page` was released twice.  

After more debugging, I found a test case which triggers a double release of a page (8d08f07).
You can verify it, by running the test and adding a check in `sendToConnection`, such as:
```
if half.closed {
    panic("sending on a closed half")
}
```

The bug happens because:
- `sendToConnection` might close a connection and thus free all pending pages, if it's a RST|FIN packet
- in `flushClose` we don't check if halfConnection is already closed before flushing next page (which was already released)

### DIFF
--- a/reassembly/tcpassembly.go
+++ b/reassembly/tcpassembly.go
@@ -1292,6 +1292,7 @@ func (a *Assembler) flushClose(conn *connection, half *halfconnection, t time.Ti
 		a.skipFlush(conn, half)
 		if half.closed {
 			closed = true
+			return flushed, closed
 		}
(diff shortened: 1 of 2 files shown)


gconnell:
Thanks!

PR #663, opened 2019-06-07
returning nil on error fix

gobwas:
Hello there,

I was greping this kind of typo at current working directory and found also this.

If it is not a typo but a feature, maybe comment above would be helpful.

Sergey.


notti:
Looks like a typo to me, but one without consequences (atm); -> both `DecodeFromBytes` return `nil` unconditionally, are inlined by go, and therefore, this `if` is optimized out by the compiler anyway.
But I still think this should be err, in case in the future something changes here...

Can you sign the CLA?

gobwas:
Hi @notti, 
Yes, definetely this is kind of typo. If found the same things in my code in few places. Looks like a good reason to write a linter or expand `go vet` or some other tool.

gobwas:
I signed it!

notti:
Thanks for the fix

Issue #669, opened 2019-06-21
OSPFv2 Type 7 LSAs not decoded


rjrivero:
OSPF layer is not being able to decode Type 7 LSAs. Type 7 LSAs are defined in RFC 3101 (https://tools.ietf.org/html/rfc3101).

To replicate the issue, please unzip the attached file, which contains a single LS Update with a Type 7 LSA:

[nssaLSA.pcap.zip](https://github.com/google/gopacket/files/3316204/nssaLSA.pcap.zip)

Then try to decode it:

```go
package main

import (
	"log"
	"os"

	"github.com/google/gopacket"
	"github.com/google/gopacket/pcap"
)

func main() {
	filename := "nssaLSA.pcap"
	if len(os.Args) > 1 {
		filename = os.Args[1]
	}
	handle, err := pcap.OpenOffline(filename)
	if err != nil {
		panic(err)
	}
	defer handle.Close()
	for packet := range gopacket.NewPacketSource(handle, handle.LinkType()).Packets() {
		log.Println(packet.ErrorLayer())
	}
}
```

PR #672, opened 2019-06-22
Fix issue #669 (OSPF layer fails to decode type 7 LSAs)

rjrivero:
Please check this PR which fixes issue https://github.com/google/gopacket/issues/669#issue-459395018

The Type 7 LSA is specified in RFC 3101 Appendix C. It has the same fields as Type 5 AS External LSA, so this PR reuses the existing ASExternalLSAV2 struct and code.


notti:
Thanks for the contribution

Issue #674, opened 2019-06-26
Wrong network interface info is returned


AjithPanneerselvam:
The routing sub-package is returning a wrong network interface to reach the destination IP passed to `Route()` method of sub-package `routing`. 

We have two networks connected to our machine - public and private network. While passing a machine's IP which is in the private network to the `Route()` method, it returns public network interface.

PR #675, opened 2019-06-26
fix issue - 674

AjithPanneerselvam:
Issue Link - https://github.com/google/gopacket/issues/674

*For the the input destination IP - `192.168.8.180`*

**Before Fix:**
```
gateway: 192.168.0.1;  preferredSrc: 192.168.0.5
```

**After fix:**
```
gateway:<nil>; preferredSrc:192.168.8.243
```

### DIFF
--- a/routing/routing.go
+++ b/routing/routing.go
@@ -136,10 +136,13 @@ func (r *router) route(routes routeSlice, input net.HardwareAddr, src, dst net.I
 		if rt.Src != nil && !rt.Src.Contains(src) {
 			continue
 		}
-		if rt.Dst != nil && !rt.Dst.Contains(dst) {
-			continue
+		if rt.Dst != nil {
+			if !rt.Dst.Contains(dst) {
+				continue
+			}
+			return int(rt.OutputIface), rt.Gateway, rt.PrefSrc, nil
 		}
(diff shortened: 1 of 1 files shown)


AjithPanneerselvam:
I signed it!

AjithPanneerselvam:
@notti  A gentle reminder.

routing/routing.go
```diff
@@ -136,10 +136,13 @@ func (r *router) route(routes routeSlice, input net.HardwareAddr, src, dst net.I
...
 		}
-		if rt.Dst != nil && !rt.Dst.Contains(dst) {
-			continue
+		if rt.Dst != nil {
+			if !rt.Dst.Contains(dst) {
+				continue
+			}
+			return int(rt.OutputIface), rt.Gateway, rt.PrefSrc, nil
 		}
-		return int(rt.OutputIface), rt.Gateway, rt.PrefSrc, nil
```
notti: If you do this, then the default route doesn't match any longer (default route will have both `rt.Src` and `rt.Dst` equal to `nil`) - that just works by coincidence in your case.

AjithPanneerselvam:
@notti Makes sense. Should return the network interface which has the longest common prefix with the destination IP. Will do that.

PR #682, opened 2019-07-10
Implement MPLS Echo layer

Diego-Perez-Botero:
**SUMMARY**
Implementing MPLS Echo protocol, as specified by [RFC8029](https://tools.ietf.org/html/rfc8029). This enables Label Switched Path (LSP) Ping/Traceroute operations to be conducted inside Multiprotocol Label Switching (MPLS) networks for data plane failure detection and localization.

An MPLS echo request/reply is a (possibly labeled) IPv4 or IPv6 UDP packet. Wireshark already displays MPLS Echo as a separate layer in packet captures.

**VALIDATION**
- Passes `./gc`.
- Adding automated tests for all new code, achieving 93% mpls_echo.go code coverage.
- I've gotten real CISCO routers to reply to my MPLS Echo requests crafted using this code.
- Wireshark packet captures confirm that my crafted packets are identical to those generated by CISCO's "ping mpls" command.

### DIFF
--- a/layers/mpls_echo.go
+++ b/layers/mpls_echo.go
@@ -1,0 +1,529 @@
+// Copyright 2018 GoPacket Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"github.com/google/gopacket" // BSD license
+	"net"
+)
+
+// This file implements a layer that enables Label Switched Path (LSP) Ping/Traceroute operations to be conducted
+// inside Multiprotocol Label Switching (MPLS) networks for data plane failure detection and localization, per RFC8029.
+
+// The version number is to be incremented whenever a change is made that affects the ability of an implementation to
+// correctly parse or process an MPLS echo request/reply.
+type MPLSEchoVersion uint16
+
+const (
+	MPLSEchoVersion1 MPLSEchoVersion = iota + 1
+)
+
+// Bitmask type that represents an MPLS echo request/reply's global flags.
+type MPLSEchoGlobalFlags uint16
+
+const (
+	// Flag is 1 if the sender wants the receiver to perform FEC Stack validation;
+	// if flag is 0, the choice is left to the receiver.
+	MPLSEchoFlagValidateFECStack MPLSEchoGlobalFlags = 1 << iota // 1
+
+	// Can only be set in echo request packets.  If set to 1 in an incoming echo request and the TTL of the incoming
+	// MPLS label is greater than 1, the receiving node MUST drop the oacket and MUST NOT send back an echo reply.
+	MPLSEchoFlagRespondOnlyIfTTLExpired // 2
+
+	// If this flag is set in the echo request, the Responder SHOULD return reverse-path FEC information (see RFC6426).
+	MPLSEchoFlagValidateReversePath // 4
+)
+
+// Define bitmask operations.
+func SetMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) MPLSEchoGlobalFlags   { return b | flag }
+func ClearMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) MPLSEchoGlobalFlags { return b &^ flag }
+func HasMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) bool                  { return b&flag != 0 }
+
+// Specifies the MPLS echo message type (request/reply/...).
+type MPLSEchoMessageType uint8
+
+const (
+	MPLSEchoRequest MPLSEchoMessageType = iota + 1
+	MPLSEchoReply
+)
+
+// Specifies the desired receiver-side behavior.
+type MPLSEchoReplyMode uint8
+
+const (
+	// For one-way connectivity tests.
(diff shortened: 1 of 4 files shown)


Diego-Perez-Botero:
Getting some golint errors...will fix them shortly ;)

Diego-Perez-Botero:
Fixed

---
In reply to: [510266494](https://github.com/google/gopacket/pull/682#issuecomment-510266494) [](ancestors = 510266494)

Diego-Perez-Botero:
@notti - This is ready for review :)

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+	// MPLS label is greater than 1, the receiving node MUST drop the oacket and MUST NOT send back an echo reply.
+	MPLSEchoFlagRespondOnlyIfTTLExpired // 2
+
+	// MPLSEchoFlagValidateReversePath acts as an opt-in for reverse-path FECs.
+	// If this flag is set in the echo request, the Responder SHOULD return reverse-path FEC information (see RFC6426).
+	MPLSEchoFlagValidateReversePath // 4
+)
+
+// SetMPLSEchoGlobalFlag returns an MPLSEchoGlobalFlag bitmask with the given flag enabled.
+func SetMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) MPLSEchoGlobalFlags { return b | flag }
```
notti: Is this function necessary? It is not used anywhere and looks trivial.

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+
+// ClearMPLSEchoGlobalFlag returns an MPLSEchoGlobalFlag bitmask with the given flag disabled.
+func ClearMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) MPLSEchoGlobalFlags { return b &^ flag }
```
notti: Is this function necessary? It is not used anywhere and looks trivial.

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+
+// HasMPLSEchoGlobalFlag returns TRUE if the given flag is enabled in the bitmask.
+func HasMPLSEchoGlobalFlag(b, flag MPLSEchoGlobalFlags) bool { return b&flag != 0 }
```
notti: Is this function necessary? It is not used anywhere and looks trivial.

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+	// request was received.
+	TimestampReceivedSeconds         uint32
+	TimestampReceivedSecondsFraction uint32 // divide by 2^32 to get the fractions of a second
+
+	// TLV types less than 32768 (i.e., with the high-order bit equal to 0) are mandatory TLVs that MUST either be
+	// supported by an implementation or result in MPLSEchoReturnCodeTLVNotUnderstood being sent in the echo response.
+	//
+	// Types greater than or equal to 32768 (i.e., with the high-order bit equal to 1) are optional TLVs that SHOULD be
+	// ignored if the implementation does not understand or support them.
+	TLVs []*MPLSEchoTLV
```
notti: Could you make this `TLVs []MPLSEchoTLV`, which would allow preallocating like options in IP4 or TCP?

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+	m.SenderHandle = binary.BigEndian.Uint32(data[8:12])
+	m.SequenceNumber = binary.BigEndian.Uint32(data[12:16])
+	m.TimestampSentSeconds = binary.BigEndian.Uint32(data[16:20])
+	m.TimestampSentSecondsFraction = binary.BigEndian.Uint32(data[20:24])
+	m.TimestampReceivedSeconds = binary.BigEndian.Uint32(data[24:28])
+	m.TimestampReceivedSecondsFraction = binary.BigEndian.Uint32(data[28:32])
+
+	currentTLVStartOffset := 32
+	for currentTLVStartOffset < len(data) {
+		currentTLV := &MPLSEchoTLV{}
```
notti: See comment in the data structure (use preallocation like options in IP4 or TCP) and I think not using a pointer would be better here.
Memory allocations should be avoided.

layers/mpls_echo.go
```diff
@@ -0,0 +1,550 @@
...
+	binary.BigEndian.PutUint32(bytes[8:], m.SenderHandle)
+	binary.BigEndian.PutUint32(bytes[12:], m.SequenceNumber)
+	binary.BigEndian.PutUint32(bytes[16:], m.TimestampSentSeconds)
+	binary.BigEndian.PutUint32(bytes[20:], m.TimestampSentSecondsFraction)
+	binary.BigEndian.PutUint32(bytes[24:], m.TimestampReceivedSeconds)
+	binary.BigEndian.PutUint32(bytes[28:], m.TimestampReceivedSecondsFraction)
+
+	currentTLVStartOffset := 32
+	for _, currentTLV := range m.TLVs {
+		tlvByteRepresentation, err := currentTLV.EncodeAsBytes()
```
notti: This allocates a byte slice for every TLV. Memory allocations should be avoided. Can you change `EncodeAsBytes` so it writes to the target buffer directly?

layers/mpls_echo_test.go
```diff
@@ -0,0 +1,933 @@
+// Copyright 2018 GoPacket Authors. All rights reserved.
```
notti: Can you make the testfile more in line with the existing tests? (without the helper functions)

PR #684, opened 2019-07-13
Added support for writing pcaps with nanosecond timestamp

rkulseth:
This will allow a user to write pcaps with nanosecond precision timestamps without having to use pcapNG format.  The change is 100% backwards compatible.


rkulseth:
I signed it.

pcapgo/write.go
```diff
@@ -50,14 +72,20 @@ const versionMinor = 4
 //  w2.WritePacket(gopacket.CaptureInfo{...}, data2)
 //  f2.Close()
 func NewWriter(w io.Writer) *Writer {
-	return &Writer{w: w}
+	return &Writer{w: w, tsScaler: nanosPerMicro}
 }
 
 // WriteFileHeader writes a file header out to the writer.
-// This must be called exactly once per output.
+// This must be called exactly once per output.  set usenanos to
```
notti: Looks like the `usenanos` remark is from an older version of the patch. Looks like this can be left out

rkulseth: Yep, that's exactly what happened.  Good catch.  Thank You.

notti:
Thanks for the contribution.

PR #685, opened 2019-07-13
Support RouterV2 details for ospf LSA type1

hrntknr:
The ospf LSA type 1 packet was returned with empty Router information.
Implemented to return details of Router information.


hrntknr:
I signed it!

notti:
Thanks for the improvement.

PR #694, opened 2019-08-09
Make layers.Dot11's DecodeFromBytes method properly reset the structure.

### LINKED ISSUE #693: layers.Dot11's DecodeFromBytes does not reset DataLayer
The comment on the DecodingLayer interface declaration of DecodeFromBytes says the following:

```
	// DecodeFromBytes resets the internal state of this layer to the state
	// defined by the passed-in bytes.  Slices in the DecodingLayer may
	// reference the passed-in data, so care should be taken to copy it
	// first should later modification of data be required before the
	// DecodingLayer is discarded.
```

However, it appears that the DecodeFromBytes defined on layers.Dot11 does not completely reset the internal state of the layer; while it is possible for DecodeFromBytes to *set* the DataLayer struct member, it is never cleared. That means that once we have decoded one packet containing a data layer, the NextLayerType() function may be inaccurate for later packets: 

```
func (m *Dot11) NextLayerType() gopacket.LayerType {
	if m.DataLayer != nil {
		if m.Flags.WEP() {
			return LayerTypeDot11WEP
		}
		return m.DataLayer.(gopacket.DecodingLayer).NextLayerType()
	}
	return m.Type.LayerType()
}
```

I noticed this while attempting to extract SSIDs from the Dot11InformationElement; I found that I could decode the first few packets, but after I saw a data packet I could no longer extract any Dot11InformationElement layers. Setting the DataLayer to nil near the start of layers.Dot11.DecodeFromBytes resulted in correct behavior:

```
func (m *Dot11) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
	if len(data) < 10 {
		df.SetTruncated()
		return fmt.Errorf("Dot11 length %v too short, %v required", len(data), 10)
	}
	m.Type = Dot11Type((data[0])&0xFC) >> 2

	m.DataLayer = nil
	m.Proto = uint8(data[0]) & 0x0003
	m.Flags = Dot11Flags(data[1])
	m.DurationID = binary.LittleEndian.Uint16(data[2:4])
	m.Address1 = net.HardwareAddr(data[4:10])
```

If this is a bug, I'm happy to submit a pull request if necessary.

floren:
Sets the DataLayer field to nil at the start of DecodeFromBytes. This appears to fix #693.

### DIFF
--- a/layers/dot11.go
+++ b/layers/dot11.go
@@ -922,6 +922,7 @@ func (m *Dot11) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 	}
 	m.Type = Dot11Type((data[0])&0xFC) >> 2
 
+	m.DataLayer = nil
 	m.Proto = uint8(data[0]) & 0x0003
 	m.Flags = Dot11Flags(data[1])
 	m.DurationID = binary.LittleEndian.Uint16(data[2:4])


notti:
Thanks for the fix

PR #701, opened 2019-08-27
Add length checks in DecodeFromBytes()

alexhartl:
Length check of passed data was missing in DecodeFromBytes() for several protocols.

### DIFF
--- a/layers/arp.go
+++ b/layers/arp.go
@@ -10,6 +10,7 @@ package layers
 import (
 	"encoding/binary"
 	"errors"
+	"fmt"
 
 	"github.com/google/gopacket"
 )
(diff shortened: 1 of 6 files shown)


layers/arp.go
```diff
@@ -39,6 +40,10 @@ func (arp *ARP) LayerType() gopacket.LayerType { return LayerTypeARP }
 
 // DecodeFromBytes decodes the given bytes into this layer.
 func (arp *ARP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	if len(data) < 8 || len(data) < int(8 + 2*data[4] + 2*data[5]) {
```
notti: Can you change this to first check (data) < 8 and then later check with `HwAddressSize` and `ProtAddressSize` after they have been decoded? This way it is more clear what is going on

layers/eap.go
```diff
@@ -47,6 +47,10 @@ func (e *EAP) LayerType() gopacket.LayerType { return LayerTypeEAP }
 
 // DecodeFromBytes decodes the given bytes into this layer.
 func (e *EAP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	if len(data) < 4 || len(data) < int(binary.BigEndian.Uint16(data[2:4])) {
```
notti: Same here with `Length`

PR #702, opened 2019-09-02
Set BaseLayer.Contents correctly in SIP layer.

daroot:
Make layers.SIP.LayerContents() work correctly, returning the full SIP
message.

### DIFF
--- a/layers/sip.go
+++ b/layers/sip.go
@@ -245,14 +245,11 @@ func (s *SIP) NextLayerType() gopacket.LayerType {
 
 // DecodeFromBytes decodes the slice into the SIP struct.
 func (s *SIP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
-
 	// Init some vars for parsing follow-up
 	var countLines int
 	var line []byte
 	var err error
-
-	// Clean leading new line
-	data = bytes.Trim(data, "\n")
+	var offset int
 
 	// Iterate on all lines of the SIP Headers
 	// and stop when we reach the SDP (aka when the new line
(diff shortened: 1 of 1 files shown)


daroot:
@googlebot I fixed it.

layers/sip.go
```diff
@@ -245,6 +245,9 @@ func (s *SIP) NextLayerType() gopacket.LayerType {
 
 // DecodeFromBytes decodes the slice into the SIP struct.
 func (s *SIP) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+	// There are no further layers, so set BaseLayer's Contents should point at
+	// this layer.
+	s.BaseLayer = BaseLayer{Contents: data[:len(data)]}
```
notti: Well I agree, that `Contents` is missing, but "no further leayers" is not completely true (e.g., INVITE or 200 OK with SDP attached):
https://github.com/google/gopacket/blob/90781af512195c8b22c46db7cce068bc30ef8394/layers/sip.go#L280-L283

daroot: Okay I've updated the comments, and as an added bonus corrected the spelling from Paypload -> Payload.

If you'd like me to, I can squash the commits.

notti: A sorry for not being clear enough: I didn't just mean the comments - also the code is not completely correct. In case there is a payload, the `BaseLayer.Contents` shouldn't hold all the data, but just the part that belongs to this layer (so everything except the Payload).

daroot: Aha.   Now I understand.   I had the wrong assumption that even though there was another logical layer (SDP) since there was no actual decoder layer, that the base layer should include everything.

I've reworked the patch to properly separate LayerContents and Payload.

notti:
Thanks for the contribution

PR #703, opened 2019-09-06
Added SegmentRouting decoding support to ipv6 layer

Zygimantass:
Type 4 Routing header now decodes correctly instead of throwing an error saying that it cannot decode this type of header.

### DIFF
--- a/layers/ip6.go
+++ b/layers/ip6.go
@@ -543,9 +543,9 @@ type IPv6Routing struct {
 	// This segment is supposed to be zero according to RFC2460, the second set of
 	// 4 bytes in the extension.
 	Reserved []byte
-	// SourceRoutingIPs is the set of IPv6 addresses requested for source routing,
-	// set only if RoutingType == 0.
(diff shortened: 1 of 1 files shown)


Zygimantass:
@googlebot I fixed it.

layers/ip6.go
```diff
@@ -543,9 +543,9 @@ type IPv6Routing struct {
 	// This segment is supposed to be zero according to RFC2460, the second set of
 	// 4 bytes in the extension.
 	Reserved []byte
-	// SourceRoutingIPs is the set of IPv6 addresses requested for source routing,
-	// set only if RoutingType == 0.
-	SourceRoutingIPs []net.IP
+	// RoutingIPs is the set of IPv6 addresses requested for source/segment routing,
+	// set only if RoutingType == 0 || RoutingType == 4.
+	RoutingIPs []net.IP
```
notti: We can't change exitsing interfaces, as this will break software. So maybe add a ned field, or keep using the existing one.

PR #716, opened 2019-10-08
Extract extended DNS Rcode from OPT record

x-way:
Add decoding of the extended DNS ResponseCode from OPT records as specified in RFC 6891 section 6.1.3.

Additionally also adds the `BADCOOKIE` DNS ResponseCode (from RFC 7873) to the list of known values.

### DIFF
--- a/layers/dns.go
+++ b/layers/dns.go
@@ -126,25 +126,26 @@ type DNSResponseCode uint8
 
 // DNSResponseCode known values.
 const (
-	DNSResponseCodeNoErr    DNSResponseCode = 0  // No error
-	DNSResponseCodeFormErr  DNSResponseCode = 1  // Format Error                       [RFC1035]
-	DNSResponseCodeServFail DNSResponseCode = 2  // Server Failure                     [RFC1035]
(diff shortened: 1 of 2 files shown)


layers/dns.go
```diff
@@ -370,6 +373,12 @@ func (d *DNS) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
 			d.Additionals = d.Additionals[:i] // strip off erroneous value
 			return err
 		}
+		// extract extended RCODE from OPT RRs, RFC 6891 section 6.1.3
+		if d.Additionals[i].Type == DNSTypeOPT {
+			ttldata := make([]byte, 4)
+			binary.BigEndian.PutUint32(ttldata[:], d.Additionals[i].TTL)
+			d.ResponseCode = DNSResponseCode(uint8(d.ResponseCode) | uint8(ttldata[0])<<4)
```
notti: Isn't there some better way to do that than to allocate memory? I think some simple bit arithmetic should work (TTL will be in host byte order at this point - so if I'm not mistaken `d.Additionals[i].TTL>>24` and then `<<4` or shift by 16 and mask out lower 4 bits)

notti:
Thanks

PR #718, opened 2019-10-09
add descriptive error message on activate

FireFart:
This PR adds a more detailed error message if activation fails.

From https://www.tcpdump.org/manpages/pcap_activate.3pcap.txt:
```text
PCAP_ERROR
       Another error occurred.  pcap_geterr() or pcap_perror()  may  be
       called  with  p  as  an  argument  to fetch or display a message
       describing the error.
```

This happened when using bettercap with the https://github.com/aircrack-ng/rtl8812au driver (see here https://github.com/bettercap/bettercap/issues/614).

Without the patch:
```text
[err] error while activating handle: unknown activated error: -1
```

After the patch:
```text
[err] error while activating handle: Error: wlan0: SIOCGIWPRIV: Argument list too long
```

### DIFF
--- a/pcap/pcap.go
+++ b/pcap/pcap.go
@@ -271,6 +271,7 @@ const (
 	aeDenied       = activateError(pcapErrorDenied)
 	aeNotUp        = activateError(pcapErrorNotUp)
 	aeWarning      = activateError(pcapWarning)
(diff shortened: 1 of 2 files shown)


notti:
Thanks for the contribution

PR #720, opened 2019-10-18
afpacket: Fix support for 32-bit x86 arch

adriansr:
afpacket uses syscalls unix.SYS_GETSOCKOPT and unix.SYS_SETSOCKOPT
when it needs to call [gs]etsockopt() passing a raw pointer. This
doesn't work for 32-bit x86 platforms as there's no such syscall,
resulting in:

   setsockopt packet_rx_ring: function not implemented

   unix.Syscall(unix.SYS_GETSOCKOPT[=366],...) = 38 // ENOSYS

The correct way to call [gs]etsockopt() in this platform is to use
the SYS_SOCKETCALL syscall with the right call parameter.

This patch refactors the raw [gs]etsockopt() calls in afpacket to use
unix.GetsockoptString and unix.SetsockoptString so that it relies on
Go runtime to call the appropriate syscall.


adriansr:
@googlebot I signed it!

adriansr:
@googlebot I signed it!

adriansr:
@notti can I get a review? Sorry for the ping :)

PR #726, opened 2019-11-12
Fix Resolution() method return value

mateumann:
Resolution() methods defined for pcap.Handle and pcapgo.Reader returned
an invalid definition of gopacket.TimestampResolution (µs instead of ns
and vice-versa).  This patch fixes that.


mateumann:
@googlebot I signed it!

PR #737, opened 2019-12-13
Adds a fuzz target starting from every layer

catenacyber:
This is to allow fuzzing with platforms such as oss-fuzz cf https://github.com/google/oss-fuzz/pull/3110

I can change the copyright if you need to

There is already a fuzz target for go packet here :
https://github.com/dvyukov/go-fuzz-corpus/blob/master/gopacket/main.go

But it is only for the DNS layer, it does not print the contents, and it recovers from panics when accessing slices out of bounds

### DIFF
--- a/layers/fuzz_layer.go
+++ b/layers/fuzz_layer.go
@@ -1,8 +1,7 @@
-// Copyright 2019 Catena cyber All rights reserved.
+// Copyright 2019 The GoPacket Authors. All rights reserved.
 //
-// Use of this source code is governed by a BSD-style license
-// that can be found in the LICENSE file in the root of the source
-// tree.
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file in the root of the source tree.
 
 package layers
 
@@ -12,19 +11,24 @@ import (
 	"github.com/google/gopacket"
 )
 
-var fuzzOpts = gopacket.DecodeOptions{
-	Lazy:                     false,
-	NoCopy:                   true,
-	SkipDecodeRecovery:       true,
-	DecodeStreamsAsDatagrams: true,
-}
-
+// FuzzLayer is a fuzz target for the layers package of gopacket
+// A fuzz target is a function processing a binary blob (byte slice)
+// The process here is to interpret this data as a packet, and print the layers contents.
+// The decoding options and the starting layer are encoded in the first bytes.
+// The function returns 1 if this is a valid packet (no error layer)
 func FuzzLayer(data []byte) int {
-	if len(data) < 2 {
-		return 1
+	if len(data) < 3 {
+		return 0
 	}
+	// use the first two bytes to choose the top level layer
 	startLayer := binary.BigEndian.Uint16(data[:2])
-	p := gopacket.NewPacket(data[2:], gopacket.LayerType(startLayer), fuzzOpts)
+	var fuzzOpts = gopacket.DecodeOptions{
+		Lazy:                     data[2]&0x1 != 0,
+		NoCopy:                   data[2]&0x2 != 0,
+		SkipDecodeRecovery:       data[2]&0x4 != 0,
+		DecodeStreamsAsDatagrams: data[2]&0x8 != 0,
+	}
+	p := gopacket.NewPacket(data[3:], gopacket.LayerType(startLayer), fuzzOpts)
 	for _, l := range p.Layers() {
 		gopacket.LayerString(l)
 	}
--- a/layers/fuzz_layer.go
+++ b/layers/fuzz_layer.go
@@ -1,0 +1,35 @@
+// Copyright 2019 Catena cyber All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file in the root of the source
+// tree.
+
+package layers
+
+import (
+	"encoding/binary"
+
+	"github.com/google/gopacket"
+)
+
+var fuzzOpts = gopacket.DecodeOptions{
+	Lazy:                     false,
+	NoCopy:                   true,
+	SkipDecodeRecovery:       true,
+	DecodeStreamsAsDatagrams: true,
+}
+
+func FuzzLayer(data []byte) int {
+	if len(data) < 2 {
+		return 1
+	}
+	startLayer := binary.BigEndian.Uint16(data[:2])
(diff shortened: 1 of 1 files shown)


layers/fuzz_layer.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2019 Catena cyber All rights reserved.
```
gconnell: Please change to:

```
// Copyright 2019 The GoPacket Authors. All rights reserved.
//
// Use of this source code is governed by a BSD-style license that can be found
// in the LICENSE file in the root of the source tree.
```

layers/fuzz_layer.go
```diff
@@ -0,0 +1,35 @@
...
+
+package layers
+
+import (
+	"encoding/binary"
+
+	"github.com/google/gopacket"
+)
+
+var fuzzOpts = gopacket.DecodeOptions{
```
gconnell: could this be passed into FuzzLayer?

layers/fuzz_layer.go
```diff
@@ -0,0 +1,35 @@
...
+	NoCopy:                   true,
+	SkipDecodeRecovery:       true,
+	DecodeStreamsAsDatagrams: true,
+}
+
+func FuzzLayer(data []byte) int {
+	if len(data) < 2 {
+		return 1
+	}
+	startLayer := binary.BigEndian.Uint16(data[:2])
```
gconnell: could you add a comment on what you're doing here?  My impression is that you're allowing the fuzzer to chose its own top-level layer by using the first 2 bytes to get it?

layers/fuzz_layer.go
```diff
@@ -0,0 +1,35 @@
...
+	p := gopacket.NewPacket(data[2:], gopacket.LayerType(startLayer), fuzzOpts)
+	for _, l := range p.Layers() {
+		gopacket.LayerString(l)
```
gconnell: This seems like it would slow down the fuzzing a good bit.  Have you tried removing lines 28-30 to see if there's a speedup?

layers/fuzz_layer.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var fuzzOpts = gopacket.DecodeOptions{
+	Lazy:                     false,
+	NoCopy:                   true,
+	SkipDecodeRecovery:       true,
+	DecodeStreamsAsDatagrams: true,
+}
+
+func FuzzLayer(data []byte) int {
```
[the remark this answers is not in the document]
catenacyber: ok

layers/fuzz_layer.go
catenacyber: ok

layers/fuzz_layer.go
catenacyber: Indeed, this is a good idea.
I chose these options as they seem to maximize the number of panics we can find

layers/fuzz_layer.go
catenacyber: Ok.
Your impression is correct.
This is kind of a speed optimization, so as not to parse Ethernet/IP/TCP over again to finally get to the application layer.
Plus there may be protocols that I do not know that are not reachable if we start from Ethernet...

layers/fuzz_layer.go
catenacyber: When I added it, it finds new bugs (one in tcp layer) to be balanced with the performance cost.
Overall, the performance is over a few thousands executions per second, so I didn't try to optimize too much. The optimization of the start layer was significant enough.

catenacyber:
Happy new year and ping @gconnell for these changes

layers/fuzz_layer.go
notti: I'm not completely sure, but I think @gconnell  meant to make `fuzzOpts` an argument of `FuzzLayer`, if possible, so one can change the settings more easily.

catenacyber: This is indeed a good idea but not yet possible.
This is called structure ware fuzzing.
Cf https://github.com/golang/go/issues/19109 https://docs.google.com/document/u/1/d/1zXR-TFL3BfnceEAWytV8bnzB2Tfp6EPFinWVJ5V4QC8/pub https://twitter.com/dvyukov/status/1219175271594676224

catenacyber: Maybe we can do something with protobuf
Do you want me to try @notti ?

Dor1s:
@motti @gconnell ping from Google's OSS-Fuzz. Could you please take another look here? Thanks!

layers/fuzz_layer.go
```diff
@@ -0,0 +1,39 @@
...
+	if len(data) < 3 {
+		return 0
+	}
+	// use the first two bytes to choose the top level layer
+	startLayer := binary.BigEndian.Uint16(data[:2])
+	var fuzzOpts = gopacket.DecodeOptions{
+		Lazy:                     data[2]&0x1 != 0,
+		NoCopy:                   data[2]&0x2 != 0,
+		SkipDecodeRecovery:       data[2]&0x4 != 0,
```
tomfitzhenry, line 28: If SkipDecodeRecovery is false (as it would be ~half of the time), then if a layer panics, [gopacket.NewPacket will recover](https://github.com/google/gopacket/blob/1d829e51f0c85294eeedb06477a6d369fb5be0ea/packet.go#L197-L201), and thus this crash will be hidden from the fuzzer.

Is it worth setting SkipDecodeRecovery to true to avoid this from happening?

layers/fuzz_layer.go
tomfitzhenry: > I'm not completely sure, but I think @gconnell meant to make fuzzOpts an argument of FuzzLayer, if possible, so one can change the settings more easily.

I don't think this is possible at the moment. Per https://github.com/dvyukov/go-fuzz/blob/8cb20381268133974b070fff8f404435b3d189c1/go-fuzz-build/main.go#L378 , the type signature must be `[]byte -> int`.

layers/fuzz_layer.go
catenacyber: Indeed.

But when SkipDecodeRecovery is false, the fuzz target can still time out, or crash while calling `gopacket.LayerString`

So, it seems worth to keep it, even if it will not find most crashes

PR #740, opened 2019-12-27
remove index ref

charles-d-burton:
There is no guarantee that the iface number will match an index position in a slice.  Removing the check as it causes a failure condition in servers where interfaces may be added/removed dynamically.

Fixes issue:
https://github.com/google/gopacket/issues/698


charles-d-burton:
Closing in favor of other PRs.

PR #741, opened 2019-12-29
Add support for read/write Options in EPB

kor44:
- Change NgOption structure and some structure fields to be public
- Add optional parameters NgOption to WritePacket.
- Add option to NgReader for enable/disable reading options
- Add test and benchmark for read/write packets with options
- Change some test function: add t.Run to be more easy find error
testcase


kor44:
Hello.
What can I do to merge my pull request? May be PR has some issues?

Issue #744, opened 2020-01-08
The SIP parser does not fail on partial SIP messages


rkoshy:
If you attempt to parse a truncated or partial SIP message, the parser does not throw any type of error when used as:
```
parser := gopacket.NewDecodingLayerParser( layers.LayerTypeSIP, sip, )
```
I believe there should a call to SetTruncated() in the SIP parser where:
```
		// Read next line
		line, err = buffer.ReadBytes(byte('\n'))
		if err != nil {
			if err == io.EOF {
				break
			} else {
				return err
			}
		}
```
should actually be:
```
		// Read next line
		line, err = buffer.ReadBytes(byte('\n'))
		if err != nil {
			if err == io.EOF {
				if len(bytes.Trim(line, "\r\n")) > 0 {
					df.SetTruncated()
				}
				break
			} else {
				return err
			}
		}
```

PR #747, opened 2020-01-10
Fixes SIP layer to correctly indicate truncated SIP packets

rkoshy:
Currently, if a SIP packet is truncated, the parser does not indicate this.  The fix calls `SetTruncated()` if this condition is detected.   SIP packets can be considered as truncated if each line does not end in a CR-LF


rkoshy:
@googlebot I signed it!

rkoshy:
Closes #744

rkoshy:
Any chance of getting this pulled?  Do I need to do anything else?

gconnell:
There you go, sorry for the delay!

PR #753, opened 2020-02-10
layers: remove unused func errorFunc

tklauser:
It is unused since commit 8c5cfc5ae25c ("Attempt to reduce memory usage
caused by enums.go strings.")


notti:
Thanks for the contribution

PR #755, opened 2020-02-13
deliver data on PSH flag

rjm2718:
All available data in the stream should be delivered the application when the TCP PSH flag is set.

### DIFF
--- a/reassembly/tcpassembly.go
+++ b/reassembly/tcpassembly.go
@@ -703,6 +703,12 @@ func (a *Assembler) AssembleWithContext(netFlow gopacket.Flow, t *layers.TCP, ac
 			}
 			half.nextSeq = seq
 			action.queue = false
+		} else if t.PSH {
+			if *debugLog {
+				log.Printf("%v PSH encountered", key)
+			}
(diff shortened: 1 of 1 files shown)


reassembly/tcpassembly.go
```diff
@@ -703,6 +703,12 @@ func (a *Assembler) AssembleWithContext(netFlow gopacket.Flow, t *layers.TCP, ac
 			}
 			half.nextSeq = seq
 			action.queue = false
+		} else if t.PSH {
+			if *debugLog {
+				log.Printf("%v PSH encountered", key)
+			}
+			half.nextSeq = seq
```
gconnell: I don't htink we want to set nextSeq with this?

rjm2718: I think nextSeq does need to get set.  The code I added is within the "if half.nextSeq == invalidSequence {" block a dozen lines up.  The PSH flag in this scenario is effectively another way to indicate a start if a SYN was missed (and *start was not overridden in Accept()).
I think this is not a common case, but I believe the code is correct.

PR #757, opened 2020-02-17
Adding feature for serialization of IGMPv2 packet.

ekarjun:
This feature is to add serialization methods for igmp v2 packet. Method for calculation of checksum is also added.


ekarjun:
@googlebot I signed it!

PR #760, opened 2020-02-25
Igmp membership report packet serialization

ekarjun:
Adding support for serializing igmp v2 and v3 membership report packets.

### DIFF
--- a/layers/igmp.go
+++ b/layers/igmp.go
@@ -353,3 +353,45 @@ func decodeIGMP(data []byte, p gopacket.PacketBuilder) error {
 
 	return errors.New("Unable to determine IGMP type.")
 }
+
+// SerializeTo writes the serialized form of IGMP V2 packet layer into the
+// SerializationBuffer, implementing gopacket.SerializableLayer.
+func (igmp IGMP) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
+       data, err := b.PrependBytes(8915)
+       if err != nil {
+               return err
+       }
+
+       data[0] = byte(igmp.Type)
+       data[1] = byte(igmp.MaxResponseTime)
+       data[2] = 0
+       data[3] = 0
+       copy(data[4:8], igmp.GroupAddress.To4())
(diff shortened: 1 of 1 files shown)


ekarjun:
This submission is related to serialization feature for igmp v2 and v3 membership report. Please look into the submission and let me know your comments.
@gconnell

ekarjun:
@gconnell Would you please look into this submission and let me know your comments on it. I was not able to add reviewer when I created this pull request.

layers/igmp.go
```diff
@@ -353,3 +353,48 @@ func decodeIGMP(data []byte, p gopacket.PacketBuilder) error {
 
 	return errors.New("Unable to determine IGMP type.")
 }
+
+// SerializeTo writes the serialized form of IGMP V2 and V3 packet layer into the
+// SerializationBuffer, implementing gopacket.SerializableLayer.
+func (igmp IGMP) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOptions) error {
+       data, err := b.PrependBytes(8915)
```
gconnell: this seems really high... are all IGMP packets 8915 bytes?

From https://tools.ietf.org/html/rfc2236 and from https://tools.ietf.org/html/rfc3376 I'd expect these packets to be far smaller.

Bytes should only be prepended/appended to 'b' if they're going to be part of the final packet.

So, given the code below, I'd expect PrependBytes(8) for v2 and PrependBytes(8 + x*#groups) or something like that.

ekarjun: I'll implement a logic to calculate it with respect to the igmp version of the packet received and submit this again.

ekarjun: Hi @gconnell,
I have submitted a new commit for the fix. Please go through the code and let me know if its good. In byte size calculation for igmpv3 packets, I haven't considered Auxiliary data as its not being used according to RFC 3376.

ekarjun:
Hi @gconnell, would you please take look at my latest commit and see if its good for merge?
Regards,
Arjun

Issue #667, opened 2019-06-21
ICMPv6 Echo missing payload


ABalanuta:
Trying to get the payload from a ICMPv6 packet. The full data packet contains the payload but it is not shown in the ICMPv6 Echo Layer.
 
```
--- FULL PACKET DATA (118 bytes) ------------------------------------
00000000  ca fe be ee ee ff b8 27  eb 4e 94 81 86 dd 60 02  |.......'.N....`.|
00000010  91 e4 00 40 3a 40 fe 80  00 00 00 00 00 00 f9 23  |...@:@.........#|
00000020  9d 8b 43 f3 55 3d fe 80  00 00 00 00 00 00 ca fe  |..C.U=..........|
00000030  be ee ee ee ee ff 80 00  2a c3 40 73 00 01 7f 1d  |........*.@s....|
00000040  0d 5d 08 12 00 00 08 09  0a 0b 0c 0d 0e 0f 10 11  |.]..............|
00000050  12 13 14 15 16 17 18 19  1a 1b 1c 1d 1e 1f 20 21  |.............. !|
00000060  22 23 24 25 26 27 28 29  2a 2b 2c 2d 2e 2f 30 31  |"#$%&'()*+,-./01|
00000070  32 33 34 35 36 37                                 |234567|
--- Layer 1 ---
Ethernet        {Contents=[..14..] Payload=[..104..] SrcMAC=b8:27:eb:4e:94:81 DstMAC=ca:fe:be:ee:ee:ff EthernetType=IPv6 Length=0}
00000000  ca fe be ee ee ff b8 27  eb 4e 94 81 86 dd        |.......'.N....|
--- Layer 2 ---
IPv6    {Contents=[..40..] Payload=[..64..] Version=6 TrafficClass=0 FlowLabel=168420 Length=64 NextHeader=ICMPv6 HopLimit=64 SrcIP=fe80::f923:9d8b:43f3:553d DstIP=fe80::cafe:beee:eeee:eeff HopByHop=nil}
00000000  60 02 91 e4 00 40 3a 40  fe 80 00 00 00 00 00 00  |`....@:@........|
00000010  f9 23 9d 8b 43 f3 55 3d  fe 80 00 00 00 00 00 00  |.#..C.U=........|
00000020  ca fe be ee ee ee ee ff                           |........|
--- Layer 3 ---
[...]

PR #762, opened 2020-02-27
Decode ICMPv6 Echo payload

x-way:
Fixes issue #667


x-way:
Any update regarding this fix? @gconnell

PR #764, opened 2020-03-03
add support for VRRPv3 decode and serialize

avivdolev:
Changed vrrp.go and vrrp_test.go to add VRRPv3, decoding and serializing.
Also touched layers/tls_test.go and routing/other.go to make the gc script happy.


avivdolev:
@googlebot I signed it!

PR #767, opened 2020-03-13
change ip6 layer hbh field to pointer type 

procself:
Change the hbh filed of ipv6 layer to *IPv6HopByHop to allow re-usage of the struct when calling DecodeFromBytes.


procself:
@googlebot I signed it!

PR #769, opened 2020-03-26
Add Support for DNS NAPTR Records

tadamra:
NAPTR records RFC2915:

- Added Support for NAPTR encoding of the records

-  add a function to retrieve the length of type <character-string> and a function to retrieve it as string 

- extended the unit tests to support testing NAPTR Records


tadamra:
@notti whom should I ping in order to get my code reviewed. You did review some of my code in the past

lhausermann:
Hi @tadamra 

I see you PR was not merged yet. I had a look and I see you are implementing three function to move from `[]byte` to `string` (eg `encodeCharacterString`). 

I did not proceed to a careful review of all the corner cases but I wanted to know if you had a look to the standard golang `strings` package and if we can switch these to this standard calls. It will be safer. Then I can merge it

Best

PR #772, opened 2020-04-21
reassembly: Fix offset calculation in reassemblyObject.CaptureInfo

kku1993:
CaptureInfo is supposed to return info of packet corresponding to the
given offset. The current implementation does not take the livePacket
or page length into account, so any offset other than the first offset
of a packet will get the CaptureInfo of the next packet.

This PR also removes a redundant `CaptureInfo` field within
`livePacket`.


kku1993:
@googlebot I signed it!

kku1993:
Ping :)

gconnell:
Thanks!

Issue #743, opened 2020-01-07
ZeroCopyReadPacketData of TPacket panic after calling Close in other goroutine


kery:
(*TPacket).ZeroCopyReadPacketData is called in a goroutine and it blocking that goroutine. I want to cancel that goroutine by calling (*TPacket).Close in other gorouting, then panic occurred. To align with golang convention, it should be safe to call (*TPacket).Close method in other goroutine, and the calling of ZeroCopyReadPacketData should return after the calling of Close.

goroutine 11 [running]:
runtime.throw(0x930fed, 0x5)
        /snap/go/4901/src/runtime/panic.go:774 +0x72 fp=0xc000173c48 sp=0xc000173c18 pc=0x42ebd2
runtime.sigpanic()
        /snap/go/4901/src/runtime/signal_unix.go:401 +0x3de fp=0xc000173c78 sp=0xc000173c48 pc=0x4436ae
github.com/google/gopacket/afpacket.(*v2header).getStatus(0x7f9b10000000, 0x1)
        /home/kery/go/src/github.com/google/gopacket/afpacket/header.go:114 +0x5 fp=0xc000173c80 sp=0xc000173c78 pc=0x754f85
github.com/google/gopacket/afpacket.(*TPacket).pollForFirstPacket(0xc00018c900, 0x9ed360, 0x7f9b10000000, 0x475c4c, 0xc0000a8040)
        /home/kery/go/src/github.com/google/gopacket/afpacket/afpacket.go:453 +0x5e fp=0xc000173cd8 sp=0xc000173c80 pc=0x754cce
github.com/google/gopacket/afpacket.(*TPacket).ZeroCopyReadPacketData(0xc00018c900, 0xc000000006, 0x93c52e, 0xc, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/kery/go/src/github.com/google/gopacket/afpacket/afpacket.go:302 +0x136 fp=0xc000173d70 sp=0xc000173cd8 pc=0x754526
sim-data/tunnel.(*GtpuConn).Read(0xc000176b10, 0xc0001ec000, 0x5ec, 0x5ec, 0x0, 0x0, 0x0)

caseybarker:
[...]

PR #777, opened 2020-04-30
Allow asynchronous Close() in afpacket (Issue #743)

caseybarker:
Issue #743 concerns the common Go convention of cleanly interrupting a blocking `Read()` call by calling `Close()` from a different goroutine. There are a couple of underlying issues in `afpacket` that can cause crashes if `TPacket.Close()` is called while a read is happening asynchronously.

First, the existing `Close()` method would close the socket, and then also `Munmap()` the ring buffer. Doing so immediately invalidates the buffer, meaning that any asynchronous operation operating on data in the buffer would panic. This is fundamentally incompatible with allowing read and close ops to happen asynchronously.

Second, the existing `afpacket` read call path did not check for a `poll()` response indicating that socket had been closed (`unix.POLLHUP`). When the poll returned no data, it would loop around and try again, and the header `getStatus()` call would panic trying to read the ring buffer. This is the call trace recorded in #743.

Note, the `TPacket` does have a mutex to protect its state, but this mutex is locked for the duration of any read operation, so the `Close()` method cannot simply block on it when the intent is to _interrupt_ the read call.

The guarantee that `ZeroCopyReadPacketData()` makes to callers is that the read data is valid until the next call to `ZeroCopyReadPacketData()`. The root of this issue is that an async call to `Close()` violates that guarantee (even if ultimately it's being caused by the caller's code operating asynchronously), and there's no clean way for a caller to work around this. So to allow async read/close, we have to preserve the ring buffer past the `Close()` call, potentially for the lifetime of the `TPacket`, or at least until the reading goroutine declares that it's done with the data.

The fix proposed in this PR does the following:
- `Close()` now only closes the socket FD, which interrupts any outstanding read ops but does not invalidate existing data.
- The `fdClosed` flag is added (as opposed to the use of a -1 sentry value) to indicate that the socket is closed, so any future access to the FD will fail as closed, rather than attempting to use an invalid FD. This is just a cleanup.
- The read path, specifically `pollForFirstPacket()`, now checks for additional errors returned from the `poll()` operation, including `POLLHUP`, and passes an error to the caller.
- A new `Finalize()` method is exported to allow package callers to specify when they're finished with the outstanding data. This method is registered as the `*TPacket` finalizer, so existing code will still clean up correctly once `TPacket` is garbage collected. However, callers may immediately invoke `Finalize()` to clean up manually.

This seemed like the most straightforward way to preserve the integrity of the data while allowing an asynchronous `Close()` operation. One downside to this approach is that it may delay the cleanup of the ring buffer in existing code that does not explicitly call `Finalize().` Alternatively, we could attempt to detect the closure and then Finalize the ring buffer somewhere in the read call path, at the expense of a bit more implementation complexity. I'm willing to explore this option if desired.

### DIFF
--- a/afpacket/afpacket.go
+++ b/afpacket/afpacket.go
@@ -103,6 +103,8 @@ type TPacket struct {
 	stats Stats
 	// fd is the C file descriptor.
 	fd int
+	// fdClosed indicates the fd has been closed.
+	fdClosed bool
 	// ring points to the memory space of the ring buffer shared by tpacket and the kernel.
 	ring []byte
 	// rawring is the unsafe pointer that we use to poll for packets
@@ -212,18 +214,29 @@ func (h *TPacket) setUpRing() (err error) {
 	return nil
 }
 
-// Close cleans up the TPacket.  It should not be used after the Close call.
-func (h *TPacket) Close() {
-	if h.fd == -1 {
-		return // already closed.
-	}
+// Finalize closes the socket (if needed) and also unmaps the ring buffer.
+// This invalidates any previous zero-copy read data. May be called manually to
+// force immediate clean-up, but otherwise will be called automatically when
+// TPacket is garbage collected.
+func (h *TPacket) Finalize() {
+	runtime.SetFinalizer(h, nil)
+	h.Close()
 	if h.ring != nil {
 		unix.Munmap(h.ring)
 	}
 	h.ring = nil
-	unix.Close(h.fd)
-	h.fd = -1
-	runtime.SetFinalizer(h, nil)
+}
+
+// Close closes the socket and causes any pending reads to return an error.
+// The TPacket should not be used after the Close call.
+// Close does not invalidate previously read data.
+func (h *TPacket) Close() {
+	if !h.fdClosed {
+		h.fdClosed = true
+		unix.Close(h.fd)
+	}
+	// There may still be outstanding references to the ring buffer,
+	// so h.ring is not unmapped until Finalize().
 }
 
 // NewTPacket returns a new TPacket object for reading packets off the wire.
@@ -254,7 +267,7 @@ func NewTPacket(opts ...interface{}) (h *TPacket, err error) {
 	if err = h.InitSocketStats(); err != nil {
 		goto errlbl
 	}
-	runtime.SetFinalizer(h, (*TPacket).Close)
+	runtime.SetFinalizer(h, (*TPacket).Finalize)
 	return h, nil
 errlbl:
 	h.Close()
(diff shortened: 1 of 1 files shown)


afpacket/afpacket.go
```diff
@@ -212,18 +214,29 @@ func (h *TPacket) setUpRing() (err error) {
 	return nil
 }
 
-// Close cleans up the TPacket.  It should not be used after the Close call.
-func (h *TPacket) Close() {
-	if h.fd == -1 {
-		return // already closed.
-	}
+// Finalize closes the socket (if needed) and also unmaps the ring buffer.
```
gconnell, line 217: keep this unexported and just export Close?

caseybarker: I put this through extensive testing and found that the background GC might not reclaim fast enough if I open/read/close in a tight loop, and eventually the OS would kill the process (probably OOM). Making an explicit call to `Finalize()` fixes it.

I'm also thinking about what happens if calling code keeps a reference to previously read data even after the `TPacket` goes out of scope.

Both of these concerns seem unlikely, but I felt that exporting the method (and its godoc) made it clearer that these problems might exist, and gave the implementer a better understanding and control over the validity and life cycle of the data buffer.

Of course, it's still a strange, hacky thing to expose. So if it's better to hide the method, I'll see if we can have the read path at least attempt to handle the cleanup itself. I didn't investigate, but it might not be that difficult...

caseybarker: Actually, it turned out pretty easy to just call `finalize()` if the read fails and the FD is closed. That _mostly_ works, so I'll push a commit when my tests finish.

However, I'll note that my test harness is around production code that uses a `context.Context` to trigger both the `Close()` and also to exit the read loop. (This seems pretty representative of a real use case.) And sometimes, the read loop exits early because it sees the `context` is done, and not because of an actual read failure. In this case, we'd need to make _one final_ call to the read method to make sure the buffer gets finalized. If I don't do that, my stress test loop will still eventually get killed because the GC isn't finalizing and un-mapping buffers fast enough.

negbie:
Run into same issue while working with afpacket fanout. @caseybarker thanks for the patch!

afpacket/afpacket.go
```diff
@@ -212,18 +214,29 @@ func (h *TPacket) setUpRing() (err error) {
...
+// finalize closes the socket (if needed) and also unmaps the ring buffer.
+// This invalidates any previous zero-copy read data. Called if an in-progress
+// read detects the closure, but otherwise will be called automatically when
+// TPacket is garbage collected.
+func (h *TPacket) finalize() {
+	runtime.SetFinalizer(h, nil)
+	h.Close()
 	if h.ring != nil {
 		unix.Munmap(h.ring)
+		h.ring = nil
```
IDisposable, line 226: Ideally, wouldn't we want to _also_ finally do the `h.fd = -1` here as well?

PR #781, opened 2020-05-02
Add missing EAPOLTypes to EAPOLTypeMetadata

pyther:
EAPOLTypes 1, 2, and 4 return UnknownEAPOLType instead of EAPOLStart, EAPOLLogOff, and EAPOLASFAlert.

I'm not sure if this is the correct way to fix this. The comment under `initActualTypeData()` in enums.go implies that these should get out generated by gen2.go. I spent some time seeing if I could follow the logic, but being new to Go, I was unable to find any issues with the generation.

Before Patch
```
88:71:b1:22:a1:e1: UnknownEAPOLType v2, len 0
88:71:b1:22:a1:e1: UnknownEAPOLType v2, len 0
88:71:b1:22:a1:e1: UnknownEAPOLType v2, len 0
00:90:d0:63:ff:01: EAP v1, len 4
00:90:d0:63:ff:01: EAP v1, len 15
```

After Patch
```
88:71:b1:22:a1:e1: EAPOLLogOff v2, len 0
88:71:b1:22:a1:e1: EAPOLStart v2, len 0
88:71:b1:22:a1:e1: EAPOLStart v2, len 0
00:90:d0:63:ff:01: EAP v1, len 4
00:90:d0:63:ff:01: EAP v1, len 15
```

Code: https://gist.github.com/pyther/d8ad389358cffa3ef96b81105fdc6763


pyther:
@googlebot I signed it!

PR #148, opened 2015-10-31
Fix 802.11 management probe request/response

jpillora:
The next protocol `LayerTypeDot11InformationElement` is never decoded because bytes placed in contents instead of payload

PR #790, opened 2020-05-26
dot11.go: set Dot11MgmtProbeReq payload properly

### LINKED ISSUE #749: How to decode InformationElement in Dot11MgmtProbeReq
For 802.11 Beacon frames, I can use ```layer.(*layers.Dot11InformationElement)``` to get the IE, like this:

```
for _, layer := range packet.Layers() {
	if layer.LayerType() == layers.LayerTypeDot11InformationElement {
		dot11info, ok := layer.(*layers.Dot11InformationElement)
        }
}
```

but for ProbeReq frames, ```layer.(*layers.Dot11InformationElement)```doesn't exist, I can only use ```
packet.Layer(layers.LayerTypeDot11MgmtProbeReq).(*layers.Dot11MgmtProbeReq)```
to get the Contents and Payload. It seems that the Information elements are in the Contents. 

I'm new to golang, could you please give me an example of how to decode information elements in DOt11MgmtProbeReq, Thanks

reidrankin:
This fixes #749. It's a subset of the requested changes from #148 and #542, but those never got merged. Hopefully something a bit leaner and meaner will fair better.

### DIFF
--- a/layers/dot11.go
+++ b/layers/dot11.go
@@ -1830,6 +1830,10 @@ func (m *Dot11MgmtProbeReq) CanDecode() gopacket.LayerClass { return LayerTypeDo
 func (m *Dot11MgmtProbeReq) NextLayerType() gopacket.LayerType {
 	return LayerTypeDot11InformationElement
 }
+func (m *Dot11MgmtProbeReq) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error {
+    m.Payload = data
+    return m.Dot11Mgmt.DecodeFromBytes(data, df)
+}
 
 type Dot11MgmtProbeResp struct {
 	Dot11Mgmt


giacomoferretti:
I'm currently working with `Dot11` packets and noticed that `gopacket` didn't decode `Dot11InformationElement` layers from `Dot11MgmtProbeReq`, then I found this PR and tried it. 

This code unfortunately don't pass the tests as it fails `TestPacketPrism` at the last `Dot11InformationElement` layer as it's missing 4 bytes.
I tried to remove this `len(data) - 4`  from here https://github.com/google/gopacket/blob/3aa782ce48d4a525acaebab344cedabfb561f870/layers/dot11.go#L1064
but then the other tests obviously fail.

So my questions are:
* Is it correct to remove the last 4 bytes from `Dot11` packets?
* Is the [Prism packet](https://github.com/google/gopacket/blob/3aa782ce48d4a525acaebab344cedabfb561f870/layers/prism_test.go#L40-L58) correct?

PR #796, opened 2020-07-08
Fix bidirectional/main.go

vinhjaxt:
Should be s.bidi.lastPacketSeen.Before(r.Seen) instead of s.bidi.lastPacketSeen.After(r.Seen)


vinhjaxt:
@googlebot I signed it.

gconnell:
Thanks!

PR #802, opened 2020-07-22
Discard master and slave vocabulary

bopopescu:
For diversity reasons, it would be nice to try to avoid 'master' and 'slave' terminology in this repository which can be associated to slavery. The master-slave terminology could be problematic for people in several countries which has the history of slavery like Romania, USA and many others. Thank you for considering the proposal. Let me know if any changes in the PR are needed, I would be happy to implement them.


gconnell:
I very much appreciate the work, the reasoning, and the idea, but unfortunately I can't condone this within these particular files.  The two files targeted here are actually just local caches/copies of publicly available datasets, specifically:

layers/iana_ports.go = http://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xml
macs/valid_mac_prefixes.go:  http://standards-oui.ieee.org/oui/oui.csv

If you can get the IANA or IEEE to change the names, I'm happy to have them changed here.  But it's important that we try to maintain the mapping to these widely-known names to the best of our abilities.

Issue #628, opened 2019-03-29
LLDP Layers hold redundant data


Galadros:
As best I understand it, layers are supposed to be non-overlapping.  By this I mean that fields of a given layer struct (in this case `LinkLayerDiscovery`) shouldn't contain any data that is also contained in other layer structs.  As-is, the information in the `LinkLayerDiscoveryInfo` layer is entirely redundant with the `Values` field of `LinkLayerDiscovery`:

```
type LinkLayerDiscovery struct {
	BaseLayer
	ChassisID LLDPChassisID
	PortID    LLDPPortID
	TTL       uint16
	Values    []LinkLayerDiscoveryValue
}
```

All information that the `LinkLayerDiscoveryInfo` struct contains is parsed purely from the Values list in `LinkLayerDiscovery` in lines [846](https://github.com/google/gopacket/blob/master/layers/lldp.go#L846) to [883](https://github.com/google/gopacket/blob/master/layers/lldp.go#L883) of [lldp.go](https://github.com/google/gopacket/blob/master/layers/lldp.go).  Even if I'm mistaken and layers _can_ overlap in such a fashion, the way that these two layers are currently written makes it impossible to implement the `DecodingLayer` interface.  Since information from the earlier layer is required to decode the `LinkLayerDiscoveryInfo` layer, I can't write a `DecodeFromBytes` method.  (As-is, the `DecodeLinkLayerDiscovery()` function decodes into two different layers).

There are two different fixes that are possible.  One would be to entirely merge the `LinkLayerDiscoveryInfo` layer into the `LinkLayerDiscovery` layer, since they're already redundant.  The other would be to remove the `Values` field from the `LinkLayerDiscovery` struct (said field is never used except internally within `DecodeLinkLayerDiscovery()`)and separate out the latter half of parsing done in `DecodeLinkLayerDiscovery()`into its own method to decode `LinkLayerDiscoveryInfo`.  I plan to submit a pull request changing this as well as implementing the DecodingLayer interface for both of these layers, but I wanted to first check that I haven't misunderstood the requirements of what a "layer" is.

**TD;DR:** The `LinkLayerDiscovery` and `LinkLayerDiscoveryInfo` layers overlap and can't be turned into DecodingLayers; I plan to fix this as well as implementing said interface.

notti:
Well layer documentation says "Layer represents a single decoded packet layer (using either the OSI or TCP/IP definition of a layer)." so actually LLDP should have been just one layer.
[...]

PR #809, opened 2020-08-04
Add serialization of non-mandatory LLDP TLVs

hslatman:
My use case was to programmatically create an LLDP packet and writing it to a PCAP file. Most of that works, except for TLVs other than the mandatory ones, of which there are many. A small example:

```
	lldpLayer := &layers.LinkLayerDiscovery{
		ChassisID: layers.LLDPChassisID{
			Subtype: layers.LLDPChassisIDSubTypeMACAddr,
			ID:      []byte{0x01, 0x02, 0x03, 0x04, 0x05, 0x06},
		},
		PortID: layers.LLDPPortID{
			Subtype: layers.LLDPPortIDSubtypeIfaceName,
			ID:      []byte{0x31, 0x2f, 0x31, 0x31, 0x2f, 0x31, 0x31, 0x2f, 0x31, 0x31, 0x2f, 0x31},
		},
		TTL: 300,
	}

	// NOTE: example values taken from the Wireshark lldp.detailed.pcap example
	value := []byte{0x00, 0x12, 0x0f, 0x02}
	value = append(value, []byte{0x07, 0x01, 0x00}...)
	lldpValue := layers.LinkLayerDiscoveryValue{
		Type: layers.LLDPTLVOrgSpecific,
		Length: uint16(len(value)),
		Value:  value,
	}

	lldpLayer.Values = append(lldpLayer.Values, lldpValue)
```

The code is based on what I found [here](https://github.com/google/gopacket/commit/ffe439921b3bd5510313e3dd812a6910aa52b7de#diff-c3dcb92e52d5c5b310a5b7788647d6ccL815). I tested the code with two different vendor specific TLVs and it seems to work OK.

I'm working at Layer 2, `LinkLayerDiscovery`. I'm aware there's also a Layer 3 `LinkLayerDiscoveryInfo`, which contains mostly the same data as the `Values` in the `LinkLayerDiscovery` layer. There's some discussion about how the information is shown redundantly [here](https://github.com/google/gopacket/issues/628), but I thought Layer 2 seemed OK to implement this at, but I'm curious to know what others think about that.

### DIFF
--- a/layers/lldp.go
+++ b/layers/lldp.go
@@ -779,6 +779,16 @@ func (c *LinkLayerDiscovery) SerializeTo(b gopacket.SerializeBuffer, opts gopack
 	binary.BigEndian.PutUint16(vb[chassIDLen+portIDLen:], ttlIDLen)
 	binary.BigEndian.PutUint16(vb[chassIDLen+portIDLen+2:], c.TTL)
 
+	for _, v := range c.Values {
+		vb, err := b.AppendBytes(int(v.Length) + 2) // +2 for TLV type and length; 1 byte for subtype is included in v.Value
+		if err != nil {
+			return err
+		}
+		idLen := ((uint16(v.Type) << 9) | v.Length)
+		binary.BigEndian.PutUint16(vb[0:2], idLen)
+		copy(vb[2:], v.Value)
+	}
+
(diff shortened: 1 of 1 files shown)


hslatman:
Thank you, @gconnell

PR #813, opened 2020-08-13
Fix for unhandled error in router.RouteWithSrc

PraveenBSD:
## Fix for unhandled error in router.RouteWithSrc

**Issue:**
panic: runtime error: index out of range at line [109](https://github.com/RealImage/gopacket/blob/3ef9fc8f0c56cf922a9b5ce1e5209235809224cd/routing/routing.go#L109) due to unhandled errors


PraveenBSD:
@googlebot I signed it!

PR #816, opened 2020-08-19
Fix for panic with trailing DNS bytes in data slice

bmeadors:
If the "data" slice passed to decodeRData contains extra (garbage) trailing bytes, the loop mechanism in the parser will continue trying to parse this data and end up walking off the end of the buffer and crashing.  We had some slightly malformed DNS packets come over the wire that were causing this crash to happen.  A simple patch was made to truncate the "data" slice at the determined end so that only the valid bytes would be passed to decodeRData.  This prevents the parsing logic from looping off the end of the slice.


gconnell:
Thanks!

PR #821, opened 2020-09-12
Initial implemented RADIUS Protocol

takumin:
- Support Attributes Type String():
  - [RFC 2865 \- Remote Authentication Dial In User Service \(RADIUS\)](https://tools.ietf.org/html/rfc2865)
  - [RFC 2866 \- RADIUS Accounting](https://tools.ietf.org/html/rfc2866)
  - [RFC 2867 \- RADIUS Accounting Modifications for Tunnel Protocol Support](https://tools.ietf.org/html/rfc2867)
  - [RFC 2868 \- RADIUS Attributes for Tunnel Protocol Support](https://tools.ietf.org/html/rfc2868)
  - [RFC 2869 \- RADIUS Extensions](https://tools.ietf.org/html/rfc2869)
- Partial support EAP-Message attributes


takumin:
@googlebot I signed it!

takumin:
again...

@googlebot I signed it!

layers/radius.go
```diff
@@ -0,0 +1,517 @@
...
+	if len(data) < radiusMinimumRecordSizeInBytes {
+		df.SetTruncated()
+		return fmt.Errorf("RADIUS length %d too short", len(data))
+	}
+
+	radius.BaseLayer = BaseLayer{Contents: data}
+
+	radius.Code = RADIUSCode(data[0])
+	radius.Identifier = RADIUSIdentifier(data[1])
+	radius.Length = RADIUSLength(binary.BigEndian.Uint16(data[2:4]))
```
tomfitzhenry, line 389: What should happen if radius.Length < radiusMinimumRecordSizeInBytes ?

layers/radius.go
```diff
@@ -0,0 +1,517 @@
...
+	copy(radius.Authenticator[:], data[4:20])
+
+	if len(data) == radiusMinimumRecordSizeInBytes {
+		return nil
+	}
+
+	reader := bytes.NewReader(data[20:])
```
tomfitzhenry, line 396: "Octets outside the range of the Length field MUST be treated as padding and ignored on reception." https://tools.ietf.org/html/rfc2865#page-14

By using the entirety of `data`, does this code use octets outside of the range of the Length field?

layers/radius.go
```diff
@@ -0,0 +1,517 @@
...
+
+	radius.Code = RADIUSCode(data[0])
+	radius.Identifier = RADIUSIdentifier(data[1])
+	radius.Length = RADIUSLength(binary.BigEndian.Uint16(data[2:4]))
+	copy(radius.Authenticator[:], data[4:20])
+
+	if len(data) == radiusMinimumRecordSizeInBytes {
+		return nil
+	}
+
```
tomfitzhenry, line 395: What should happen if radius.Length > len(data)?

"If the packet is shorter than the Length field indicates, it MUST be silently discarded." https://tools.ietf.org/html/rfc2865#page-14

layers/radius.go
```diff
@@ -0,0 +1,517 @@
...
+	for {
+		header = header[:2]
+
+		if _, err := reader.Read(header); err == io.EOF {
+			break
+		} else if err != nil {
+			return fmt.Errorf("RADIUS unknown error: %s", err)
+		}
+
+		value = value[:header[1]-2]
```
tomfitzhenry, line 408: Consider using variables for `header[1]` to make it clearer that this is a length field.

layers/radius.go
tomfitzhenry: I think this should be checked and truncated, e.g.:
```
if len(data) > radius.Length {
  df.setTruncated()
  data = data[:radius.Length]
}
```

layers/radius.go
tomfitzhenry: I think your code handles this fine: line 402's `err == io.EOF` will cause attribute parsing to early-exit.

layers/radius.go
takumin: `RFC 2865 3. Packet Format` does not have a definition of less than 20 bytes, so change it to execute` df.SetTruncated () `.

https://tools.ietf.org/html/rfc2865#page-15

takumin: Implement the following code on line 370

```go
	if int(radius.Length) < radiusMinimumRecordSizeInBytes {
		df.SetTruncated()
		data = data[:radiusMinimumRecordSizeInBytes]
	}
```

layers/radius.go
takumin: Thank you for your advice! I will implement it as you commented.

layers/radius.go
takumin: I think the code below is appropriate, but how about it?

```go
	if int(radius.Length) < radiusMinimumRecordSizeInBytes {
		df.SetTruncated()
		return fmt.Errorf("RADIUS length %d too short", radius.Length)
	}
```

layers/radius.go
takumin: When I added a test case based on your advice, there was a problem that was difficult to control, so I decided to rewrite the code drastically.

takumin:
@tomfitzhenry 
The length check has been enhanced and test cases have been added.
Please review again!
Thank you very much

tomfitzhenry (approved):
LGTM!

gconnell:
THanks a ton for the submission!

PR #824, opened 2020-09-28
reassembly: Update skip count in Assember.cleanSG

kku1993:
As `cleanSG` iterates through `byteContainers` to convert them to pages, it needs to update the skip count (number of bytes to skip from the start of each container). Failure to do so may result in a panic, as demonstrated with the unit test.


gconnell:
Thanks!

PR #825, opened 2020-09-28
reassembly: Add AssemblerContext method to ScatterGather interface.

kku1993:
This method retrieves the AssemblerContext associated with a particular
offset when processsing reassembled streams, allowing callers to supply
additional metadata through custom implementations of the
AssemblerContext interface.

### DIFF
--- a/reassembly/tcpassembly.go
+++ b/reassembly/tcpassembly.go
@@ -98,6 +98,9 @@ type ScatterGather interface {
 	KeepFrom(offset int)
 	// Return CaptureInfo of packet corresponding to given offset
 	CaptureInfo(offset int) gopacket.CaptureInfo
+	// Returns AssemblerContext of packet corresponding to given offset. Returns
+	// nil if the offset is invalid.
+	AssemblerContext(offset int) AssemblerContext
 	// Return some info about the reassembled chunks
 	Info() (direction TCPFlowDirection, start bool, end bool, skip int)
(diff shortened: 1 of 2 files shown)


gconnell:
Could you provide a little more detail into the problem this is trying to solve?  What can I do with this integrated that I was unable to do beforehand?

shirsa:
@gconnell this PR enables to synchronize bidirectional TCP streams using TCP context, such as seq, ack numbers.
Without the seq and ack for each offset in the assembled chunk, it is very hard to synchronize between the s2c and c2s streams.

kku1993:
(sorry this thread must have fallen through the cracks of my email)

Exactly what @shirsa said. I've been using `AssembleWithContext` to record the TCP seq and ack numbers in a custom implementation of `AssemblerContext` [1], and having the changes in this PR allowed me to access that information during reassembly to match up bidirectional TCP streams (e.g. for HTTP request and response [2]).

[1] https://github.com/akitasoftware/akita-cli/blob/ad383837383eb5a4089b2a8a46e449c8d9b4c254/pcap/net_parse.go#L170
[2] https://github.com/akitasoftware/akita-libs/blob/8388976a7962059b6e7c2f1fe60d0700f62d4e4a/akinet/http/parser.go#L121

shirsa:
@sentryo @darxriggs @kkirsche

PR #828, opened 2020-09-30
Fix LLDP Mgmt OID decoding

### LINKED ISSUE #797: lldp OID string value is different from wireshark OID
hi,
I got the lldp info by gopacket, but I got the LLDPTLVMgmtAddress OID value is different from wireshark. I checked the different value and found that the value in gopacket has moved 2 more bytes to the right：

github.com/google/gopacket/layers/lldp.go

		case LLDPTLVMgmtAddress:
                         ...
			if err := checkLLDPTLVLen(v, int(mlen+6+olen)); err != nil {
				return err
			}
			info.MgmtAddress.OID = string(v.Value[mlen+9 : mlen+9+olen])

Should there the 9 be 7?

x-way:
Uses the correct offset and also fixes the length check to avoid out-of-bounds reads. Fixes #797


gconnell:
Any chance that a test could be added to avoid this regressing in the future?

gconnell:
Awesome, thanks a ton for this!

PR #834, opened 2020-10-13
reassembly: fix check overlap for case 6

jandos:
Like in case 3, `start` or `end` of a tcp segment may match with an existing (`cur`) segment's boundaries , but still be contained in that `cur` segment. Therefore, it should match case 6.


gconnell:
Thanks!

PR #835, opened 2020-10-18
Sctp.chained data chunks

nyrzhun:
This PR fixes sctp data chunk parsing and serialization mentioned in #347, when sctp has bundled data chunks.


nyrzhun:
@googlebot I signed it!

mattLLVW:
@nyrzhun is there a way to access sctpdata layers without looping on **all** layers in packet?

PR #836, opened 2020-10-28
added CanDecode method to ModbusTCP 

lribeiro:
so it can be used in the gopacket.DecodingLayerContainer as it's required by the DecodingLayer interface


lribeiro:
@googlebot I signed it!

PR #838, opened 2020-11-18
fix dns parse error: rr.data could be nil

cshsonice:
In my case, I found that payload
```
name: a.com
type: CNAME
class: 254(QCLASS None)   # [RFC 2136]
ttl: 0
data-length: 0
data: nil
```
it will report a error `errDNSNameOffsetTooHigh` in v1.1.19, and it will regard the next RR's name as this RR's CNAME value in v1.1.18.

### DIFF
--- a/layers/dns.go
+++ b/layers/dns.go
@@ -720,8 +720,10 @@ func (rr *DNSResourceRecord) decode(data []byte, offset int, df gopacket.DecodeF
 	}
 	rr.Data = data[endq+10 : end]
 
-	if err = rr.decodeRData(data[:end], endq+10, buffer); err != nil {
-		return 0, err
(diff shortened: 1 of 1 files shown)


cshsonice:
@googlebot I signed it!

cshsonice:
@googlebot I fixed it.

cshsonice:
@googlebot  I fixed it.

gconnell:
Thanks!  And thanks for helping out googlebot :)

gconnell:
Would it be possible to add your original failure as a test case for DNS?

cshsonice:
> Would it be possible to add your original failure as a test case for DNS?

yes, I have added it.  #839

PR #839, opened 2020-11-19
Add test case 

cshsonice:
add the test case which is mentioned in [PR/838](https://github.com/google/gopacket/pull/838)


gconnell:
Thanks so much, really appreciate the extra effort here :)

PR #841, opened 2020-12-01
added DNSTypeNAPTR and DNSType65

mrizvic:
I observed many Unknown DNSTypes using gopacket.NewDecodingLayerParser(). As I drilled down I found those Unknown's belong to RR's which are detected by ISC BIND as NAPTR and TYPE65.


mrizvic:
@googlebot I signed it!

PR #847, opened 2020-12-16
Return nil error on empty slice in (*DecodingLayerParser).DecodeLayers

### LINKED ISSUE #846: Changed behavior of (*DecodingLayerParser).DecodeLayers on empty data in v1.1.18
Before v1.1.18, the following code would not print any message, i.e. `dlp.DecodeLayers` returns a `nil` error:

```go
import (
    "fmt"

    "github.com/google/gopacket"
    "github.com/google/gopacket/layers"
)

func main() {
        dlp := gopacket.NewDecodingLayerParser(layer.LayerTypeEthernet, &layers.Ethernet{}, &layers.IPv4{}, &layers.TCP{}, &layersUDP{})
        decoded := []gopacket.LayerType{}
        empty := []byte{}
        err := dlp.DecodeLayers(empty, &decoded)
        if err != nil {
                fmt.Printf("Error from dlp parser: %v\n", err)
        }
}
```

In v1.1.18 this behavior changed and an error is now returned from the first layer, i.e. `layers.Ethernet` in this case:

```
Error from dlp parser: Ethernet packet too small
```

This seems to be due to 04f6565bda0a7c4e84b3b3d6024cab497663674f (PR #660), namely the following change:

```diff
@@ -153,23 +304,15 @@ func (l *DecodingLayerParser) DecodeLayers(data []byte, decoded *[]LayerType) (e
        if !l.IgnorePanic {
                defer panicToError(&err)
        }
-       typ := l.first
-       *decoded = (*decoded)[:0] // Truncated decoded layers.
-       for len(data) > 0 {
-               decoder, ok := l.decoders[typ]
-               if !ok {
-                       if l.IgnoreUnsupported {
-                               return nil
-                       }
-                       return UnsupportedLayerType(typ)
-               } else if err = decoder.DecodeFromBytes(data, l.df); err != nil {
-                       return err
+       typ, err := l.decodeFunc(data, decoded)
+       if typ != LayerTypeZero {
+               // no decoder
+               if l.IgnoreUnsupported {
+                       return nil
                }
-               *decoded = append(*decoded, typ)
-               typ = decoder.NextLayerType()
-               data = decoder.LayerPayload()
+               return UnsupportedLayerType(typ)
        }
-       return nil
+       return err
 }
```

We have code that depends on this behavior and strictly speaking this is an API breakage, so my suggestion would be to keep returning `nil` for empty `data`. The other option of course is to check `len(data) != 0` before calling `(*DecodingLayerParser).DecodeLayers` in our but this would be rather cumbersome.

/cc @notti @yerden

tklauser:
Before v1.1.18 passing an empty data slice to
`(*DecodingLayerParser).DecodeLayers` would return a `nil` error. Commit
04f6565 (PR #660) changed this behavior, passing the empty data to the
to return an error from the first layer that considers an empty slice to
be invalid.

Strictly speaking this is an API breakage, so restore the previous
behavior by returning nil straight away on empty data.

Fixes #846


parser.go
```diff
@@ -304,6 +304,12 @@ func (l *DecodingLayerParser) DecodeLayers(data []byte, decoded *[]LayerType) (e
 	if !l.IgnorePanic {
 		defer panicToError(&err)
 	}
+	if len(data) == 0 {
+		// For compatibility with pre 1.1.18 behavior where an empty
+		// slice didn't cause an error. See
+		// https://github.com/google/gopacket/issues/846
```
yerden, line 310: I think we have to do `*decoded = (*decoded)[:0]`.

yerden:
If `DecodeLayers` returning `nil` on an empty packet is ok then I think it is more consistent to also update `decoded` slice to contain no decoded LayerTypes as well.

Regarding whether `DecodeLayers` should return `nil` in this case, I don't have a strong opinion. You'd have to check afterwards if `len(decoded) != 0` anyway, right?

parser.go
tklauser: Agree, thanks for spotting.

tklauser:
> If `DecodeLayers` returning `nil` on an empty packet is ok then I think it is more consistent to also update `decoded` slice to contain no decoded LayerTypes as well.

Agree, truncating `decoded` on an empty packet also matches v1.1.17 behavior. Updated as suggested.

> Regarding whether `DecodeLayers` should return `nil` in this case, I don't have a strong opinion. You'd have to check afterwards if `len(decoded) != 0` anyway, right?

Yes, in our code we're already checking that.

PR #850, opened 2021-01-22
Update function to get all hosts on the subnet

jonasbostoen:
I thought this way of finding all hosts on the network was a little more intuitive.

### DIFF
--- a/examples/arpscan/arpscan.go
+++ b/examples/arpscan/arpscan.go
@@ -178,9 +178,8 @@ func ips(n *net.IPNet) (out []net.IP) {
 	mask := binary.BigEndian.Uint32([]byte(n.Mask))
 	net := num & mask
 	broadcast := net | ^mask
-	for net++; net < broadcast-1; {
+	for net++; net < broadcast-1; net++ {
 		var buf [4]byte
-		net++
 		binary.BigEndian.PutUint32(buf[:], net)
 		out = append(out, net.IP(buf[:]))
 	}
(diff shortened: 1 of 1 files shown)


jonasbostoen:
@googlebot I signed it!

examples/arpscan/arpscan.go
```diff
@@ -176,13 +176,13 @@ func writeARP(handle *pcap.Handle, iface *net.Interface, addr *net.IPNet) error
...
 	num := binary.BigEndian.Uint32([]byte(n.IP))
 	mask := binary.BigEndian.Uint32([]byte(n.Mask))
-	num &= mask
-	for mask < 0xffffffff {
+	net := num & mask
+	broadcast := net | ^mask
+	for net++; net < broadcast-1; {
 		var buf [4]byte
-		binary.BigEndian.PutUint32(buf[:], num)
+		net++
```
gconnell, line 183: rather than this, how abut make line 181 be:

for net++; net<broadcast; net++ {
  ...
}

jonasbostoen: Right! Missed that completely, thanks.

examples/arpscan/arpscan.go
```diff
@@ -178,9 +178,8 @@ func ips(n *net.IPNet) (out []net.IP) {
 	mask := binary.BigEndian.Uint32([]byte(n.Mask))
 	net := num & mask
 	broadcast := net | ^mask
-	for net++; net < broadcast-1; {
+	for net++; net < broadcast-1; net++ {
```
gconnell, line 181: given this new approach, I don't think you want "broadcast - 1" anymore; you won't hit .254 in a /24 with this code.

jonasbostoen: Fixed it!

gconnell:
THanks!

PR #855, opened 2021-02-06
Fix pointer error in for loop

mashaz:
All addresses of loop variable are the same. So it's always return the wrong route if the MAC is provided.


gconnell:
Thanks!

Issue #860, opened 2021-03-01
FastHash Collision Rate High for Layer4 Flows


upsampled:
I am building a small analysis program to produce something like an annotated Wireshark "Follow TCP Stream" output by merging the two directional streams of a TCP conversation. My hope was that using the FashHash (Layer 3 and 4) output I could map the two directional flows together; however, I experienced a problem with my very small (three tcp stream) test pcap file. Two of the three TCP conversations had the same Fast Hash values, but used different ports:

* server port 7028 and client port 64717 result `a0d19934c49f0278`
* server port 7029 and client port 64718 result `a0d19934c49f0278`

I initially thought I did something wrong, so I wrote [test ](https://play.golang.org/p/MrTxDmFC6N_x)code that showed the collision

My goal in pointing this out is to:
* make sure the documentation mentions FastHash is at least not currently collision resistant with other Flow sets 
* propose that either FastHash is updated or a another hash method be added (PrettyFastHash, FastishHash, etc)  that is collision resistant.

secmobi:
_disclaimer: I'm only a gopacket user like you, not the maintainer._ 

I think current doc has **implicitly** suggested this. In the project's general introduction on usages, it says and here I quote: 

> For load-balancing purposes, both Flow and Endpoint have FastHash() functions, which provide quick, non-cryptographic hashes of their contents.

And in [the function's description](https://pkg.go.dev/github.com/google/gopacket#Flow.FastHash) it mentioned similar idea again that: 

> FastHash provides a quick hashing function for a flow, useful if you'd like to split up flows by modulos or other load-balancing techniques.

That's been said, I believe a more **explicit** note would be necessary, to emphasize that it's improper to use the function's return values for session/flow's unique key purpose. 

On the second part of your suggestion on finding other collision resistant hash, I feel a paradox here: the hash shall be commutative (i.e., output the same hash for flow of A -> B and for flow B -> A therefore two flows in the same session will have the same hash values); however, a commutative hash won't be a cryptographic hash function due to the characteristic of h(A+B) == h(B+A).

upsampled:
After digging into this more and writing a brute forcer, the (FNV+FNV) hash performs worse than expected for the Layer 4 order-independent, port-pair combinations. I am seeing approx 5.5K collisions for every combination. I believe this is mostly due to the single FNV hash function `fnvHash` performing poorly with small inputs. For example:

1. Given port 1 the hash is `8328707b4eb6e3a`
[...]

PR #864, opened 2021-03-09
New fasthash that is collision free for all but IPv6/MAC.

upsampled:
Addresses #860.

The exiting FNV algorithm used by the FastHash methods is a very fast but has poor diffusion. This is especially apparent when dealing with Layer 4 Flows that use 2-byte ports. Writing a crude brute forcer, I was finding approximately 5.5K collisions for each port pair.

Given that FastHash returns a `uint64` and the vast majority endpoints have 32-bits or less addresses, a more direct mapping can produce collision free hashes. One way to do this is breaking the address into `uint`s and xoring them together. For Endpoint's FastHash this is done with `uin64`s. For Flow's FashHash this is done to each address with `uint32`s which as then concatenated together in an order independent way.

### DIFF
--- a/flows.go
+++ b/flows.go
@@ -54,31 +54,18 @@ func (a Endpoint) LessThan(b Endpoint) bool {
 	return a.typ < b.typ || (a.typ == b.typ && bytes.Compare(a.raw[:a.len], b.raw[:b.len]) < 0)
 }
 
-// fnvHash is used by our FastHash functions, and implements the FNV hash
-// created by Glenn Fowler, Landon Curt Noll, and Phong Vo.
-// See http://isthe.com/chongo/tech/comp/fnv/.
-func fnvHash(s []byte) (h uint64) {
-	h = fnvBasis
-	for i := 0; i < len(s); i++ {
-		h ^= uint64(s[i])
-		h *= fnvPrime
-	}
-	return
-}
-
-const fnvBasis = 14695981039346656037
-const fnvPrime = 1099511628211
-
 // FastHash provides a quick hashing function for an endpoint, useful if you'd
 // like to split up endpoints by modulos or other load-balancing techniques.
-// It uses a variant of Fowler-Noll-Vo hashing.
+// It creates the hash by xor-ing the bytes of the address in a 64-bit, round robin
+// fashion (hash ^= addr[0],hash ^= addr[1]<<8,...,hash ^= addr[9]<<8). For
+// addresses less than or equal to 64-bits this guarantees  collision free hashes.
 //
 // The output of FastHash is not guaranteed to remain the same through future
 // code revisions, so should not be used to key values in persistent storage.
 func (a Endpoint) FastHash() (h uint64) {
-	h = fnvHash(a.raw[:a.len])
-	h ^= uint64(a.typ)
-	h *= fnvPrime
+	for i := 0; i < a.len; i++ {
+		h ^= uint64(a.raw[i]) << (8 * (uint(i) % 8))
+	}
 	return
 }
 
@@ -158,18 +145,29 @@ func FlowFromEndpoints(src, dst Endpoint) (_ Flow, err error) {
 
 // FastHash provides a quick hashing function for a flow, useful if you'd
 // like to split up flows by modulos or other load-balancing techniques.
-// It uses a variant of Fowler-Noll-Vo hashing, and is guaranteed to collide
-// with its reverse flow.  IE: the flow A->B will have the same hash as the flow
-// B->A.
+// It  is guaranteed to collide with its reverse flow.  IE: the flow A->B
+// will have the same hash as the flow B->A.
+//
+// It creates the hash by first, for each endpoint, xor-ing the bytes of the
+// address in a 32-bit round robin fashion (hash ^= addr[0],hash ^= addr[1]<<8,
+// ...,hash ^= addr[5]<<8). The result for each endpoint is then concatenated
+// together in an order independent way. For addresses less than or equal to
+// 32-bits this guarantees collision free hashes. IE: the flows A->B and B->A
+// will have a hash that no other combination of addresses can produce.
 //
 // The output of FastHash is not guaranteed to remain the same through future
 // code revisions, so should not be used to key values in persistent storage.
-func (f Flow) FastHash() (h uint64) {
-	// This combination must be commutative.  We don't use ^, since that would
-	// give the same hash for all A->A flows.
-	h = fnvHash(f.src[:f.slen]) + fnvHash(f.dst[:f.dlen])
-	h ^= uint64(f.typ)
-	h *= fnvPrime
+func (f Flow) FastHash() (a uint64) {
+	var b uint64
+	for i := 0; i < f.slen; i++ {
+		a ^= uint64(f.src[i]) << (8 * (uint(i) % 4))
+		b ^= uint64(f.dst[i]) << (8 * (uint(i) % 4))
+	}
+	if a > b {
+		a += (b << 32)
+		return
+	}
+	a = b + (a << 32)
 	return
 }
 
--- a/flows.go
+++ b/flows.go
@@ -159,15 +159,10 @@ func FlowFromEndpoints(src, dst Endpoint) (_ Flow, err error) {
 // code revisions, so should not be used to key values in persistent storage.
 func (f Flow) FastHash() (a uint64) {
 	var b uint64
-	var addrlen int
-	if f.slen > f.dlen {
-		addrlen = f.slen
-	} else {
-		addrlen = f.dlen
-	}
-	for i := 0; i < addrlen; i++ {
-		a ^= uint64(f.src[i]) << (16 * (uint(i) % 4))
-		b ^= uint64(f.dst[i]) << (16 * (uint(i) % 4))
+
+	for i := 0; i < f.slen; i++ {
+		a ^= uint64(f.src[f.slen-1-i]) << (16 * (uint(i) % 4))
+		b ^= uint64(f.dst[f.slen-1-i]) << (16 * (uint(i) % 4))
 	}
 	if a > b {
 		a += (b << 8)
--- a/flows.go
+++ b/flows.go
@@ -64,7 +64,7 @@ func (a Endpoint) LessThan(b Endpoint) bool {
 // code revisions, so should not be used to key values in persistent storage.
 func (a Endpoint) FastHash() (h uint64) {
 	for i := 0; i < a.len; i++ {
-		h ^= uint64(a.raw[i]) << (8 * (uint(i) % 8))
+		h ^= uint64(a.raw[a.len-1-i]) << (8 * (uint(i) % 8))
 	}
 	return
 }
--- a/flows.go
+++ b/flows.go
@@ -159,7 +159,13 @@ func FlowFromEndpoints(src, dst Endpoint) (_ Flow, err error) {
 // code revisions, so should not be used to key values in persistent storage.
 func (f Flow) FastHash() (a uint64) {
 	var b uint64
-	for i := 0; i < f.slen; i++ {
+	var addrlen int
+	if f.slen > f.dlen {
+		addrlen = f.slen
+	} else {
+		addrlen = f.dlen
+	}
+	for i := 0; i < addrlen; i++ {
 		a ^= uint64(f.src[i]) << (8 * (uint(i) % 4))
 		b ^= uint64(f.dst[i]) << (8 * (uint(i) % 4))
 	}
--- a/flows.go
+++ b/flows.go
@@ -166,14 +166,14 @@ func (f Flow) FastHash() (a uint64) {
 		addrlen = f.dlen
 	}
 	for i := 0; i < addrlen; i++ {
-		a ^= uint64(f.src[i]) << (8 * (uint(i) % 4))
-		b ^= uint64(f.dst[i]) << (8 * (uint(i) % 4))
+		a ^= uint64(f.src[i]) << (16 * (uint(i) % 4))
+		b ^= uint64(f.dst[i]) << (16 * (uint(i) % 4))
 	}
 	if a > b {
-		a += (b << 32)
+		a += (b << 8)
 		return
 	}
-	a = b + (a << 32)
+	a = b + (a << 8)
 	return
 }
 


gconnell:
Thanks a ton for this!  Before I go into specific review of the code, I've got another suggestion that I think might be worth exploring:  https://golang.org/pkg/hash/maphash/

IE: for flow:
```
  var h maphash.Hash
  h.Write(f.raw[:f.len])
  return h.SumUint64()
```

and for endpoint
```
  var ah, bh maphash.Hash
  ah.Write(f.src[:f.slen])
  bh.Write(f.dst[:f.dlen])
  a, b := ah.SumUint64(), b.SumUint64()
  return a ^ b  // or some other communicative operation
```

upsampled:
@gconnell I looked into trying to use other libraries like HashMap, but I could not seem to use any without a relatively large performance hit due to the FastHash functions being over the inline budget. 

For example, using HashMap, even skipping the seed step, still has a cost well over the max inline budget.
```
$go build -gcflags='-m=2' ./ 2>&1 | grep flows\.go | grep inlin | grep Fast
./flows.go:68:6: cannot inline Endpoint.FastHash: function too complex: cost 133 exceeds budget 80
```
This is reflected in the benchmarks
```
cpu: Intel(R) Core(TM) i7-4510U CPU @ 2.00GHz
BenchmarkEndpointFastHashShort-4        24000384                47.18 ns/op
BenchmarkEndpointFastHashLong-4         26994243                44.77 ns/op
```

The original FastHash implementations were both under the inline budget, so I kept this design choice and was able to do the same with this PR:
```
./flows.go:65:6: can inline Endpoint.FastHash with cost 28 as: ...
./flows.go:160:6: can inline Flow.FastHash with cost 63 as: ...
```
This PR's benchmark:
```
BenchmarkEndpointFastHashShort-4        306184075                3.912 ns/op
BenchmarkEndpointFastHashLong-4         72255205                14.43 ns/op
BenchmarkFlowFastHashShort-4            535044524                2.242 ns/op
BenchmarkFlowFastHashLong-4             79395997                15.05 ns/op
``` 
The original FastHash benchmarks:
```
BenchmarkEndpointFastHashShort-4        650002077                1.945 ns/op
BenchmarkEndpointFastHashLong-4         130605961                8.848 ns/op
BenchmarkFlowFastHashShort-4            355440339                3.579 ns/op
BenchmarkFlowFastHashLong-4             54367524                20.02 ns/op
```

flows.go
```diff
@@ -158,18 +145,29 @@ func FlowFromEndpoints(src, dst Endpoint) (_ Flow, err error) {
...
-func (f Flow) FastHash() (h uint64) {
-	// This combination must be commutative.  We don't use ^, since that would
-	// give the same hash for all A->A flows.
-	h = fnvHash(f.src[:f.slen]) + fnvHash(f.dst[:f.dlen])
-	h ^= uint64(f.typ)
-	h *= fnvPrime
+func (f Flow) FastHash() (a uint64) {
+	var b uint64
+	for i := 0; i < f.slen; i++ {
+		a ^= uint64(f.src[i]) << (8 * (uint(i) % 4))
```
gconnell, line 163: I'm worried about this approach for large values, specifically IPv6 addresses.  As well as being > 4 bytes, they're also somewhat likely to have some "structure" to their bits set/unset.  For example, I could see a network having blah:1:1:1:1 and blah:2:2:2:2, which would evaluate to the same hash due to the XOR canceling out the first 1:1 with the second, and doing the same for the first/second 2:2.

Could we go with your original approach of just doing the 32-bit method for values <= 4 bytes?  I think it might be easily simplified as:

```
if f.slen <= 4 && f.dlen <= 4 {
  // LittleEndian in case only the first or first 2 bits are set.  Assumes bits after xlen are all zeros.
  a := binary.LittleEndian.Uint32(a.src[:4])
  b := binary.LittleEndian.Uint32(a.dst[:4])
  if a > b { a, b = b, a }
  return uint64(a) << 32 | uint64(b)
}
```

flows.go
```diff
@@ -158,18 +145,29 @@ func FlowFromEndpoints(src, dst Endpoint) (_ Flow, err error) {
...
 // code revisions, so should not be used to key values in persistent storage.
-func (f Flow) FastHash() (h uint64) {
-	// This combination must be commutative.  We don't use ^, since that would
-	// give the same hash for all A->A flows.
-	h = fnvHash(f.src[:f.slen]) + fnvHash(f.dst[:f.dlen])
-	h ^= uint64(f.typ)
-	h *= fnvPrime
+func (f Flow) FastHash() (a uint64) {
+	var b uint64
+	for i := 0; i < f.slen; i++ {
```
gconnell, line 162: This makes an assumption that `slen == dlen`, which AFAIK is 100% correct right now but may not be future-proof.  Might be safer to code around the fact that that may not be the case in the future.

flows.go
```diff
@@ -54,31 +54,18 @@ func (a Endpoint) LessThan(b Endpoint) bool {
...
+// fashion (hash ^= addr[0],hash ^= addr[1]<<8,...,hash ^= addr[9]<<8). For
+// addresses less than or equal to 64-bits this guarantees  collision free hashes.
 //
 // The output of FastHash is not guaranteed to remain the same through future
 // code revisions, so should not be used to key values in persistent storage.
 func (a Endpoint) FastHash() (h uint64) {
-	h = fnvHash(a.raw[:a.len])
-	h ^= uint64(a.typ)
-	h *= fnvPrime
+	for i := 0; i < a.len; i++ {
```
gconnell, line 66: effectively same comments as below, here (sorry, I'm reviewing this upside down):

```
if a.len <= 8 {
  return binary.LittleEndian.Uint64(a.raw[:8])
}
// keep old fnv for larger values.
```

flows.go
upsampled: Completely understand, there are two main issues: 
1. Inline budget 
2. The Hash collision cases

**Inline budget**
If both FNV and Direct Hashing are implemented, it will throw the FastHashes compile cost over the inline budget and give them a 10x slowdown. I have tried every way I can think of to combine them and have not even gotten close. Adding the change you suggested results in:
```
./flows.go:175:6: cannot inline Flow.FastHash: function too complex: cost 146 exceeds budget 80
```

**The Direct Hash collision cases**
You're right with regards to Flow cases:
* `fdff::1:1:1:1` ->`fdff::2:2:2:2`
* `fdff::2:2:2:2` ->`fdff::3:3:3:3`

My only argument is that the FNV still has issues with IPv6 flows as well. They aren't as structural but they occur in common analysis. I am sure I can find a collision with addresses closer together, but just porting my initial TCP Port collision from #860 to IPv6:
* `fdff::1b74` -> `fdff::fccd`
* `fdff::1b75` -> `fdff::fcce`

It is very hard for these smaller hash functions to combat all the different ways you can deploy IPv6 and are forced to analyze it. The best way may be to just handle IPv6 hashing in the `layer` package and allow for different Flow Label functions for different IPv6 situations.

flows.go
upsampled: Assuming the byte array is all `0`s above the `f.slen` or `f.dlen` index, I can do the following and *just* be at the inline limit:
```
func (f Flow) FastHash() (a uint64) {
	var b uint64
	var alen int
	if f.slen > f.dlen {
		alen = f.slen
	} else {
		alen = f.dlen
	}
	for i := 0; i < alen; i++ {
...
``
```
./flows.go:160:6: can inline Flow.FastHash with cost 80 as:...
````
Will that work?

File: flows.go (hunk 3 above)
Comment (upsampled) [reply flows.go:66]: see first thread

**Comment by upsampled** [comment]
Summarizing the feedback and what I have learned looking at this, I think these are the paths forward:

**Keep FastHash super fast and start adding protocol specific-hashing in the layers package**
Keeping the FastHash methods inlined basically requires only one algorithm (FNV or this hash) be used. The algorithm, regardless of which one, will have pros and cons. Considering the one in the PR is ideal for everything but IPv6 (and kinda MAC), I believe it better suites this purpose. To compensate for the cons of any particular FastHash implementation, per protocol Hash functions can be added to the `layers` package. 
```
package layers
...
// if this hash is used
//IPv6FlowHash is an IPv6 specific FastHash implementation
func IPv6FlowHash(f *gopacket.Flow) (h uint64){
	if f.EndpointType() != EndpointIPv6{
		return
	}
	...
	return
}
//if we keep with FNV
func Layer4PortsFlowHash(f *gopacket.Flow) (h uint64){
       typ := f.EndpointType()  
	if typ < EndpointTCPPort || typ >= EndpointPPP {
		return
	}
	...
	return
}
```

**Allow FastHash to just be fast, default to a hybrid hash alg and allow protocol specific hashing to be registered to an EndpointType**
If we abandon the need for FastHash to be inlined, we can do the hybrid hash algorithm as the default case and also allow for the registration of per protocol FastHash implementations. The additional logic does not greatly impact the performance after the initial non-inlined slowdown. It would work very much like `Formatter`:
```
package gopacket
...
// EndpointTypeMetadata is used to register a new endpoint type.
type EndpointTypeMetadata struct {
	// Name is the string returned by an EndpointType's String function.
	Name string
	// Formatter is called from an Endpoint's String function to format the raw
	// bytes in an Endpoint into a human-readable string.
	Formatter func([]byte) string
	// FastHash is called from Endpoint's FastHash function to uniquely identify
	// an Endpoint. Flow's FashHash function will only use the 32 least significant bits
	// of this function's output.
	FastHash func([]byte, int) uint64
}
```
```
package layers
...
EndpointIPv6 = gopacket.RegisterEndpointType(2, 
    gopacket.EndpointTypeMetadata{
        Name: "IPv6", 
        Formatter: func(b []byte) string {return net.IP(b).String()}, 
        FastHash: fastHashIPv6Endpoint
    })
```

**Comment by gconnell** [comment]
Looking at this, I've got one final concern, and I'd love to hear your thoughts.

Here's how I'd naively expect this function to be used most of the time:

```
var handlers [SIZE]handler
for _, x := range endpointsOrFlows {
  handlers[x % len(handlers)].Handle(x)
}
```

With the new, xor-based implementation, there's a good chance for this to go pretty wrong.  For example, if SIZE is a power of 2 (which, since programmers chose it, there's a good chance it is), with the XOR version, all of the bits will be ignored.  Considering, for example, a flow on ports.  If the lower port is packed first, a modulo will effectively strip the upper bits entirely, and all packets will "load balance" based on the serving port.  So all DNS traffic will go to [53], for example, because the bits 0x??????????000035 % 1024 will always equal 53.

So, I guess here's my challenge:  we've been focussing on "hash value is unique".  I'd like to instead focus on "hash value is spread such that low-modulus outputs of it are spread out."  Less on FastHash(x) == FastHash(y), and more on FastHash(x)%SIZE being uniform for SIZE<1M, and specifically for SIZE=x^2.

Your thoughts on this greatly appreciated.

**Comment by gconnell** [comment]
Also, I realized I'm super dumb.  maphash is an awful thing to use, because you can just do map[Endpoint]x, and you get that :D.  That said, this is an interesting consideration.  If you're looking to map uniquely, why not just map[Endpoint] or map[Flow]?  FastHash is really designed for cases where that's too slow, mostly for these "load balance across a set of handlers" approaches.

**Comment by upsampled** [comment]
@gconnell not sure I 100% follow the example, but let me try and you can correct me:

It looks like you are trying to load balance using a fixed number of handlers and the randomness of FastHash to assign work to them (I added the `x.FastHash()` to your original snippet . 

```
var handlers [SIZE]handler
for _, x := range endpointsOrFlows {
  handlers[x.FastHash() % len(handlers)].Handle(x)
}
```
So for starters without concurrency we won't be able to load balance  because the code above is single threaded. Making the least amount of change, we can change the handlers to channels and assume SIZE number of workers workers are already launch:
```
var handlers [SIZE]chan *Flow
...
for i,_ := range handlers{
 go doFirewallWork(i,handlers[i])
}
...
for i, _ := range flows {
  handlers[flows[i].FastHash() % len(handlers)]<-&flows[i]
}
```
Ok

**Comment by gconnell** [comment]
Yep, exactly.  In my example, the implementation for Handle() would include
farming the result across threads or some other load-balancing mechanism.
It could be:

func (h *handler) Handle(x) {
  h.inputChan <- x
}

or

func (h *handler) Handle(x) {
  forwardPacketTo(h.downstreamAddress, x)
}

or some other mechanism for downstream load balancing.

I guess the higher level question is:

if map[Flow]x and map[Endpoint]x, which guarantees uniqueness entirely, is
too slow for your use case, then what is your use case?  The use case that
I believe FastHash was initially used for was the channel-based approach
you mentioned.

**Comment by gconnell** [comment]
Specifically, note the docs here:
https://pkg.go.dev/github.com/google/gopacket#hdr-Flow_And_Endpoint

**Comment by upsampled** [comment]
@gconnell I think I figured out a solution for this: interleave the hashes.

So initially the flow hash was:
```
0x00000000AAAAAAAA
0x00000000CCCCCCCC
___________________________
0xAAAAAAAACCCCCCCC
```
But if we interleave the two, we can avoid the issue you mention, and best of all leverage the high entropy, least significant bits for the endpoint hashes to the least significant bit of the flow hash. 
```
0x00AA00AA00AA00AA
0x00CC00CC00CC00CC
___________________________
0xAACCAACCAACCAACC
```

upsampled:
For use-case, I am interested in conversation analysis, which can be implemented in the load-balancing technique mentioned.

The simplest example would be something like NetFlow analysis. Basically trying to count packets between application. More advanced examples would be to combine the two directional tcp streams outputted from `reassembly` to be able to see both Client to Server and Server to Client Traffic in real-time. 

The more general use-case it to quickly place a packet in a context related to two application in both directions.

upsampled:
actually. I need to reverse the order I evaluate the addresses as addr[15] is the least significant bit of an IP address.

upsampled:
Ok so here is where the branch is at:
* Endpoint Hash and Flow Hash are collision free for under 32-bit addresses
* Load-Balancing is now covered by using interleaving and evaluating the parts of the endpoint addresses that change the most first and putting them in the least significant bits of the output hash.

I did have to remove the `slen != dlen` check to stay under the inline budget. This was a 'what if' case and the fasthash will still produce a usable value.

upsampled:
Just ran some tests and this hash performs better on the `ephemeral port -> DNS Port` test cases, but I imagine FNV is going to give better results because it mainly produces randomness in the least significant bytes of the hash. That leaves the following questions:

1. Should FastHash emphasize randomness of  uniqueness? 
I think you showed that is was created for emphasis on randomness, specifically for load balancing.
2. Should the Flow/Endpoint have a hash function that emphasizes uniqueness (`FastLabel`,`FlowLabel`)?
It would allow for a pretty efficient pairing with FNV FastHash. FNV Hash to allow load balancing and then Label to perform conversation analysis.
3. Or should there be something that better labels sessions likes Firewall Tuples (Layer3 and Layer 4 -> ID)? 
The 5-Tuple is what I mainly think of when performing analysis. Not especially fast but foolproof. That could just be implemented in another like `layers` or a separate repo.

PR #867, opened 2021-04-01
add GetIPV4Routes,GetIPV6Routes function

njuptlzf:
Currently, after` routing.New()`, the **v4,v6 routing table info** is only used for `String()` or internal use.
It makes sense to increase the interface to obtain ipv4 and ipv6 routing table information separately.
For example, I use it to compare whether the ip segment is in the corresponding routing table.


njuptlzf:
@googlebot I signed it!

njuptlzf:
I would be happy for any friend who has passed CLA to approve and submit the PR  :)

PR #869, opened 2021-04-07
layers: use map instead of 64k array for tcpPortLayerType and udpPortLayerType

tklauser:
`tcpPortLayerType` and `udpPortLayerType` each unconditionally allocate 64k.
This happens when importing the package (i.e. at init time) even if the
functionality is never used. Instead, use a `switch` for the known layer
types and maps to store the registered overrides. Additionally use a
bitfield, storing whether a particular port holds an override to speed
up the case of not-matching-existing layers.

/cc @gconnell

### DIFF
--- a/layers/bitfield.go
+++ b/layers/bitfield.go
@@ -1,0 +1,18 @@
+// Copyright 2021 The GoPacket Authors. All rights reserved.
+//
+// Use of this source code is governed by a BSD-style license that can be found
+// in the LICENSE file in the root of the source tree.
+
+package layers
+
+type bitfield [1024]uint64
+
+// set sets bit i in bitfield b to 1.
+func (b *bitfield) set(i uint16) {
+	b[i/64] |= (1 << (i % 64))
+}
+
+// has reports whether bit i is set to 1 in bitfield b.
+func (b *bitfield) has(i uint16) bool {
+	return b[i/64]&(1<<(i%64)) != 0
+}
--- a/layers/bitfield_test.go
+++ b/layers/bitfield_test.go
@@ -1,0 +1,26 @@
+package layers
+
+import "testing"
+
+func TestBitfield(t *testing.T) {
+	var b bitfield
+
+	const uint16Max = ^uint16(0)
+
+	for i := uint16(0); i < uint16Max; i++ {
+		if b.has(i) {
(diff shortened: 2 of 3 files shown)


gconnell (commented):
A map will be measurably slower for high-throughput applications than an array access.  However, I would be supportive of moving this into a switch statement within each LayerType() function instead of having an array, if you'd be amenable to that approach.

tklauser:
Thanks for the feedback. Moving this to switch statement within each `LayerType()` func sounds good to me. I'll adjust the PR accordingly.

tklauser:
Unfortunately, it looks like changing to a switch statement will not work for `RegisterTCPPortLayerType`/`RegisterUDPPortLayerType`. Unless we introduce a map for the ports registered using these functions, i.e. along the lines of:

```go
var tcpPortLayerType = map[UDPPort]gopacket.LayerType{}

func (a TCPPort) LayerType() gopacket.LayerType {
         switch a {
         case 53:
                 return LayerTypeDNS
         // ... other cases
         }

         lt, ok := tcpPortLayerType[a]
         if ok {
                 return lt
         }
         return gopacket.LayerTypePayload
}
```

I see no in-tree users of those and a quick GitHub search doesn't show many users of these functions, so maybe this would be acceptable? @gconnell

gconnell:
Hmm, that still makes the possibly common case of not-matching-existing a bit slower, and it would also disallow overrides, which was one fo the primary reasons (that I totally forgot about) for RegisterX.

How would you feel about something like:

```
var []bitfield (64k bits)

if bitfield.has(port) {
  return map[port]
}
switch port {
  case 53: LayerTypeDNS
}
return nil

func RegisterX(port, LT) {
  bitfield.set(port)
  map[port] = LT
}
```

That would mean that the map is generally small, only checked if there's guaranteed a value in it, switch stuff gets overridden, and life is all happy rainbows.  Thoughts?

tklauser:
Thanks a lot for the suggestion @gconnell! This sounds like a good trade-off solution to me. I've pushed a new version of this PR implementing your suggested approach, PTAL.

gconnell:
Actually, those are such small changes, I'm happy to merge as is then add them in.  Let me do that.

gconnell:
threw them into https://github.com/google/gopacket/pull/871

PR #871, opened 2021-04-09
Add a stress-test for bitfield, and use fast ops.

gconnell:
I expect the compiler would handle the fast ops for us regardless most
of the time, but doing the bit-magic anyway to be safe.


tklauser (approved):
LGTM. Feel free to disregard #872 and merge this one instead.

PR #878, opened 2021-04-30
Remove the fmt.Println call in DecodeFromBytes of GTPv1U

kery:
The fmt.Println should not be used in the library for debugging purpose. The returned error already contains the information of the failure.


gconnell:
Woah, how'd that get there?   Thanks!

PR #883, opened 2021-06-04
udp: Encode 0 checksums as 0xFFFF

arthurfabre:
RFC768 states:

    If the computed checksum is zero, it is transmitted as all ones (the
    equivalent in one's complement arithmetic). An all zero transmitted
    checksum  value means that the transmitter generated no checksum.

Updated the serialization logic to correctly handle this. It isn't a big
problem for IPv4 where checksums are optional, but affects IPv6 which
has mandatory UDP checksums.

### DIFF
--- a/layers/udp.go
+++ b/layers/udp.go
@@ -89,6 +89,12 @@ func (u *UDP) SerializeTo(b gopacket.SerializeBuffer, opts gopacket.SerializeOpt
 		if err != nil {
 			return err
 		}
+		// RFC768: If the computed checksum is zero, it is transmitted as all ones (the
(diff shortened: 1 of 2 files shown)


MarcusWichelmann:
Is there a reason this branch hasn't been merged yet?
I'm currently working on refactoring some things about the checksum calculation and also adding functions to verify checksums of decoded packets, and would need to base my work on this PR. Would be great, if this could be merged beforehand. Then my PR(s) will follow.

Also the changes in this PR look all valid IMO.

joewilliams:
Please merge this PR.

PR #887, opened 2021-06-22
BFD: check lengths when decoding auth header

matzf:
Check that the length field of the authentication header matches the
data size and check that the slice is long enough before reading the keyed
MD5/SHA1 authentication section.
Note that the BFD specification (RFC 5880) restricts the valid size of
the authentication header "Data" field for all types, but this is (still) not enforced.

This fixes a "slice bounds out of range" panic when decoding the
Sequence Number field for the keyed MD5/SHA1 authentication section
on truncated messages.

Found with [go-fuzz](https://github.com/dvyukov/go-fuzz) in [scionproto/scion](https://github.com/scionproto/scion).


matzf:
Friendly ping @gconnell. Let me know if I can help in any way with getting this PR reviewed.

PR #902, opened 2021-08-15
Fixed wrong var name in docs

icamys:


icamys:
@googlebot I signed it!

PR #903, opened 2021-08-19
Support for Windows/ARM64

hdm:
This (very) simple patch allows gopacket to work with windows/arm64 builds. This requires a newish version of npcap (1.50+).

This also requires Go 1.17.


clarkmcc:
Any plans to support windows/arm64?

hdm:
This should probably move to the fork (github.com/gopacket/gopacket)

PR #911, opened 2021-09-25
Make pcapgo EthernetHandle use runtime poller

WGH-:
As of now, the `EthernetHandle` fd is kept in the blocking mode. Which means when capturing traffic, entire OS thread is blocked in `recvmsg` syscall most of the time.

It's possible to use custom fds (i.e. not created by `net` package) with the runtime poller, so using them will be as efficient as it is with e.g. TCP sockets.

This PR does exactly that.


mosajjal:
hey mate. I'm happy to merge this into the `gopacket/gopacket` fork. feel free to raise a PR over there.

Issue #916, opened 2021-10-25
TLS Handshake bytes


89z:
I have this code:

~~~go
package main

import (
   "encoding/hex"
   "fmt"
   "github.com/google/gopacket"
   "github.com/google/gopacket/layers"
)

const ipv4 = "45000181000040004006c622acd90ea30ad7ad0101bb963a77eb888ebc0fefd650182000b49c000016030300ec040000e8000189c000e20181707f176e28362f0c47652928fd581b9edd207ab9d5181cb03a39839449dc45dd78b377fc75a00877e38985211d3f4d18d9fa08328b98a687b233f2325a8dbe03b0270c901eefd308434c3719f1bed23e075e21e17e37155a77d9f52a7a71a663190124ce0b09e3a64afd181975f733748cc6112646c862c571a6d897672d0ec045fa13ffc3d48a3964767ccd6e5587e79f86f95fe634ec88aaea4c8868a3cf70fdb4d4bb8e0cc4335fe4b52e9f3c9f6b76f050f95bbdd52f769e9d1266add50875b1d5682cf53642cb7ace3d1fcaa4743b973fa0512a7f456bfe4396e21f1cc31403030001011603030020b0aae794fe71c67c2dbde28684f19df7e07984a3644ee6635abbefc28989dd041703030038f55a18e4017989eecdffcba058a557543fa5d277badbd606fa08a973c3717cfd3539f9a2375f0be478de6985c61fc829bd566f6dcfb65438"

func main() {
   data, err := hex.DecodeString(ipv4)
   if err != nil {
      panic(err)
   }
   packet := gopacket.NewPacket(data, layers.LayerTypeIPv4, gopacket.Default)
   payload := packet.TransportLayer().LayerPayload()
   var layer layers.TLS
   layer.DecodeFromBytes(payload, gopacket.NilDecodeFeedback)
   fmt.Printf("%#v\n", layer)
}
~~~

Where the input is a single IPv4 packet with a TLS Layer (345 bytes), and four TLS Records:

Record                      | bytes
----------------------------|------
New Session Ticket          | 241
Change Cipher Spec          | 6
Encrypted Handshake Message | 37
Application Data            | 61

Application Data gets saved as `.AppData` and `.BaseLayer`, and Change Cipher Spec is saved as `.ChangeCipherSpec`. But New Session Ticket is just gone, and you only get the Handshake metadata, rather than the actual Handshake bytes. I am needing to extract Handshake bytes from all packets, is that possible?

[...]

Issue #289, opened 2017-01-26
LinkType enum is uint8 but tcpdump defines linktypes > 255


PaulMcMillan:
The LinkType type is a uint8:
https://github.com/google/gopacket/blob/master/layers/enums.go#L93
However, the canonical list of linktype numbers currently has a max of 266. 
http://www.tcpdump.org/linktypes.html

Additionally, looking at the pcap documentation, it looks like that is returned as an int
http://www.tcpdump.org/manpages/pcap_datalink.3pcap.txt
and when we use that function, we also use an int
https://github.com/google/gopacket/blob/master/pcap/pcap.go#L651
Edit: oh, I see how that works. The `layers.LinkType()` is casting the int to uint8.

I'm also not sure how to solve this in a fully backwards compatible way, but I figured I should bring it up since I noticed it...

lhausermann:
Hi. Thanks for reporting it. Uint8 are used to optimize the lookup per the number of layer we actually support. I think that would be updated once we'll see the need for it.

PR #933, opened 2022-02-02
Add SLL2 layer decoding

dylandreimerink:
This pull request adds SLL2(Linux "cooked" capture encapsulation v2) layer decoding. SLL2 seems to be the preferred encapsulation method used by `tcpdump` when capturing packets in cooked mode("-i any").

The `LinkType` value for [SLL2](https://www.tcpdump.org/linktypes/LINKTYPE_LINUX_SLL2.html) is 276, according to https://www.tcpdump.org/linktypes.html. This exceeds the max value of `layers.LinkType` which is a `uint8`. 

To accommodate link types above 255 we have to change `layers.LinkType` from a `uint8` to a `uint16`. From what I have read in existing issues(#916 #289), the downside of doing this is the added memory usage of the `layers.LinkTypeMetadata` array. Just changing the type `layers.LinkTypeMetadata` is not a good idea for high performance decoding. Changing `layers.LinkType` from a `uint8` to a `uint16` and increasing the size of `layers.LinkTypeMetadata` to 65535 could cause out of bounds access which is also not acceptable.

This PR modifies `layers/gen2.go` to now generate bounds checks for the `Decode`, `String` and `LayerType` methods on the enum types, which prevents panics when a enum value is out of bounds of a `...Metadata` array, allowing the arrays to be smaller than all possible values of the enum type.

This does make decoding a little bit slower(about 5% judging by the total benchmark times) but we don't have to allocate memory which will not be used. Since `gen2.go` generates code for all enums, this penalty is shared across all of them. However, the benefits can also be shared by all of them. While this PR doesn't modify more than necessary, the `EthernetType` could benefit from shrinking the Metadata array to `0x9001`(36865) saving 28672 unused values from being allocated. Same goes for the `PPPType` which can save 64893 values since only up to 0x0283(643) is used.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,6 +34,7 @@ Kane Mathers <kane@kanemathers.name>
 Jose Selvi <jselvi@pentester.es>
 Yerden Zhumabekov <yerden.zhumabekov@gmail.com>
 Jensen Hwa <jensenhwa@gmail.com>
+Dylan Reimerink <97.dylan@gmail.com>
 
 -----------------------------------------------
 FORKED FROM github.com/akrennmair/gopcap
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.12
 require (
 	github.com/vishvananda/netlink v1.1.0
 	github.com/vishvananda/netns v0.0.0-20210104183010-2eb08e3e575f
-	golang.org/x/lint v0.0.0-20200302205851-738671d3881b // indirect
+	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
 	golang.org/x/net v0.0.0-20190620200207-3b0461eec859
 	golang.org/x/sys v0.0.0-20200217220822-9197077df867
 )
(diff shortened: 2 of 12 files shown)


wader:
@dylandreimerink Hello, was this fixed in some other branch? don't see it in master. Or was there something wrong with the PR?

dylandreimerink:
I closed it because I was cleaning up my list of open PRs so I have a neat overview of PRs to comment on or fix. This PR has been open for 4 months now without review or comment, so I sort of assumed it would never get merged anyway. But I am happy to reopen it.

wader:
Ok make sense. I noticed that there was some activity in #1016 which made me hopefully and adding SLL2 support looks quite straightforward and would be very useful. Currently i use some hacky byte mangling to transform SLL2 into SLL that i would like to get rid of.

dylandreimerink:
Feel free to use my fork or to cherry pick the commits into your own fork. I hope we will get some active maintainers on this project, would be a shame if the only way forward would be to fork it permanently.

wader:
Yeap agree and let's hope so. I pinged about your PR in the project status PR.

wader:
@dylandreimerink Hey, a modified version of this PR got merge https://github.com/gopacket/gopacket/pull/1 🥳

PR #1046, opened 2022-08-07
Fix two panics identified by fuzz testing

cirego:
This commit adds additional bounds checking, one for DNSResourceRecord
and another for DNSQuestion. The DNSResourceRecord panic was observed
in production and the second panic was caught by fuzzing.

Add a fuzz test with seeds to reproduce these two panics.


cirego:
@gconnell, are you still accepting pull requests for small fixes like this PR? I realize that it takes a lot of time to maintain a project and I'm very grateful for the contributions I've added so far. I don't think I can maintain this myself but we use this library at day-job and have discovered a few panics. Hopefully small fixes like this can be released.

cirego:
Thank you! :partying_face:

PR #1047, opened 2022-08-11
delete minor unreachable code caused by t.Fatal

Abirdcfly:
Signed-off-by: Abirdcfly <fp544037857@gmail.com>

### DIFF
--- a/ip4defrag/defrag_test.go
+++ b/ip4defrag/defrag_test.go
@@ -237,7 +237,6 @@ func TestDefragMaxSize(t *testing.T) {
 
 	_, err := defrag.DefragIPv4(&ip2)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ip4defrag/defrag_test.go
```diff
@@ -237,7 +237,6 @@ func TestDefragMaxSize(t *testing.T) {
 
 	_, err := defrag.DefragIPv4(&ip2)
 	if err != nil {
-		t.Fatal(err)
 		t.Fatalf("defrag: Maximum length is supposed to be 65535")
```
rski, line 240: it would be good for debugging if the error were still printed, that can be done either by changing the removed Fatal to Error instead of removing or by adding the error to the next Fatalf.

PR #1057, opened 2022-09-07
Add feature to verify checksums

MarcusWichelmann:
This commit adds a `LayerWithChecksum` interface for layers that have a checksum with a function to verify it and implements this function for all layers that use the internet checksum from RFC1071 and where the checksum computation was implemented (for `SerializeTo`).

It also adds a way to verify the checksums of all layers of a packet and retrieve a list of all found mismatches.

To implement this, I refactored the internet checksum calculation code, deduplicated it and made it a little bit more modular. There are now two functions: `ComputeChecksum` and `FoldChecksum`.
These split-up functions make it possible to verify checksums without having to zero-out the checksum that's already set in the packet. Instead, the checksum of the whole packet (including the existing checksum) is computed and then the existing checksum is subtracted, before it is folded. This way we don't need to copy the packet buffer and zero-out the checksum field, just to verify it.

The refactoring and checksum-verification are split up into two commits to make this easier to review.

For the UDP layer, there is a special case, because the UDP checksum is optional. That's why this PR is based on https://github.com/google/gopacket/pull/883 and extends it to make the checksum verification succeed, if the existing UDP checksum was unset.

### DIFF
--- a/base.go
+++ b/base.go
@@ -26,6 +26,13 @@ type Layer interface {
 	LayerPayload() []byte
 }
 
+// LayerWithChecksum should be implemented by layers that contain a checksum
+// that can be verified after a packet has been decoded.
+type LayerWithChecksum interface {
+	// VerifyChecksum verifies the checksum and returns the result.
+	VerifyChecksum() (error, ChecksumVerificationResult)
+}
+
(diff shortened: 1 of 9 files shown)


joewilliams:
Also, please merge this PR.

PR #1066, opened 2022-10-08
Fix PFLog parsing on OpenBSD.

### LINKED ISSUE #823: FreeBSD PFLog vs OpenBSD PFLog
There appear to be some differences between [OpenBSD PFlog](https://man.openbsd.org/pflog.4) and [FreeBSD PFLog](http://www.freebsd.org/cgi/man.cgi?query=pflog&sektion=4), causing issues when decoding packets, that come from OpenBSD rather than FreeBSD.

PACKET: 152 bytes, wire length 152 cap length 152 @ 2020-09-28 20:52:59.163598 +0200 CEST
- Layer 1 (152 bytes) = DecodeFailure   Packet decoding error: PFLog header length should be 3 less than multiple of

dennisoelkers:
This PR is fixing PFLog parsing for OpenBSD. On OpenBSD the packet has a different structure (see https://github.com/openbsd/src/blob/master/sys/net/if_pflog.h) where the header does not have a length that is n*4-3. It also  does not require the padding of 3 bytes being added to the header length.

Fix #823.

### DIFF
--- a/layers/pflog.go
+++ b/layers/pflog.go
@@ -57,10 +57,11 @@ func (pf *PFLog) DecodeFromBytes(data []byte, df gopacket.DecodeFeedback) error 
 	pf.RuleUID = binary.BigEndian.Uint32(data[52:56])
 	pf.RulePID = int32(binary.BigEndian.Uint32(data[56:60]))
 	pf.Direction = PFDirection(data[60])
-	if pf.Length%4 != 1 {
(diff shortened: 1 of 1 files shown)


dennisoelkers:
Hey @gconnell,

I have signed the CLA and my mail address matches the one of the commits in this PR, but the CLA check still fails. Do you have an idea what I am doing wrong?

Thanks!

PR #1075, opened 2022-11-16
Fix GENEVE test case error message 

simeonmiteff:
Error message says "VXLAN". Possibly a cut&paste error.


simeonmiteff:
Fixed in the https://github.com/gopacket/gopacket fork.

PR #1076, opened 2022-11-22
Implement support for packet options

icedream:
Adds ability to read and write packet comments as well as custom binary and string data attached to packets.

By default, the NgReader will behave as before, skipping the parsing of options due to computational effort required. The reader option "WantPacketOptions" needs to be set to true.

Testing is still to be done.


RangelReale:
this would be very useful to have.

icedream:
I have rebased this branch in hopes to get some activity back into this. The patch has actively been in use on our end for a long while now.

PR #1077, opened 2022-12-02
Support removing BPF filter from AF_PACKET socket

simeonmiteff:
Do a setsockopt with SO_DETACH_FILTER when TPacket.SetBPF() is called with a empty BPF instruction list.


simeonmiteff:
Merged in the https://github.com/gopacket/gopacket fork.

PR #1079, opened 2022-12-03
fix Resolution() function (ms/ns precision)

svlobanov:
nanoSecsFactor set to 1 for ns precision and sets to 1000 for ms precision, but the opposite logic is used in Resolution() function

this patch uses the same logic in Resolution() function as it used during reading


Nequo:
Any updates here?

PR #1082, opened 2022-12-07
Fixes for Geneve option flags/length field

simeonmiteff:
This fixes two bugs in the Geneve implementation:
- The decoder was using an incorrect number of bits for the option length field
- The encoder was generating corrupt packets on serialisation due to uninitialised buffer use


simeonmiteff:
Fixed in the https://github.com/gopacket/gopacket fork.

PR #1087, opened 2023-01-24
Fix error detection in pfring.NewRing

### LINKED ISSUE #147: "pfring NewRing error: no such device" when trying to open pfring ZC interface from zbalance_ipc
PF_RING in ZC-mode comes with a program zbalance_ipc that can send the output of one or more ZC-queues to different virtual PF_RING interfaces that end up with names like "zc:99@0" and "zc:99@1" etc.

When using gopacket (which otherwise works fine with normal ZC-interfaces like e.g. "zc:eth3") a "no such device" error is returned. As far as I can tell there is nothing special happening inside NewRing() besides calling C.pfring_open(). Using these virtual interfaces work fine with the example programs that come with PF_RING.

Is this somehow a bug with gopacket? It seems to be doing basically the same thing as the C-examples I've seen that work fine, but returns the error.

Code that fails:

    log.Printf("Opening pf_ring on interface %q", *iface)
    ring, err = pfring.NewRing(*iface, uint32(*snaplen), pfring.FlagPromisc) // <- fails
    if err != nil {
        log.Printf("Error: %s, Ring: %v\n", err, ring)
        return
    }
    
    err = ring.Enable()
    if err != nil {
        log.Printf("Error: %s\n", err)
        return
    }
    
    packetSources = append(packetSources, gopacket.NewPacketSource(ring, layers.LinkTypeEthernet))
    
    log.Printf("New pf_ring.")
    defer ring.Close()

EtienneMiret:
Fixes #147

The cgo documentation states:

> Any C function (even void functions) may be called in a multiple assignment context to retrieve both the return value (if any) and the C errno variable as an error

See: <https://pkg.go.dev/cmd/cgo>

Furthermore, the manpage for errno states:

> The  value  in  errno is significant only when the return value of the call indicated an error (i.e., -1 from most system calls; -1 or NULL from most library functions); a function that succeeds is allowed to change errno. The value of errno is never set to zero by any system call or library function.

See: man 3 errno

Therefore, I believe the value of err MUST be ignored unless we have something else that reports an error.


karlyeurl:
Hello,

If the error is to be ignored, isn't it better to avoid assigning it in the first place?

Something like `cptr, _ := C.pfring_open(dev, C.u_int32_t(snaplen), C.u_int32_t(flags))` would make it clear that we do not care about the error returned by pfring_open().

EtienneMiret:
Hi @karlyeurl!

Actually, we do care about it when `cptr` is nil. This means the `pfring_open()` failed, and in that case we are allowed (and should) look at the C `errno` variable to have more details about which error occurred.

Issue #1053, opened 2022-08-28
Allow configuration of SEESENT in bsdbpf BPF sniffer


Doridian:
Right now, there is no way to set the SEESENT parameter of the BPF packet capturing system in bsdbpf.

Since I use the system for creating a VPN interface (best way to create a TAP-type VPN on macOS known so far afaik), I have no interest in the sent packets at all. In fact, I explicitly do not want to handle them.

BPF does offer the "SEESENT" ioctl for this. Sadly, while the BPF sniffer allows configuring almost all other options I need, this one is not supported.

Right now, the way I work around this problem is with some quite hacky reflection, like:
```go
        bpfCapture, err := bsdbpf.NewBPFSniffer(.....)

        [...]

	bpfFd := getPrivateField(bpfCapture, "fd").(int)
	var enable int = 0
	_, _, errno = syscall.Syscall(syscall.SYS_IOCTL, uintptr(bpfFd), uintptr(syscall.BIOCSSEESENT), uintptr(unsafe.Pointer(&enable)))
	if errno != 0 {
		bpfCapture.Close()
		closer.Close()
		return nil, fmt.Errorf("bpf ioctl error = %d", errno)
	}
```

PR #1091, opened 2023-02-05
Miscellaneous bsdbpf fixes and improvements

srgg:
In order to more or less unify afpacket and bsdbpf  the following new methods will be added to BPFSniffer:

-  SetBPFFilter(): to set BPF programm,
- WritePacketData(): to write data.

The new  configurable option `SeeSent` will be added, it allows  to turn on/off returning local packets over filter.

There is also several fixes:

- BPFSniffer creation with bpf device name provided over config options;
-  Avoid fd leak in the case of any subsequent error during BPFSniffer instance creation.

This PR also fixes https://github.com/google/gopacket/issues/1053 and makes obsolete https://github.com/google/gopacket/pull/1054

### DIFF
--- a/bsdbpf/bsd_bpf_sniffer.go
+++ b/bsdbpf/bsd_bpf_sniffer.go
@@ -4,6 +4,7 @@
 // that can be found in the LICENSE file in the root of the source
 // tree.
 
+//go:build darwin || dragonfly || freebsd || netbsd || openbsd
 // +build darwin dragonfly freebsd netbsd openbsd
 
 package bsdbpf
@@ -11,6 +12,7 @@ package bsdbpf
 import (
 	"errors"
 	"fmt"
+	"github.com/google/gopacket/pcap"
 	"syscall"
 	"time"
 	"unsafe"
@@ -22,7 +24,7 @@ import (
 const wordSize = int(unsafe.Sizeof(uintptr(0)))
 
 func bpfWordAlign(x int) int {
-	return (((x) + (wordSize - 1)) &^ (wordSize - 1))
+	return ((x) + (wordSize - 1)) &^ (wordSize - 1)
 }
 
 // Options is used to configure various properties of the BPF sniffer.
@@ -58,6 +60,9 @@ type Options struct {
 	// as provided, to the wire.
 	// The default is true.
 	PreserveLinkAddr bool
+	// SeeSent is set to false if locally generated packets on the interface should not be returned by BPF.
+	// The default is true.
+	SeeSent bool
 }
 
 var defaultOptions = Options{
@@ -67,6 +72,7 @@ var defaultOptions = Options{
 	Promisc:          true,
 	Immediate:        true,
 	PreserveLinkAddr: true,
+	SeeSent:          true,
 }
 
 // BPFSniffer is a struct used to track state of a BSD BPF ethernet sniffer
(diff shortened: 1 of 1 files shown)


bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -143,19 +147,30 @@ func NewBPFSniffer(iface string, options *Options) (*BPFSniffer, error) {
...
 		err = syscall.SetBpfPromisc(sniffer.fd, enable)
 		if err != nil {
-			return nil, err
+			goto err
+		}
+	}
+
+	if !sniffer.options.SeeSent {
+		// See sent is set by default, need to turn it off
+		err = unix.IoctlSetInt(sniffer.fd, syscall.BIOCSSEESENT, 0)
```
Doridian, line 164:
```suggestion
		err = unix.IoctlSetPointerInt(sniffer.fd, syscall.BIOCSSEESENT, 0)
```

bsdbpf/bsd_bpf_sniffer.go
```diff
@@ -165,13 +180,26 @@ func (b *BPFSniffer) Close() error {
...
+	if len(b.options.BPFDeviceName) > 0 {
 		b.fd, err = syscall.Open(b.options.BPFDeviceName, syscall.O_RDWR, 0)
 		if err == nil {
 			return
 		}
+	} else {
+		b.options.BPFDeviceName = ""
+		for i := 0; i < 99; i++ {
+			name := fmt.Sprintf("/dev/bpf%d", i)
+			b.fd, err = syscall.Open(b.options.BPFDeviceName, syscall.O_RDWR, 0)
```
Doridian, line 192:
```suggestion
			b.fd, err = syscall.Open(name, syscall.O_RDWR, 0)
```

bsdbpf/bsd_bpf_sniffer.go
srgg: @Doridian, I do not understand this suggestion. Can you please elaborate on it? Here is a definition from bpf man page saying it is **u_int**:
```
     BIOCSSEESENT
     BIOCGSEESENT    u_int These commands are	obsolete but left for compati-
		     ability.  ... This flag is initialized to
		     one by default.
```
from https://man.freebsd.org/cgi/man.cgi?bpf(4)

Thank you for the review

Doridian: @srgg Most write `ioctl` calls actually expect pointers to the values you set. Including `SEESENT`.
This is based on automated testing of my VPN project, where using only `IoctlSetPointerInt` yielded a functional tunnel (the tunnel server, when seeing its own packets, just freaks out as it then tries to relay its own packets back)

srgg: @Doridian, I see your point. Curious if it is a difference between Darwin and BSD? And AFAIK, it works on Darwin. 

Agreed IOCTLs are weird, but they should be used as expected by the correspondent kernel module (and these two calls are pretty different, I've provided a code snippet below). I cannot access the project I worked on when this PR was created. Can I merge your suggestion w/o extra verification? 


```
// IoctlSetInt performs an ioctl operation which sets an integer value
// on fd, using the specified request number.
func IoctlSetInt(fd int, req uint, value int) error {
	return ioctl(fd, req, uintptr(value))
}

// IoctlSetPointerInt performs an ioctl operation which sets an
// integer value on fd, using the specified request number. The ioctl
// argument is called with a pointer to the integer value, rather than
// passing the integer value directly.
func IoctlSetPointerInt(fd int, req uint, value int) error {
	v := int32(value)
	return ioctl(fd, req, uintptr(unsafe.Pointer(&v)))
}

```

Doridian: @srgg I was testing this on GitHub's macOS runners (and my MacBook), and only `IoctlSetPointerInt` worked. I actually did not test this on a more "normal" BSD.

srgg:
@Doridian, I applied your suggestions. Thank you for the review!

srgg:
@Doridian  can you please accept CLA to get a chance to get this merged at some point?

PR #1124, opened 2023-05-23
fix(ip4defrag): allow final fragment to be less than 8 octets

niklaskb:
This PR will fix a bug where the defragmenter fails on pcap files that has a final fragment that is smaller than 8 octets.
The final fragment should be allowed to be less than 8 octets according to the standard.
 
For reference:
https://packetpushers.net/ip-fragmentation-in-detail/
https://stackoverflow.com/a/7846487/1642369


niklaskb:
See https://github.com/gopacket/gopacket/pull/35#issuecomment-1814248569 for test

PR #1128, opened 2023-07-06
fix: Truncate RADIUS attributes before decoding

hacel:
I noticed that when trying to decode a RADIUS layer multiple times using the same `RADIUS` struct, the attributes from the new packet are appended to the attributes of all previous packets.

As far as I can tell, the current RADIUS implementation is broken for in-place decoding.


hacel:
Signed.

PR #1140, opened 2023-10-22
Making pcapgo Writer routine safe

gboddin:
Calling write from parallel routines will cause the pcap file to corrupt.

Using a mutex ensure we don't corrupt the file by delaying concurrent writes.

An alternative method would be to use a buffered channel but requires an additional routine


mosajjal:
hey mate, can you move this to `gopacket/gopacket` and raise it there. I'll merge it over there.

PR #1151, opened 2024-01-04
Add functions to write TLS key-logs to PcapNG files

9072997:
This PR adds 2 functions to support writing additional block types in PcapNG files.

- WriteTLSKeyLog, which allows embedding TLS secrets into a PcapNG file. This is especially handy if you have an application which is monitoring itself. It makes it possible to generate self-contained capture files which can be easily viewed in Wireshark.
- WriteCustomBlock, which makes it easier for users of the library to add support for block types not supported by the library.

No support is added for reading additional block types.


wolveix:
Hey @9072997, would you mind opening this PR on https://github.com/gopacket/gopacket? It's an actively maintained community fork :)

PR #1202, opened 2025-02-07
make Geneve implement DecodingLayer

kaworu:
Before this patch, the `Geneve` layer would not fully implement `DecodingLayer`. This would prevent `Geneve` to be added as a `DecodingLayer` to `gopacket.NewDecodingLayerParser()`:

    cannot use &Geneve{} (value of type *Geneve) as gopacket.DecodingLayer value in argument to gopacket.NewDecodingLayerParser: *Geneve does not implement gopacket.DecodingLayer (missing method CanDecode)

This patch implements `(*Geneve).CanDecode()` and a test to ensure `Geneve` can be used as a `DecodingLayer`.

### DIFF
--- a/layers/geneve.go
+++ b/layers/geneve.go
@@ -48,6 +48,9 @@ type GeneveOption struct {
 	Data   []byte
 }
 
+// ensure Geneve implements DecodingLayer.
+var _ gopacket.DecodingLayer = (*Geneve)(nil)
+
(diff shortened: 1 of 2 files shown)


kaworu:
Didn't see there is already a PR open for this bug https://github.com/google/gopacket/pull/1172

PR #1217, opened 2025-10-04
Add IV and KeyID parsing to Dot11WEP

joejoe-am:
This PR adds support for parsing the IV and Key ID fields in the Dot11WEP layer.

Details
  •	Added IV [3]byte and KeyID uint8 fields to Dot11WEP
  •	Updated DecodeFromBytes to:
  •	Extract the 24-bit IV from the first three bytes
  •	Extract the 2-bit Key ID from the top bits of the fourth byte
  •	Updated GoDoc for Dot11WEP to document the header layout

Why

The IV and Key ID are essential for WEP decryption and analysis. Previously, users needed to parse these fields manually from the payload. Exposing them in Dot11WEP simplifies working with WEP frames and brings the layer in line with how other 802.11 fields are exposed.

### DIFF
--- a/layers/dot11.go
+++ b/layers/dot11.go
@@ -1160,9 +1160,20 @@ func decodeDot11Ctrl(data []byte, p gopacket.PacketBuilder) error {
 	return decodingLayerDecoder(d, data, p)
 }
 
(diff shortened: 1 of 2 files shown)


joejoe-am:
Hi @gconnell, can you review this please ?

PR #1235, opened 2026-05-01
pcapgo: reject EPB/PB with CaptureLength exceeding block length

adilburaksen:
## Summary

`NgReader.readPacketHeader()` in `pcapgo/ngread.go` applied a `SnapLength` bounds-check only for `ngBlockTypeSimplePacket`, leaving `ngBlockTypeEnhancedPacket` (EPB) and `ngBlockTypePacket` (legacy PB) unguarded. A crafted pcapng file could trigger an unbounded allocation in `ReadPacketData()` and `ZeroCopyReadPacketData()` via the subsequent `make([]byte, r.ci.CaptureLength)` call.

**Reproducer**: an 80-byte pcapng containing a valid SHB + IDB + EPB with `CaptureLength = 0xFFFFFFFF` causes a **4 GB** allocation attempt before returning EOF.

## Changes

- `pcapgo/ngread.go`: for EPB and legacy PB, validate that `CaptureLength` does not exceed the remaining block length before allocating; then clamp against interface `SnapLength` when set (matching existing SPB behaviour).
- `pcapgo/ngread_test.go`: add `TestNgEPBOversizeCaptureLength` and `TestNgEPBOversizeCaptureLengthZeroCopy` — each constructs a 32-byte pcapng with a 512 MB `CaptureLength` and asserts that both read paths return an error rather than attempting the allocation.

## Test plan
- [ ] `go test ./pcapgo/...` passes (includes new regression tests)
- [ ] `TestNgEPBOversizeCaptureLength` and `TestNgEPBOversizeCaptureLengthZeroCopy` pass
- [ ] Existing round-trip and file-read tests unchanged


adilburaksen:
Gentle ping — happy to address any review feedback or adjust the fix if needed.

PR #1236, opened 2026-05-17
fix: close packet channel on non-temporary errors

### LINKED ISSUE #1143: `(*PacketSource).packetsToChannel()` error handling is incomplete
## TL;DR 
The error handling in `(*PacketSource).packetsToChannel()` is incomplete and may lead to starvation. A catch-all rule is missing. 

### Implementation
```go
	f, _ := os.Open("../testdata/somepackets.pcap")
	defer f.Close()
	r, _ := pcapgo.NewReader(f)

	pktsrc := gopacket.NewPacketSource(r, r.LinkType())
	
	f.Close() // close the reader here

	for packet := range pktsrc.Packets() { // this will block after the last packets
		// do something
	}
```
### Behavior
The program hangs instead of terminating. 

## Analysis 

https://github.com/google/gopacket/blob/32ee38206866f44a74a6033ec26aeeb474506804/packet.go#L815-L845

The function implementation missed the case where the returned error is not matched to any known cases. 

In the current latest version of Go (1.21.4), `(*os.File).Read()` returns `read /path/to/the/file: file already closed`, which wraps `os.ErrClosed` and could be tested with `errors.Is(err, os.ErrClosed)`, but is not __equal__ to any existing error. Therefore it will not break out of the `for` loop and the channel will never be closed. 

## Proposed fix
- Testing against known errors should use `errors.Is` or its equivalent, not `err == ErrToCompare`. 
- Should also add a catch-all case for unknown errors, since it is possible for the user to use a custom `io.Reader`.

I'd be happy to open a PR if the proposed fix sounds reasonable.

wyf027:
## Summary
- Close `PacketSource.Packets()` when `NextPacket` returns an unknown non-temporary error instead of retrying forever.
- Match known retry/terminal errors through a small `Unwrap`-aware helper so wrapped errors are classified consistently without requiring `errors.Is` (the module still declares Go 1.12).
- Add a regression test covering the starvation case where `Packets()` previously never closed after a persistent non-temporary read error.
- Update `Packets` documentation to describe the non-temporary error behavior.

## Related Issues / Closes
Closes #1143

## Testing
- `go test . -run TestPacketSourcePacketsClosesOnNonTemporaryError -count=1` (passes)
- `go test .` (passes)
- `go vet .` (passes)
- `git diff --check` (passes)

Additional validation:
- I also ran `go test ./...`; it fails in unrelated platform/environment-dependent packages on this macOS runner:
  - `examples/afpacket`: Linux-only `afpacket` build constraints exclude all Go files.
  - `pfring` / `examples/pfdump`: missing system header `pfring.h`.
  - `pcap`: `TestPCAPGoNgWrite` reports `could not read first packet: Read Error`.

## AI assistance disclosure
This draft PR was prepared with AI assistance from Hermes Agent. I reviewed the issue context, implemented the minimal change, and ran the validation listed above.


wyf027:
Closing this stale PR because it has been blocked on CLA verification since May 17, 2026 and has not had a clear merge path for more than 14 days. Thanks.

